PR #107, opened 2014-02-22
Allow to run make loop manually instead of a timeout

panrafal:
This allows using fibers-based flow control in make targets.
make.run() is always called only once. If you omit it, it will run at the end of the script as usual.

```
var Sync = require('sync'),
  // require shelljs/make at the beggining, but store the exports
  Make = require('shelljs/make') 

Sync(function() {
  target.all = function() {
    var result = asyncFunction.sync();
  }

  // run make targets now 
  Make.run()
})
```

### DIFF
--- a/make.js
+++ b/make.js
@@ -5,9 +5,12 @@ global.target = {};
 
 // This ensures we only execute the script targets after the entire script has
 // been evaluated
-var args = process.argv.slice(2);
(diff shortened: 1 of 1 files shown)


arturadib:
closing as tests don't pass and it seems abandoned, please reopen if not!

PR #134, opened 2014-06-15
Search PATHEXT instead of 3 hardcoded values

isiahmeadows:
Better to search the PATHEXT variable instead of hard-coding an incomplete list for Windows executables. A default fallback consisting of XP's default value is also introduced just in case. The fallback value is hardcoded as a constant at the top of the file.

### DIFF
--- a/src/which.js
+++ b/src/which.js
@@ -2,6 +2,9 @@ var common = require('./common');
 var fs = require('fs');
 var path = require('path');
 
+// XP's system default value for PATHEXT system variable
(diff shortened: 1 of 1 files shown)


arturadib:
Hi- thanks for the patch. Alas I no longer have a Windows machine to test this on and Travis doesn't support Windows, can someone please confirm the tests pass? Thanks!

nfischer:
@isiahmeadows would you be able to rebase this off master? This might be a valuable addition, since we ran into that issue in #239 (and the PATHEXT variable was brought to my attention). @BYK could you review?

isiahmeadows:
Closed this and redid it in #290.

Problem was that I have changed my username since I made the original PR.

PR #137, opened 2014-06-29
Use inspect() to show a list in console instead of raw array object

piranna:
As shown on http://nodejs.org/api/util.html#util_customizing_util_inspect_colors, objects can has a ```inspect()``` method that will be used by ```util.inspect()``` on the Node.js REPL, so we can show on the console an output more similar to the actual one on the shell, while still returning the actual array so it can be used by other functions:

```Javascript
> require('./global')
{}
> var a = ls()
undefined
> a
LICENSE
README.md
bin
global.js
make.js
package.json
scripts
shell.js
src
test
> a instanceof Array
true
> a.join()
'LICENSE,README.md,bin,global.js,make.js,package.json,scripts,shell.js,src,test'
> a.inspect
[Function: inspect]
> a.inspect()
'LICENSE\nREADME.md\nbin\nglobal.js\nmake.js\npackage.json\nscripts\nshell.js\nsrc\ntest'
> JSON.stringify(a)
'["LICENSE","README.md","bin","global.js","make.js","package.json","scripts","shell.js","src","test"]'
```

I've only apply this to the array returned by ```ls()```, but it could be done in some other places. A good candidate to do so is ```grep()```, where it's currently returning a [multiline string](https://github.com/arturadib/shelljs/blob/master/src/grep.js#L46) while it could be better to return an array (with ```inspect()``` method) instead as ```ls()``` does.

I'm using ```Object.defineProperty()``` to set ```inspect()``` so it doesn't became an enumerable property.

### DIFF
--- a/scripts/run-tests.js
+++ b/scripts/run-tests.js
@@ -30,8 +30,9 @@ if (exec(JSHINT_BIN + ' *.js test/*.js').code !== 0) {
 //
 cd(__dirname + '/../test');
 ls('*.js').forEach(function(file) {
-  echo('Running test:', file);
-  if (exec('node ' + file).code !== 123) { // 123 avoids false positives (e.g. premature exit)
+  var name = file.name
+  echo('Running test:', name);
+  if (exec('node ' + name).code !== 123) { // 123 avoids false positives (e.g. premature exit)
     failed = true;
     echo('*** TEST FAILED! (missing exit code "123")');
     echo();
--- a/src/cd.js
+++ b/src/cd.js
@@ -1,12 +1,18 @@
-var fs = require('fs');
+var fs   = require('fs');
+var path = require('path');
+
 var common = require('./common');
 
 //@
 //@ ### cd('dir')
 //@ Changes to directory `dir` for the duration of the script
 function _cd(options, dir) {
-  if (!dir)
-    common.error('directory not specified');
+  var env = process.env
+
+  if (!dir) dir = env.HOME;
+
+  dir.replace(/^~\//, function () { return env.HOME + '/' });
+  dir = path.resolve(env.PWD, dir);
 
   if (!fs.existsSync(dir))
     common.error('no such file or directory: ' + dir);
@@ -15,5 +21,6 @@ function _cd(options, dir) {
     common.error('not a directory: ' + dir);
 
   process.chdir(dir);
+  env.PWD = dir;
 }
 module.exports = _cd;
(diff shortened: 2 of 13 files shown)


nfischer:
@ariporad what are your thoughts? I like the idea of having `grep()` and `sed()` returns lists similarly to how `ls()` does (it seems more useful to me). The output on the node commandline isn't as important to me, and I think it might actually be confusing to display something other than what is returned.

@piranna the colors definitely rock though! The only significant one missing from `ls()` is green for executables, I think.

@piranna would you be interested in refactoring this into several smaller PRs? I think `ls -l`, for example, would benefit users, and we could merge that much sooner than the whole chunk.

ariporad:
Yeah. I'm not sure about the inspect bit... It could be confusing to some people.

ariporad:
@piranna: I'm not really sure about the `inspect()` thing, I feel like it's a little counter-intuitive.

@nfischer: I personally vote to close this. What do you think?

nfischer:
@ariporad I don't think we can merge this as-is, because it implements too many things.

@piranna I'd really like to merge some of these features, but I would need them to be done in separate PRs. If you or anyone else would like to refactor some of these features into separate PRs, that would be very appreciated.

piranna:
> The output on the node commandline isn't as important to me, and I think it might actually be confusing to display something other than what is returned.

When used on interactive REPL it matters, specially with the `-l` flag.


> @piranna the colors definitely rock though! The only significant one missing from ls() is green for executables, I think.

Colors come from the `$LS_COLORS` environment variable, so they are the ones defined on the system, I can't do here nothing.


> @piranna would you be interested in refactoring this into several smaller PRs? I think ls -l, for example, would benefit users, and we could merge that much sooner than the whole chunk.

I started [Coreutils.js](https://github.com/piranna/coreutils.js), my own implementation on the Unix commands using streams of objects instead of arrays. I think this is fairly a better approach to that problem.


> @piranna I'd really like to merge some of these features, but I would need them to be done in separate PRs. If you or anyone else would like to refactor some of these features into separate PRs, that would be very appreciated.

You can cherry pick the parts you are interested about, but definitely I think `inspect()` support is a plus, if not it would not be used on the Node.js core libraries... One thing is how are the data internally, and how it's represented, and an example of this is the `-l` flag that's only purposse are visual artifacts.

nfischer:
@ariporad I vote to put off adding `inspect()`, and focus on the underlying details, like `ls -l`, and turning multi-line output into arrays where it makes sense.

ariporad:
@nfischer: SGTM.

ariporad:
@nfischer: I vote we close this pull request, because it's really tied to the current `ls` implementation, which is about to be replaced.

nfischer:
SGTM

PR #139, opened 2014-07-06
redux adding -l and -s support to grep. 

idearat:
Merged with the latest master and tested locally and on travis @ https://travis-ci.org/idearat/shelljs and everything passed cleanly. Hopefully we'll have better luck this time :).

### DIFF
--- a/src/grep.js
+++ b/src/grep.js
@@ -19,14 +19,18 @@ var fs = require('fs');
 //@ file that match the given `regex_filter`. Wildcard `*` accepted.
 function _grep(options, regex, files) {
   options = common.parseOptions(options, {
-    'v': 'inverse'
+    'v': 'inverse',
+    'l': 'files-with-matches',
+    's': 'no-messages'
   });
 
-  if (!files)
+  if (!files) {
     common.error('no paths given');
+  }
 
-  if (typeof files === 'string')
+  if (typeof files === 'string') {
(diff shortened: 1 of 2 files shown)


nfischer:
@idearat thanks for the PR! Would you be able to rebase off of master?

Also, what does `-s` add that `config.silent` doesn't? My concern is that `common.error()` isn't called if the `-s` flag is passed, so a `grep()` command that finds no results will technically succeed (based on checking with `shell.error()`), which differs from unix behavior.

idearat:
@nfischer It's been a while since I authored this. I'll need to look back at it and take your comments re: common.error() into consideration. Will try to update within a day or two. And thanks for taking it under consideration!

nfischer:
@idearat thanks! And I completely understand, it's been quite some time. We're trying to get on top of the old PRs, and it's probably better to follow up than to just let them sit longer.

ariporad:
@nfischer, @idearat: bump.

nfischer:
`grep -l` has been implemented and merged. I think `grep -s` isn't worth implementing, since it can be emulated with `config.silent = true; grep('-s' ...); config.silent = false;`. I'm going to close this for now, feel free to re-open if `-s` is important (or if `-l` support isn't quite right).

idearat:
Seems reasonable.

PR #149, opened 2014-08-13
Added option to print the unix like command itself

geekflyer:
This is similar to the option `set -o verbose` in regular shell scripts. Partially related to #8.
Without this option some commands are completely silent, e.g. `mkdir('somedir')` would just create a directory silently. With the logCmd option it will additionally print `mkdir somedir`, to give the user an idea what the script is actually doing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,6 +467,7 @@ Available options (all `false` by default):
 
 + `async`: Asynchronous execution. Defaults to true if a callback is provided.
 + `silent`: Do not echo program output to console.
++ `logCmd`: Prints the unix like command itself to the console. Similar to `set -o verbose` in shell scripts.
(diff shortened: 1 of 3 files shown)


nfischer:
This could be a cool option to add as part of a `set` command. @geekflyer would you have any interest in writing a `set` command to act as a wrapper for `config` variables? If we can get that out for `config.fatal`, then we might be able to add this as well.

If it uses a `set()` interface, it'll probably be more intuitive to Bash scripters, and might help tame the docs section (which is rather lengthy).

nfischer:
Thanks so much for the PR @geekflyer! I think we'll have to close this, though, since this has been incorporated in #329

PR #150, opened 2014-08-15
Fixed force linking to an existing directory

wesleytodd:
Previously, trying to symlink to an existing directory with the `force` flag would error.  It was calling `unlink` on a directory.  This fixes that by checking if it is a directory before removing.

Let me know if there is anything else you need to merge this!

### DIFF
--- a/src/ln.js
+++ b/src/ln.js
@@ -40,8 +40,12 @@ function _ln(options, source, dest) {
     if (!options.force) {
       common.error('Destination file exists', true);
     }
-
-    fs.unlinkSync(dest);
+    var stat = fs.lstatSync(dest);
+    if (stat.isDirectory()) {
+      fs.rmdirSync(dest);
+    } else {
+      fs.unlinkSync(dest);
+    }
   }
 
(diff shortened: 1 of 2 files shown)


nfischer:
@wesleytodd does this conform with unix behavior? I can't reproduce the expected result. To me it looks like it's supposed to put the symlink inside the directory `dir/`

wesleytodd:
Tbh I dont know if I have used this since I submitted this. I can try to look tomorrow. But no promises 😀

nfischer:
:+1: I figure it's better to follow up on an old PR than to just let it sit even longer.

wesleytodd:
Yep!  Sadly since I'm not using it anymore I am not very incentivized to help out. That is the price of leaving OS issues for too long :). If you dont think it is a good pr then feel free to close. Wont offend me

nfischer:
I think I'm going to opt to close this, since I think it differs from unix behavior. Thanks for the contribution though @wesleytodd! If you think this is an important feature, please feel free to ping this thread or reopen.

PR #151, opened 2014-08-17
support costom echo in exec command

tminglei:
`shelljs` makes me happy when writing automatical scripts. Thanks so much! :+1: 

But when executing parallel jobs, to distinguish different child process' console out, I had to make my custom execAsync:
```javascript
function execAsync(cmd, opts, callback) {
  opts = opts || {};

  var c = child.exec(cmd, { env: process.env }, function (err) {
    if (callback) callback(err ? err.code : 0);
  });

  var echo = opts.echo || process.stdout.write;
  c.stdout.on('data', function (data) {
  	echo(data);
  });

  c.stderr.on('data', function (data) {
  	echo(data);
  });

  return c;
}
exports.execAsync = execAsync;
```

Which enable me pass in my custom echo.

And my custom echo will auto add a specified prefix to every echo message. Its factory function is like this:
```javascript
function prefixEcho(prefix) {
  return function (message) {
    echo('[' + prefix + '] ' + message.replace(/\n$/, ''));
  }
}
exports.prefixEcho = prefixEcho;
```

Realizing this feature will also be useful for other people, I made the pull request.

**Can you help review and give your suggestions?**

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -96,11 +96,12 @@ function execSync(cmd, opts) {
 
 // Wrapper around exec() to enable echoing output to console in real time
 function execAsync(cmd, opts, callback) {
-  var output = '';
-
   var options = common.extend({
     silent: common.config.silent
   }, opts);
+
+  var echo = options.echo || process.stdout.write,
+    output = '';
 
   var c = child.exec(cmd, {env: process.env, maxBuffer: 20*1024*1024}, function(err) {
     if (callback)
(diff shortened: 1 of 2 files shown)


nfischer:
@tminglei Thanks for the PR! Is it not sufficient to output a process's `stdout` after it finishes? That would be achievable with a callback.

Or, if that doesn't work, would you be able to achieve this with a user-defined function if it could be used as a callback on each line of output? #277 requests the feature for line-by-line callbacks, which seems like it could help solve your issue.

tminglei:
Yes, it can. Let's continue it in #277.
_Thanks!_

nfischer:
:+1:

PR #152, opened 2014-08-20
Add array of commands in exec

### LINKED ISSUE #103: Allow multiple string arguments for exec()
I've found that I end up doing a lot of messy string concatenation with `exec()` commands. It would be nice to allow multiple string arguments that would be joined together with a space, such as:

``` js
exec('npm', 'run-script', 'test');   // run "npm run-script test"
```

This could easily be backwards compatible with the current function since only the first argument is currently a string.

isRuslan:
Hi everyone.
I suggest to add possibility to set array of commands in first argument of exec method.
Example:
    
    shell.exec(['git pull', 'npm i'], function(code, output) {
    
    });

Fixes #103

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -122,6 +122,21 @@ function execAsync(cmd, opts, callback) {
   return c;
 }
 
+// Wrapper for series execAsync() running
+function series (cmds, opts, callback) {
+  var out = '';
+
+  var next = function(output) {
+    execAsync(cmds.shift(), opts, function(code, output) {
+      if (code) return callback(code, null);
+      out += output;
+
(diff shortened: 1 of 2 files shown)


ariporad:
@nfischer: What are your thoughts? I'm not so sure about this.

nfischer:
I vote to close this. I think Array.join(';') is good enough for most users

ariporad:
Me too. Closing.

ariporad:
@isRuslan: Thanks so much for your PR, but unfortunately we're going to have to close this. It's just a very strange use-case, and `[command1, command2, ...commandN].join(';')` should really work for most cases. Thanks!

PR #163, opened 2014-10-17
Provide `cwd` option for exec.

taoyuan:


oNaiPs:
+1

nfischer:
@taoyuan Thanks for submitting the PR!

@taoyuan @oNaiPs What is the benefit of this PR vs. the following?

```Javascript
cd('path/to/dir');
exec('git status');
cd('-'); // <- the "jump back" feature used here will be released in 6.x, but it's currently merged on master
```

This has 3 lines of code, but it's very intuitive to a bash programmer. If there's an advantage to the `cwd` option, then we can consider merging :+1:

oNaiPs:
If you want to use shelljs asynchronously, it makes things more difficult to manage.

```
cd('path/to/dir1');
exec('git status', function(){
   cd('-');
});

cd('path/to/dir2');
exec('git status', function(){
   cd('-');
});
```

Would this even work?

nfischer:
@taoyuan I think it would look a bit like this, right?

```Javascript
cd('path/to/dir1');
exec('git status', {async: true}); // executes asynchronously in directory 1
cd('-');

cd('path/to/dir2');
exec('git status', {async: true}); // executes asynchronously in directory 2
cd('-');
```

Correct me if I'm missing something

ariporad:
@nfischer: I think that this is a good idea. There are a whole host of reasons you might want to specify a `cwd

taoyuan:
@nfischer Yes. It works fine in most cases. But in some case, I want to exec command without change current directory, so I must wrap a function:

```js
funciton exec(..., opts) {
  ...
  var owd = process.cwd();

  if (opts.cwd) cd(opts.cwd);
  result = exec(command, opts); 
  if (opts.cwd) cd(owd);
}
```

If we support `options.cwd`,  It will be simple :

```js
exec(command, opts); 
```

README.md
```diff
@@ -467,6 +467,8 @@ Available options (all `false` by default):
 
 + `async`: Asynchronous execution. Defaults to true if a callback is provided.
 + `silent`: Do not echo program output to console.
++ `cwd`: Working directory. Defaults to process.cwd().
```
nfischer: Make this change in the comments. Run `scripts/generate-docs.js` to generate the README

nfischer:
@taoyuan Could you please rebase off master? We've had quite a few changes, and there are now merge conflicts. Also, please remove the section about `env`. Feel free to open that as another PR, but I'm less in favor of that. If someone needs that flexibility, they can always use `child_process.exec()`. This is supposed to be a simple wrapper for shell-like use cases.

I left a comment above that should be addressed. Also, please add a unit test to test the `cwd` option. Perhaps you could run `git status` in the `..` directory and check that it fails, and run git status without the `cwd` option and check that it succeeds.

ariporad:
@taoyuan: As much as I hate to reject a PR, it looks like this has been fixed by #335, so I have to close this. Sorry.

PR #169, opened 2014-12-09
Add existsSync module that works on node 0.6

ljharb:
This change makes all the tests pass in `node` `0.6` - which also means that any dependents of it won't automatically be broken in `v0.6` :-)

### DIFF
--- a/src/cat.js
+++ b/src/cat.js
@@ -1,5 +1,6 @@
 var common = require('./common');
 var fs = require('fs');
+var existsSync = require('./existsSync');
 
 //@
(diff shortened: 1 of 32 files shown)


ljharb:
Now that there's two stable `iojs` and `node` versions that ship with an `npm` that supports `^`, I've decided to stop bothering with `0.6` support. Sorry for the noise :-)

arturadib:
sweet! now if only we replaced my `execSync` hack with the shiny new `child_process.execSync`.. :)

Issue #49, opened 2013-01-07
Error ENOTEMPTY when deleting a directory recursively.


rmariuzzo:
Sometimes, when deleting a directory that has sub-directories and files it throw an error witch code: `ENOTEMPTY`.

When executing this: `rm('-rf', '/path/to/build/zip')` the following error is thrown:

> rm: could not remove directory (code ENOTEMPTY): build/zip/vendors/bootstrap

Note: It can happen with any sub-directory, is not related to a particular directory only. It doesn't happens with files.

andreialecu:
+1 Seem to be getting the same error here on Windows occasionally. It's annoying becuase I'm running a pretty big test suite and the tests randomly fail because of this and I need to restart it every time until it eventually works.

andreialecu:
I have changed the code block at https://github.com/arturadib/shelljs/blob/master/src/rm.js#L49 with this:

```
  var result, files;
  try {
    files = fs.readdirSync(dir);
    result = fs.rmdirSync(dir);
  } catch(e) {
    common.error('could not remove directory (code '+e.code+'): ' + dir, true);
    if (e.code == "ENOTEMPTY") {
      common.error('files found: ' + files.join(', '));
    }
```

And I can indeed see that there are some files left behind sometimes, randomly.

andreialecu:
I think the only way to fix this on Windows is to auto-retry. I have found a library that doing exactly this here:
https://github.com/isaacs/rimraf

andreialecu:
I made a pull request for this that auto-retries for up to 1 second on Windows. See #179

nfischer:
Closing this since #179 is merged. Please reopen if this error persists

dpmott:
For me, shx version 0.3.2 exhibits this same intermittent behavior when run from an npm script from within VSCode; when it fails, it fails immediately (no waiting for up to 1 second).  It would appear that the changes from https://github.com/shelljs/shelljs/pull/179 may have been lost.

nfischer:
I still see the changes:

https://github.com/shelljs/shelljs/blob/4e38240eda2f377e2219f61dcf53e8f9bff74c0a/src/rm.js#L53-L78

I don't know enough about this problem to know if the bug still exists though.

PR #179, opened 2015-01-11
Fix issue #49: Retry rmdirSync on Windows for up to 1 second if files still exist.

andreialecu:
See issue #49

### DIFF
--- a/src/rm.js
+++ b/src/rm.js
@@ -48,6 +48,7 @@ function rmdirSyncRecursive(dir, force) {
 
   var result;
   try {
+    // Retry on windows, sometimes it takes a little time before all the files in the directory are gone
     var start = Date.now();
     while (true) {
       try {
@@ -55,10 +56,8 @@ function rmdirSyncRecursive(dir, force) {
         break;
       } catch(er) {
         if (process.platform === "win32" && (er.code === "ENOTEMPTY" || er.code === "EBUSY" || er.code === "EPERM" )) {
-          // Retry on windows, sometimes it takes a little time before all the files in the directory
-          // are gone
           if (Date.now() - start > 1000) throw er;
-        } else if(er.code === "ENOENT") {
+        } else if (er.code === "ENOENT") {
           break;
         } else {
           throw er;
--- a/src/rm.js
+++ b/src/rm.js
@@ -53,11 +53,14 @@ function rmdirSyncRecursive(dir, force) {
     while (true) {
       try {
         result = fs.rmdirSync(dir);
+        if (fs.existsSync(dir)) throw { code: "EEXIST" }
         break;
       } catch(er) {
-        if (process.platform === "win32" && (er.code === "ENOTEMPTY" || er.code === "EBUSY" || er.code === "EPERM" )) {
+        // In addition to error codes, also check if the directory still exists and loop again if true
+        if (process.platform === "win32" && (er.code === "ENOTEMPTY" || er.code === "EBUSY" || er.code === "EPERM" || er.code === "EEXIST")) {
           if (Date.now() - start > 1000) throw er;
         } else if (er.code === "ENOENT") {
+          // Directory did not exist, deletion was successful
           break;
         } else {
           throw er;
--- a/src/rm.js
+++ b/src/rm.js
@@ -48,7 +48,23 @@ function rmdirSyncRecursive(dir, force) {
 
   var result;
   try {
-    result = fs.rmdirSync(dir);
+    var start = Date.now();
+    while (true) {
+      try {
+        result = fs.rmdirSync(dir);
+        break;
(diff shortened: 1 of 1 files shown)


andreialecu:
I have an idea on how to make a test for this using https://github.com/cjohansen/Sinon.JS by stubbing `fs.rmdirSync` to return that error on first call, and process to return `win32` but I don't know if you want to take a dependency for it. Thoughts?

arturadib:
Looks good, I don't think we need to a specific platform test for this.. I need a second pair of eyes though - @Schoonology ?

Schoonology:
Looking over that issue again, I wouldn't retry on `EPERM` - getting that error is the best way to know you've got multiple processes contending for those files.

Schoonology:
I've left all the comments I can think of for now. Thoughts, rebuttals, etc.? Though this is a relatively small amount of code, I feel like `rm` is somewhat important to get right...

andreialecu:
@Schoonology I fixed the style issues you pointed out and re-pushed. 

Regarding EPERM: It was one of the errors I was getting for which retry helped. For the EPERM case, it could be an antivirus related issue. I don't think 1 second of retrying is that much of a big deal if the file is actually in use.

Regarding checking for win32: on other plaforms the files should disappear even if they're in use, it's only Windows that locks files from being deleted under such circumstances. If an error occurs on Linux for instance, it's probably because of permissions, not because of quirks such as this one. (at least - I couldn't reproduce this error on Linux and MacOSX, it works fine there)

andreialecu:
There's one more issue here: In my usage, after doing a rm -rf, I'm recreating the same directory I deleted.

This sometimes results in an error in mkdir because the directory **didn't disappear yet**. There needs to be an additional check on Windows whether the directory is gone before returning from the function. There won't be any error in this case in rmdirSync and nothing logged, but it just takes one extra milisecond for the directory to actually disappear from the file system.

To reiterate, shelljs is being used as part of a test suite for a library I'm working on and this was the final fix to all of the shelljs related problems I was getting running this test suite of 100+ tests on Windows.

andreialecu:
@Schoonology need your input on https://github.com/arturadib/shelljs/pull/179/files#diff-d0eda960c284276e6b0b269826fec539R56 

I thought this was the simplest, I don't think that line will ever throw on non-Windows, but if it ever does, I think it's a good thing. I assume the calling application would want to know an allegedly successful call to remove a directory somehow still left the directory behind. Thoughts?

src/rm.js
```diff
@@ -48,7 +48,25 @@ function rmdirSyncRecursive(dir, force) {
 
   var result;
   try {
-    result = fs.rmdirSync(dir);
+    // Retry on windows, sometimes it takes a little time before all the files in the directory are gone
+    var start = Date.now();
+    while (true) {
+      try {
+        result = fs.rmdirSync(dir);
+        if (fs.existsSync(dir)) throw { code: "EEXIST" }
```
Schoonology: I think this is fine, though since it's potentially user-facing, I'd like to invest a little more time in the thrown Error:

1. Throw an actual `Error`. No base class needed.
2. `EEXIST` is specific to creating files. Could we use a more generic error like `EAGAIN` or even `EOTHER`?
3. A good message.

@arturadib - Do we have a precedent for these kinds of errors?

arturadib: I think we use `common.error()` for error messages

andreialecu: Let me know what you want to do about it and I'll change it. 

@Schoonology Regarding 1 and 2: I'm not sure how you would concatenate a good message with an actual `Error` without creating a new `Error` class. Could you clarify?

@arturadib The code flow goes to the outer `catch` which does use `common.error()`. That throw there is just to give it a chance to retry. It's handled by the inner `catch` in the `if`. If the error is re-thrown, the outer `catch` handles it normally.

Schoonology: @andreialecu - I _completely_ missed that outer `catch`. Thanks for being patient with me, and pointing that out. That solves 1 and 3, and I'd just replace 2 with `EAGAIN`, because trying again _may_ succeed, whereas `EEXIST` is meant to indicate that creating a new file failed because one already exists, and the user _should not_ try again until that's resolved.

Does that make sense?

andreialecu: Hey @Schoonology, I changed it to EAGAIN. Check it out now.

Schoonology:
I'm good with this, and I think @arturadib's concerns are satisfied with that outer `catch`, just as you pointed out.

Squash it and :shipit:.

waddlesplash:
*bump*. It'd be nice if this could get merged, I'm running into it a lot on Windows...

arturadib:
changes published to version 0.5.1, thanks everyone!

sparkleholic:
Coool ! :+1:

PR #185, opened 2015-02-02
Update exec async doc

vise890:
The [exec docs](http://documentup.com/arturadib/shelljs#command-reference/exec-command-options-callback) specify:
> **exec(command [, options] [, callback])**
> Available options (all false by default):
> `async`: Asynchronous execution. Defaults to true if a callback is provided.
> `silent`: Do not echo program output to console.

That seems to imply to me that the following code can be written:
```javascript
exec("sleep 0.5", {async: false}, function() {
    echo("print first");
});
echo("print last");
```
which prints:
```
print last
print first
```

After looking at the code, however, it is clear that `options.async` is overridden to `true` if a callback is provided. This PR updates the README to be more in line with the code. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,7 +465,8 @@ Object containing environment variables (both getter and setter). Shortcut to pr
 ### exec(command [, options] [, callback])
 Available options (all `false` by default):
 
-+ `async`: Asynchronous execution. Defaults to true if a callback is provided.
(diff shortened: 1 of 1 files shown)


nfischer:
@vise890 This looks good! Could you make the change instead in the comment section and use `generate-docs`?

vise890:
Implemented in #289, closing

PR #187, opened 2015-02-11
If rename fails w/ EXDEV attempt cp & rm instead

### LINKED ISSUE #1: #mv Won't Work Across Disks
[`fs#renameSync`](https://github.com/arturadib/shelljs/blob/master/shell.js#L388) won't work across partitions or attached storage. This is by design, as `rename` is paired to the C version of the same method, which isn't able to cross disks either. (See [this closed issue](https://github.com/joyent/node/issues/2703) for more discussion about `rename`.)

A solution could be piping a `readStream` to a `writeStream`, and then cleaning up the source file.  
I'd attach a fix but the trouble is that listening for the `end` event on the pipe and firing a callback would break the synchronous nature of shelljs.

Ideas?

*Great module, by the way. Very helpful!*

tivac:
Fixes #1

Uses suggestions from https://github.com/joyent/node/issues/2703. Not sure how kosher it is to include other tasks directly like this, or to pass the options object through to them like this. Feel free to suggest alternatives! `mv`, `cp`, and `rm`, all share the `-f` option so it seemed safe.

Tests for `mv.js` still passing for me locally. A bunch of other tests are failing but I suspect that's because I am on windows.

### DIFF
--- a/src/mv.js
+++ b/src/mv.js
@@ -1,6 +1,8 @@
 var fs = require('fs');
 var path = require('path');
 var common = require('./common');
+var cp = require('./cp');
+var rm = require('./rm');
 
 //@
 //@ ### mv(source [, source ...], dest')
@@ -74,7 +76,18 @@ function _mv(options, sources, dest) {
       return; // skip file
     }
 
-    fs.renameSync(src, thisDest);
+    // Wrapped because cross-device renames will throw
+    try {
+      fs.renameSync(src, thisDest);
+    } catch(e) {
+      if(e.code !== 'EXDEV') {
+        throw e;
+      }
+
+      // Can't rename across devices, but you can copy & then unlink!
+      cp(options, src, thisDest);
+      rm(options, src);
+    }
   }); // forEach(src)
 } // mv
 module.exports = _mv;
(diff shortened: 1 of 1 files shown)


tivac:
Actually, this has weird interactions with trying to mv an entire directory at once. `cp` requires you to pass `-r` for that to work like you'd expect, so it looks like I'll need to think about this one a bit more.

nfischer:
@tivac any more work on this? I originally wrote up something similar to what you have, but decided it was unfavorable, since any errors during the `cp` operation would output an error message prefixed by `cp:`, even though it's actually the `mv` command.

tivac:
@nfischer I haven't touched this in almost a year, I worked around the problem by changing how we moved files around IIRC.

I still think this should be fixed, but ¯\\\_(ツ)\_/¯

nfischer:
Ah, I figured that was the case. I agree, this probably should be fixed. I think to fix it cleanly though, we'd have to do a lot of internal refactoring, and I'm hesitant to write up a hack.

Thanks for the PR though! I'll see if I can figure out an elegant solution.

PR #190, opened 2015-02-23
Update .travis.yml

arturadib:


devTristan:
does not pass, what's this extra linebreak all about?

also confused about why this is missing iojs and tired of hearing excuses.

arturadib:
sorry man, been swamped! thanks for merging it

PR #192, opened 2015-03-05
Added -n to cp

SerenadeX:


SerenadeX:
The -n option allows the user to copy things from one directory to another without overwriting existing files in the destination

arturadib:
hey there- sorry I'm late to this. mind modifying the test to actually check the file contents to ensure the file wasn't overwritten? thanks!

nfischer:
@SerenadeX would you be able to address the comment above from @arturadib? If so, I'd be interested in merging!

Also, would you be able to rebase off master?

nfischer:
@SerenadeX Thank you so much for the PR! I think we'll have to close this, however, since this is replaced by #336. Thanks for the idea!

PR #194, opened 2015-03-13
global returns shell so we wont require twice

wildeyes:
The rationale behind this, is that if we want to use the names of the functions exported by shelljs, we would either have to do an ugly swap, or we would have to require shelljs twice.

I say just return the shell instance and use it if/when the time needs - I think that's what the user expects of the require.

### DIFF
--- a/global.js
+++ b/global.js
@@ -1,3 +1,4 @@
 var shell = require('./shell.js');
 for (var cmd in shell)
   global[cmd] = shell[cmd];
+module.exports = shell;


ariporad:
@wildeyes: Thanks so much! Could you please add a test, then we can merge?

nfischer:
What is the use case for saying `var shell = require('shelljs/global');`? The global require already pollutes the global namespace, so I don't see an advantage to having a `shell` object lying around. If this could be justified, I'd be fine with merging.

nfischer:
@wildeyes do you mean, for example, if you wanted to use shelljs but wanted to write your own function named `echo()`? So you would define your own `echo()` function and then use `shell.echo()` when you wanted shelljs's echo?

ariporad:
@nfischer: Yeah, I'm pretty sure that's what he meant. (It's what I was 
thinking
at least)

nfischer:
The only concern I guess is that using shelljs in the commandline is slightly less pretty:

```Javascript
$ node
> require('shelljs/global'); // used to not display return value, now it does
{ cd: [Function],
    pwd: [Function],
    ls: [Function],
    find: [Function],
    cp: [Function],
    rm: [Function],
...
```

This doesn't change behavior, so it isn't a big concern for me.

ariporad:
@nfischer: I think this is fine, but if you're really opposed, I don't think it's that big either way.

LGTM, but I'll leave it to you to merge/not-merge it.

ariporad:
@nfischer: Thoughts?

nfischer:
Sorry, forgot to get back to it. My concern is that users might see `var shell = require('shelljs/global')` and assume that it doesn't affect the global namespace since it returns an object. I think having 2 requires isn't too difficult, and will be less confusing. Unless there's some reason that workaround won't work.

ariporad:
Ok. Closing.

ariporad:
Ok, Closing.

PR #200, opened 2015-04-04
Documentation: Add an index to the reference functions for easy navigation

moatazelmasry2:


arturadib:
hi- have you seen shelljs.org yet? it has an index on the sidebar on the right.. please reopen if I missed the point

moatazelmasry2:
Hi Arturadib. No you didn't miss the point. But I wwanted to have something similar as well in Github to ease navigation. Similar to https://github.com/petkaantonov/bluebird/blob/master/API.md

PR #207, opened 2015-05-15
Breaking: Allow -- as args separators (fixes #188)

### LINKED ISSUE #188: Pass command line params to the make tool
I'm using the make tool, and I'd like to be able to pass in command line arguments to it. I tried this:

```
$ node Makefile.js client-test -- foo.js
no such target: --
no such target: foo
```

It seems that all parameters are treated as potential targets. Would it be possible to use `--` as a delimiter to pass in command line arguments?

nzakas:
This implements a change that allows `--` to indicate arguments to be passed to `make` targets. Basically, you can do this:

```js
target.foo = function(args) {

};
```

And `args` will be everything after the `--` on the command line, so for example:

```
$ node Makefile.js foo -- bar baz
```

`args` in the previous code will be `['bar', 'baz']`.

I noticed there's a `force` argument for targets, but I couldn't find anywhere it's documented or used. So, I moved it to be the second argument as it seems mostly unused at the moment.

Other caveats:

1. It appears there aren't any tests for `make.js`, so I didn't know if it's okay to just make changes. I did test it locally with an example file and it worked
1. `npm test` fails for me (I'm on Windows using Git Bash) and I couldn't figure out how to get it to run, so I'm not really sure about this change.

@arturadib let me know if this is okay or if you'd like other changes.

### DIFF
--- a/make.js
+++ b/make.js
@@ -3,9 +3,18 @@ require('./global');
 global.config.fatal = true;
 global.target = {};
 
+var args = process.argv.slice(2),
+  targetArgs,
+  dashesLoc = args.indexOf('--');
+
+// split args, everything after -- if only for targets
+if (dashesLoc > -1) {
+  targetArgs = args.slice(dashesLoc + 1, args.length);
+  args = args.slice(0, dashesLoc);
+}
+
 // This ensures we only execute the script targets after the entire script has
 // been evaluated
-var args = process.argv.slice(2);
 setTimeout(function() {
   var t;
 
@@ -21,8 +30,8 @@ setTimeout(function() {
     (function(t, oldTarget){
 
       // Wrap it
-      global.target[t] = function(force) {
-        if (oldTarget.done && !force)
+      global.target[t] = function() {
+        if (oldTarget.done)
           return;
         oldTarget.done = true;
         return oldTarget.apply(oldTarget, arguments);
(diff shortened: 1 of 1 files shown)


make.js
```diff
@@ -5,7 +5,16 @@ global.target = {};
 
 // This ensures we only execute the script targets after the entire script has
```
arturadib: this comment was meant to go right above the `setTimeout` below, otherwise it's puzzling :)

make.js
```diff
@@ -21,7 +30,7 @@ setTimeout(function() {
     (function(t, oldTarget){
 
       // Wrap it
-      global.target[t] = function(force) {
+      global.target[t] = function(localArgs, force) {
```
arturadib: I frankly can't remember why I put `force` in here.. it's not even documented :) maybe remove it?

arturadib:
OK, so it just occurred to me that some folks (including one of our largest users, PDF.js) pass arguments directly to targets from within the script itself, see e.g. https://github.com/mozilla/pdf.js/blob/master/make.js#L473.

I don't think this would break things for them unless they started using the new `--` feature, but just wanted to give a heads-up in case I'm missing something. /cc @yurydelendik @brendandahl 

If the PDF.js folks are OK with this let's :sheep: it!

nzakas:
It would break right now because I'm always passing an empty array if `--` isn't used. If instead I passed `undefined`, then it seems like it would be backwards compatible.

make.js
nzakas: Oops!

nzakas:
That's not how it works. The `--` is separate, such as:

```
$ node Makefile.js test -- tests/lib/too.js tests/lib/bar.js
```

arturadib:
> If instead I passed undefined, then it seems like it would be backwards compatible.

Cool, can we try that?

nzakas:
Updated.

arturadib:
Merging this in. @brendandahl @yurydelendik please let me know if this breaks the build for PDF.js!

arturadib:
this is now in `shelljs@0.5.0`

nzakas:
Yay, thanks!

PR #216, opened 2015-07-06
Memoized the result of target invocation

rizowski:
For a current make script I was unable to use promises as the task had previously ran.
This commit it makes it so the targets still return a result even if they have been ran. This makes it so that the promises can still be used.

### DIFF
--- a/make.js
+++ b/make.js
@@ -31,10 +31,11 @@ setTimeout(function() {
 
       // Wrap it
       global.target[t] = function() {
-        if (oldTarget.done)
-          return;
-        oldTarget.done = true;
-        return oldTarget.apply(oldTarget, arguments);
+        if (!oldTarget.done){
+          oldTarget.done = true;
+          oldTarget.result = oldTarget.apply(oldTarget, arguments);
+        }
+        return oldTarget.result;
       };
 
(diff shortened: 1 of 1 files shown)


mtscout6:
:+1:

rizowski:
Looks like it is failing on node version 12 but failing to an unrelated change. I don't have permission to rerun the tests.

mtscout6:
@arturadib Any progress on this?

danielepolencic:
This is breaking existing functionality for me:

I used to have something like:

```js
target.build = function () { /* write to a file */ }
target.watch = function () {
  target.build.call(null);
}
```

With this change `target.build` is executed only once no matter how many times I call it from `target.watch`.

mtscout6:
This would seem on par with how makefiles work. In general Makefiles only execute each target once. If you wish to invoke a function multiple times then it would be better served as a normal function that is not one of the build targets.

An alternative you could use is:

```
function buildAction() { /* write to a file */ }
target.build = function () { buildAction(); }
target.watch = function () {
  onFileChange('*.txt', buildAction);
}
```

arturadib:
spot-on @mtscout6 thanks!

PR #222, opened 2015-08-11
Manually closing streams

JulianLaval:
Solves the stream closing issue introduced by pull [#214](https://github.com/arturadib/shelljs/pull/214).

According to the [Node Docs](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options), `process.stderr` and `process.stdout` are closed when the process exits - we can listen to this and close the `stdoutStream` accordingly

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -62,7 +62,9 @@ function execSync(cmd, opts) {
       "childProcess.stdout.pipe(stdoutStream, {end: false});",
       "childProcess.stderr.pipe(stdoutStream, {end: false});",
       "childProcess.stdout.pipe(process.stdout);",
-      "childProcess.stderr.pipe(process.stderr);"
+      "childProcess.stderr.pipe(process.stderr);",
+      "process.stdout.on('end', function(){ stdoutStream.end(); });",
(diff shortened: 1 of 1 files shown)


devTristan:
I fear this could reintroduce the bug that #214 fixed.

Perhaps try something like this:

```javascript
var stdoutEnded = false, stderrEnded = false;
function tryClosing() {
  if (stdoutEnded && stderrEnded) {
    stdoutStream.end();
  }
}
childProcess.stdout.on('end', function() {
  stdoutEnded = true;
  tryClosing();
});
childProcess.stderr.on('end', function() {
  stderrEnded = true;
  tryClosing();
});
```

JulianLaval:
@devTristan, whilst this did not explicitly reintroduce the bug fixed by #214, it was indeed misguided in its reliance on `process.stdout`.

Your solution is much more elegant and relevant; applying your fix + outputting to console at `stdStream.end()` seems to confirm the validity of this fix in the context of chained `git` commands.

Pull request updated accordingly!

arturadib:
thanks @JulianLaval and @devTristan !

arturadib:
this is now published in v0.5.3

PR #227, opened 2015-08-26
Fixes an issue with multi-symbolic mode specification (ie a-rwx,u+rw)

rezonant:
If you call:

shelljs.chmod('a-rwx,u+rw', 'file');
shelljs.chmod('a-rwx,u+rw', 'file');

Then the file will have no permissions (mode 000). This is because the permissions at the beginning of the second call are 600, and when the 'u+rw' is processed, it compares the final mode (600) with the original mode (now outdated due to 'a-rwx', but 600) and does not run the operation if they are the same.

This PR adds a few more multi-symbolic tests, and resets 'perms' to 'newPerms' after running fs.chmodSync() to allow examples like 'a-rwx,u+rw' to work correctly even if the original file permission was already set as desired.

### DIFF
--- a/src/chmod.js
+++ b/src/chmod.js
@@ -185,7 +185,7 @@ function _chmod(options, mode, filePattern) {
               log(file + ' -> ' + newPerms.toString(8));
             }
             fs.chmodSync(file, newPerms);
-            //perms = newPerms; // for the next round of changes!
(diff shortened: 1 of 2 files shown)


arturadib:
awesome thank you!

PR #228, opened 2015-08-26
feat-multisymbolic + Support for directory entry (capital X in chmod terms)

rezonant:
Includes bugfix found in feat-multisymbolic. Previously capital X was ignored, now it adds execute (entry) permission to any matched directory while leaving files as they are.


arturadib:
sweet! thank you

PR #234, opened 2015-09-12
ln: add possibility creating relative links

a-x-:
hi,
in some cases relative links are prefer.
but now `ln.js` is not provide the feature.

my case: `ln('config/dev', 'config/current')` and root dir of the `config` dir may be move on fs.

so, absolute links are not broken, too. using:
`ln('/var/www', '/home/www')`

excuse my beginner's English

### DIFF
--- a/src/ln.js
+++ b/src/ln.js
@@ -29,10 +29,10 @@ function _ln(options, source, dest) {
     common.error('Missing <source> and/or <dest>');
   }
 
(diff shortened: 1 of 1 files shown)


nfischer:
@a-x- thanks for the PR! I think this will be handled by #282. If that PR doesn't cover this case, please ping this thread and let us know and we can take another look.

a-x-:
ok, i trust you)

PR #236, opened 2015-09-24
Add kill fn to shelljs to stop a child_process created by an async exec

quietshu:
After executing some async commands, I "lost" them cause I cannot stop them later. So I add a ChildProcess `kill` wrapper function.

```javascript
var shell = require('shelljs');

// An async process
var nodeProcess = shell.exec('node', { async: true });

// ...do some things

// In event triggers or conditions
if (needToKillNode)
    shell.kill(nodeProcess, function (code, signal) {
        // The stop callback
        console.log('node process was killed successfully');
    });
```

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -103,6 +103,10 @@ exports.env = process.env;
 var _exec = require('./src/exec');
 exports.exec = common.wrap('exec', _exec, {notUnix:true});
 
+//@include ./src/kill
+var _kill = require('./src/kill');
+exports.kill = common.wrap('kill', _kill);
+
 //@include ./src/chmod
 var _chmod = require('./src/chmod');
 exports.chmod = common.wrap('chmod', _chmod);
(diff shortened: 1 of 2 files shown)


mtscout6:
Something to think about with kill is that from the cli kill command you can specify which signal to send to the process in question. For example the SIGINT, SIGKILL, SIGHUP, or SIGSTOP signals (There may be some I missed). 

As far as being able to do this with the process returned from `shell.exec` I can't immediate tell by looking at the exec script here but Node does provide a kill function which does this. See: https://nodejs.org/api/child_process.html#child_process_child_kill_signal

Another concern to address is that one of the goals of this library is to ensure that the functionality works across operating systems. So on Linux, Mac, and Windows. I can't say that I know a proper cross platform solution that would work off the top of my head. http://stackoverflow.com/questions/140111/sending-an-arbitrary-signal-in-windows

@arturadib May have other things to say about it though.

arturadib:
agreed with @mtscout6 - the goal is to stay as close as possible to *nix. this seems to only provide a way to kill processes spun up by the script correct? *nix's kill lets you kill other processes too.

to @mtscout6 's point, I'd also prefer to think through how the feature can be cross-platform. closing for now, but please reopen if you address these concerns!

Issue #238, opened 2015-10-05
"Which" command not working properly on Windows Platform.


TingluoHuang:
Install node.js on a windows box, try use shell.which to find the location of NPM.
Shell.which('npm') return "C:\Program Files\nodejs\npm" instead of the correct "C:\Program Files\nodejs\npm.cmd"
The bug is at which.js line 49.

nfischer:
I believe another related issue is that the priority of `.bat`, `.cmd`, and `.exe` is currently incorrect. Seem [my comment](https://github.com/shelljs/shelljs/pull/239#issuecomment-169908832) on the PR for details.

nfischer:
Closing since this is fixed in #239

PR #239, opened 2015-10-05
Make sure Which() on Windows platform always return the command with …

TingluoHuang:
…extension.

### DIFF
--- a/src/which.js
+++ b/src/which.js
@@ -44,19 +44,10 @@ function _which(options, cmd) {
         return; // already found it
 
       var attempt = path.resolve(dir + '/' + cmd);
-      if (checkPath(attempt)) {
-        where = attempt;
-        return;
-      }
 
       if (common.platform === 'win') {
         var baseAttempt = attempt;
         attempt = baseAttempt + '.exe';
-        if (checkPath(attempt)) {
-          where = attempt;
-          return;
-        }
-        attempt = baseAttempt + '.cmd';
         if (checkPath(attempt)) {
           where = attempt;
           return;
@@ -66,7 +57,17 @@ function _which(options, cmd) {
           where = attempt;
           return;
         }
-      } // if 'win'
+        attempt = baseAttempt + '.cmd';
+        if (checkPath(attempt)) {
+          where = attempt;
+          return;
+        }
+      } else {
+        if (checkPath(attempt)) {
+          where = attempt;
+          return;
+        }
+      }
     });
   }
 


TingluoHuang:
For issue #238.

mtscout6:
This looks reasonable to me @arturadib thoughts?

arturadib:
:+1:

nfischer:
It looks like the priority of `.bat`, `.cmd`, and `.exe` is not correct here.

I tested this on my Windows 7 64-bit machine as follows: I created files in my path named `test.bat`, `test.cmd`, and `test.exe`. Each file performs a simple print statement to help distinguish it from the other files. Here is what I saw:

```
> test # calls test.exe, not the others
this is exe
> del test.exe
> test # since there's no .exe, it calls .bat
this is bat
> del  test.bat
> test # since there's no .exe or .bat, it calls .cmd
this is cmd
```

So I think that `which()` should check for extensions in the order:

 1. `.exe`
 2. `.bat`
 3. `.cmd`

@TingluoHuang would you be able to add this to your PR? Aside from that, this looks ready to merge.

TingluoHuang:
@nfischer 
attempt = baseAttempt + '.exe';
attempt = baseAttempt + '.cmd';
attempt = baseAttempt + '.bat';
The current priority in the code is already this one you want, isn't it?

nfischer:
I believe that priority is `.exe`, `.cmd`, `.bat` (which is coincidentally reverse alphabetical order), and I believe the correct priority is `.exe`, `.bat`, `.cmd` (which is not reverse alphabetical order). Let me know if I'm missing something.

nfischer:
@TingluoHuang can you verify that the priority is as I described? If so, and we get that fixed in the code, then I think this is good to merge.

TingluoHuang:
@nfischer The priority you mentioned is correct, i update the PR.

nfischer:
@TingluoHuang Can you rebase off master and squash your commits? I think this is safe to merge after that (if it passes the CI).

Once we get Windows CI running, we should probably add a specific test case for this, but we can do that later.

nfischer:
LGTM. @arturadib @ariporad Feel free to merge

ariporad:
:+1:

BYK:
Windows has a PATHEXT environment variable that tells which files are "self executable". Instead of a fixed list of `.exe`, `.cmd` etc we should probably use that. Similar code in Python: https://github.com/scopatz/xonsh/blob/57e7c1f43879e6394342a17638f2a9b6cb80bc23/xonsh/built_ins.py#L284-L298

BYK:
FYI I'm willing to submit a PR that does this if you agree :)

nfischer:
@BYK would #134 be along the lines of what you're talking about? Perhaps we should ping that thread and see if that could get rebased off master.

I wasn't aware of the variable, but it seems like it's the right way to go. Thanks for bringing it to my attention!

BYK:
@nfischer no problem :) Also reviewed the new patch!

PR #240, opened 2015-10-06
Fix `os.tmpdir` bug

BYK:
The checked method on `os` module was `tempDir` which was not existing on any version of Node (should have been `tmpDir`). With Node 0.10, the name became `tmpdir`.

### DIFF
--- a/src/tempdir.js
+++ b/src/tempdir.js
@@ -37,7 +37,8 @@ function _tempDir() {
   if (state.tempDir)
     return state.tempDir; // from cache
 
-  state.tempDir = writeableDir(os.tempDir && os.tempDir()) || // node 0.8+
+  state.tempDir = writeableDir(os.tmpdir && os.tmpdir()) || // node 0.10+
+                  writeableDir(os.tmpDir && os.tmpDir()) || // node 0.8+
                   writeableDir(process.env['TMPDIR']) ||
                   writeableDir(process.env['TEMP']) ||
                   writeableDir(process.env['TMP']) ||


nfischer:
I agree, it looks like it should say `tmpDir`, not `tempDir`. With `v5.3.0` I can see both `tmpDir` and `tmpdir`. Is there a difference in this case?

Could you provide an example of buggy behavior and the expected behavior?

BYK:
@nfischer - you can check the ticket on ESLint that refers to this one: https://github.com/eslint/eslint/issues/4061

nfischer:
After looking into things a bit, I agree with the change of `os.tempDir` -> `os.tmpDir`. However, I think that `os.tmpDir` is exactly equivalent to `os.tmpdir` (see [the current source code](https://github.com/nodejs/node/blob/master/lib/os.js#L47), [the archived source](https://github.com/nodejs/node-v0.x-archive/blob/master/lib/os.js#L60), and the [stack overflow post](https://stackoverflow.com/questions/23494354/whats-the-difference-between-os-tmpdir-os-tmpdir) that lead me there).

Unless there's discussion of getting rid of `os.tmpDir` in favor of `os.tmpdir`, we should probably only keep `os.tmpDir` (if this name is deprecated though, I think it's better to include both).

Other than that, this looks like something we would merge fairly soon.

BYK:
@nfischer I also double checked from my console in Node 4 and `os.tmpdir === os.tmpDir` returns `true.` That said the docs make no mention of this: https://nodejs.org/api/os.html#os_os_tmpdir

So my suggestion is either drop the undocumented `tmpDir` version and also drop Node 0.8 support which is quite old, or keep it as it is for the sake of clarity. (we may add a comment linking back to this discussion).

nfischer:
Ok, you raise a good point about it being undocumented. Since this operation should only be performed once (and the result will be cached), it shouldn't bloat things too much. A comment linking back might be a good (plus, it'll trigger the CI so we can verify that it doesn't break things).

LGTM, once the CI passes

nfischer:
@ariporad what are your thoughts?

ariporad:
@nfischer: I'm not sure I understand this 100%, but I think that the best thing to do would be `(os.tempDir || os.tmpDir || os.tmpdir)(...)`.

nfischer:
@ariporad `os.tempDir` doesn't actually exist (it was a typo I think).

You suggestion works if at least one of the three functions exists. The issue is that early versions (before 0.8) don't have any of those functions. So if `os.tempDir` and `os.tmpDir` don't exist, it'll **always** call `os.tmpdir()` (which doesn't always exist).

If we do it in multiple lines as it is, it will never call the function unless it actually exists (so the code works as expected).

As this PR is written, it fixes the typo (which is good), and it prefers the newer (and actually documented) name `os.tmpdir` over `os.tmpDir`, which is probably a safer bet in the long run.

ariporad:
@nfischer: Then LGTM!

Issue #202, opened 2015-04-28
CLI Version


tobiastom:
It might be nice to add a CLI version of the commands. This would allow people to use the commands directly inside the `package.json`.

For example:

```
   …
    "scripts": {
        "copy": "shelljs cp source/* desk"
    }
    …
```

This would make the cp task available which then could be used in every operating system, no matter if it is included or not. 

Right now I would need to add a separate tool for each job (for example [cpy](https://www.npmjs.com/package/cpy)).

seangenabe:
+1 I kinda need this too. Currently I'm depending on cp (assuming msysgit is installed). I'm planning on making my own module, but this one seems a better place for it to be in.

anselmh:
That would indeed be awesome. Any plans of doing that, @arturadib?

arturadib:
No plans, could either be a separate package or if someone wants to take a stab at it and send a PR I'm happy to take a look, thanks!

codemakerlive:
There's this module (which is currently on shelljs 0.2.6): https://github.com/nzakas/shelljs-nodecli. I didn't author it but if you could contact Nicholas perhaps the 2 projects could be merged.

dandv:
That would be *so* awesome for [using npm as a build tool](http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool) instead of grunt, gulp etc.

keithamus:
It'd be great to see shelljs have a set of binaries matching all of the available commands - not just for windows compatibility, but for better compatibility between Linuxes and BSDs

nickdima:
This is pretty trivial to implement, but not sure how to keep compatibility with the current behaviour of the `shjs` CLI. I'm talking about `shjs script_name` vs `shjs command_name`.
Maybe we could have it as a separate CLI script? It would need to have a short name also.
Or make a breaking change to the way it currently works?

anselmh:
Hm, that would depend on what you want to do. If you don’t care about the old style, make it a new major version and breaking change. If you do care about it, I think a separate module would do the trick.

keithamus:
@nickdima for every command you could create a simple binary wrapper, for example:

```js
// File: bin/cp
#!/usr/bin/env node
require('shelljs/global');
cp.apply(null, process.argv.slice(2));
```

```js
// File: bin/mkdir
#!/usr/bin/env node
require('shelljs/global');
mkdir.apply(null, process.argv.slice(2));
```

[...]

PR #244, opened 2015-10-19
Add new CLI script for running commands

nickdima:
I've added a new CLI script called `shl` for running one-off commands as discussed in #202.
This is more a proposal and we can polish it some more before merging as I have the impression that a second CLI script will add some confusion.
To use the script you simple pass the command name as the first argument and any subsequent arguments will be passed to the command itself. Eg.:

`shl rm -rf some_dir`

We also need to figure out the best way to print the output of commands like `ls` which in this case returns an array.

### DIFF
--- a/bin/shl
+++ b/bin/shl
@@ -1,0 +1,19 @@
+#!/usr/bin/env node
+var shell = require('../shell');
+
+if (process.argv.length < 3) {
+  console.log('ShellJS: missing argument (command name)');
+  console.log();
+  process.exit(1);
+}
+
+var command = process.argv[2];
+var args = process.argv.slice(3);
+
+var res = shell[command].apply(null, args);
+if (res) {
+  console.log(res);
+}
+if (shell.error()) {
+  process.exit(1);
+}
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "test": "node scripts/run-tests"
   },
   "bin": {
-    "shjs": "./bin/shjs"
+    "shjs": "./bin/shjs",
+    "shl": "./bin/shl"
   },
   "dependencies": {},
   "devDependencies": {


TimothyGu:
How about `shjs -c`? `shl` is kind of far from shelljs and is too similar IMO to the SHift Left instruction.

nickdima:
Yeah, you're right. My goal was to make it work without too much typing for providing a good alternative to bash commands in npm "inline" scripts where you often chain multiple commands together.
Let's ask some other people also cc @arturadib @keithamus @anselmh @dandv @tobiastom

tobiastom:
I actually don't really have any preference on any approach. I think the `shl` one is clearer because the options would not get mixed in execution (e.h. is the `-c` for `shjs` or `rm`)

Thanks @nickdima for working on this. Really looking forward to this!

ariporad:
@nfischer: Thoughts?

bin/shl
```diff
@@ -0,0 +1,19 @@
...
+
+var command = process.argv[2];
+var args = process.argv.slice(3);
+
+var res = shell[command].apply(null, args);
+if (res) {
+  console.log(res);
+}
+if (shell.error()) {
+  process.exit(1);
```
nfischer: Once #269 is merged, let's use the error code it returns

nfischer:
@nickdima thanks! This looks great, and I'm sure a lot of people would appreciate this.

 1. Could you add a couple tests?
 2. Once #269 is merged (hopefully soon), could you address my comment above?

@ariporad what are your thoughts on `shl` vs `shjs -c` vs something else?

TimothyGu:
@tobiastom said:

> I think the `shl` one is clearer because the options would not get mixed in execution (e.h. is the `-c` for `shjs` or `rm`)

Valid concern, but if we are to stick with the semantics of `sh` it should be clear from the position of `-c` that it describes the one that precedes that.

nickdima:
Wow, I did this 4 months ago so I had to get back into context :)
Thanks for the suggestions @nfischer 
So... `shl` or `shjs -c` ?

nfischer:
@nickdima my vote would have to be for `shjs -c`. It most closely follows `bash -c`, `python -c`, etc. Let's see what @ariporad's thoughts are though.

And sorry about the delay! We'll be trying to get back on top of the PRs and issues. Thank you for your contribution!

ariporad:
I like -c. It makes the most sense to me.
We could also consider -e, like node -e

nfischer:
@nickdima would you be able to refactor this into the `shjs` script, and have it be activated by the `-c` flag (or `-e`, if you prefer)?

nfischer:
I suppose the only concern is if users get confused by whether to write:

```Bash
$ shjs -c rm -rf foo
```

or

```
$ shjs -c "rm('-rf', 'foo')"
```

ariporad:
Well, the second would be correct. I guess just throw an error for the 
first?

nfischer:
@ariporad the way this is written, the first would be correct actually. The second is not supported (hence why I think it's slightly confusing).

ariporad:
@nfischer: Hmm...

That actually makes me less sure if we should do this. Now that I think about it, I'm not sure if we should be encouraging embedding shell scripts as js inside package.json. And the way it's implemented now is very confusing. (Its just an incomplete, non-compliant implementation of bash).

So I'm not sure about this, maybe modules like 'mkdirp' or 'rimraf' are just the way to go.

nfischer:
@ariporad this comes from #202, which seems like a popular issue. It sounds like this is along the lines of what users want.

I don't think the bash syntax is necessary, but it does make porting extremely easy, I suppose.

ariporad:
@nfischer: Hmm...
What are your thoughts on pulling it in to it's own module, which could depend on shelljs? (It could be under the shelljs org). I just feel like it's really out of scope.

nfischer:
@ariporad I'm ok with making it its own module.

ariporad:
@nfischer: Sounds great!
@nickdima: Would you mind if I re-packaged this into its own module?

Everyone, what should we call it? I like `jsh`, which can either stand for **J*avascript **Sh**ell or it can be a contraction of JS and SH. Although that's not available on npm... Thoughts?

TimothyGu:
jsh is already used by other projects too: https://code.google.com/archive/p/jsh/ (which is also a "JavaScript SHell." How about `jsbox` (after `busybox`)? It has a similar syntax of `busybox <applet> [args...]`

radek-holy:
or just `shelljs-cli` :)

nickdima:
@ariporad I'm ok with you repackaging it as a new module, maybe it's better that way
let's make sure we keep the exposed bin script's name just a few letters

ariporad:
@ndelitski: Ok, Thanks!

Anyone have any naming ideas? (`sh-c`?)

nfischer:
I vote against `sh-c`. That's one space off from `sh -c`, which has a very different meaning

ariporad:
Ok everyone, I have a bad idea: What about `xec` or `xek`, short for exec. (Yes, I know both of those are terrible, but they're 3 letters, free on npm, and mostly describe what they do).

Actually, that gives me a better idea: What about `xsh`? (Also free on npm).

TimothyGu:
I actually like `shx`, which is also vacant on npm. Most UNIX shells end with sh (bash, dash, ash, fish, zsh, ksh, csh, tcsh, etc.), so `xsh` can be a bit confusing.

ariporad:
Sounds great to me, @nfischer?

nfischer:
I prefer `shx` to `xsh` for the reasons @TimothyGu noted. I'd still prefer a 4-letter name that has "js" somewhere in there (`xjsh` and `jshx` are available options along those lines). If a 3-letter name is a priority, then I'd vote for `shx`.

ariporad:
Ok, shx it is.

I'll implement tomorrow, first thing in the AM.


Ari

nfischer:
SGTM

ariporad:
It looks like this is blocked for #269.

nfischer:
@ariporad you can set up the repo now and get the basic utility built. Then error codes can be added soon after

ariporad:
Ok everyone, initial implementation in shelljs/shx. Feedback welcome!

nfischer:
If people are satisfied with the extra repo, we can probably close this issue. Thoughts?

tobiastom:
Seems to work really nice for me. Thank you very much for implementing this. 

One little side discovery: will the `shojis` binary will go away then in the future? I think it might be misleading to have two.

ariporad:
@tobiastom: We don't have a `shojis` binary :stuck_out_tongue_closed_eyes:, but as far as the `shjs` goes, I'm not sure... I don't really see a use for it. @nfischer?

ariporad:
Also, closing this.

ariporad:
For people who see this in the future, this has been moved to shelljs/shx#1. Please continue discussion there. (And I'm going to lock this thread).

nfischer:
@ariporad the `shjs` bin file is still useful. It executes shelljs scripts without the need for modifying `NODE_PATH`, which is convenient for some (I know I found it convenient before I realized I could modify that).

Issue #122, opened 2014-05-03
touch command


slang800:
I'd find this useful, and there's already a package for doing it here: https://github.com/isaacs/node-touch. It just needs to be fitted to work with the shelljs interface.

Thoughts?

n2liquid:
This is from 2014 and there's still no touch command? I'm not trying to be a prick, but how come this project doesn't have that?

slang800:
oh - I totally forgot about this issue.
@Schoonology - why don't you want deps?

Schoonology:
@arturadib can chime in with other reasons, but my biggest reason is keep the installation process of shelljs as simple, fast, and low-risk as possible.

I understand that `touch` is a relatively simple module in practice, but it brings along with it `nopt`, an option parser, because `touch` includes a `bin` script. It's well within the realm possibility that some native library is found to implement `nopt`'s philosophy, only faster. If `nopt` decides to use this library in lieu of parsing arguments in JavaScript, then we have a native dependency because we use `touch`.

Granted, the likelihood of this happening may be low, but situations like this happen all over Userland, and we have _a lot_ of users' expectations to manage. The most common use case seems to be normalizing a shell environment where little may be known about what other tools are available, and I'd rather be able to guarantee `shelljs` can install with nothing but `node`, not a compiler toolchain and Lord Knows What Else.

n2liquid:
If touch needs native deps, I think that's a good enough reason to leave it out. It's not like I can't install node-touch alongside shelljs if I need it anyway. They can be like Batman and Robin or something.

devTristan:
To add dependencies while (mostly) guaranteeing future stability, npm's [shrinkwrap](https://docs.npmjs.com/cli/shrinkwrap) feature can be employed. That way, any dependency (or sub dependency) changes must be signed off on and cannot be brought in automatically.

slang800:
touch doesn't need native deps... AFAIK, it's pure JS and so is nopt

arturadib:
it looks like `touch.js` from `node-touch` is merely ~150 lines of JS code, we could bring it over and adapt it to the way we handle arguments (so as to not use `nopt`).

I've been favoring this method of implementing features mostly because (like `touch`) most commands are only ~100 lines long, and so mapping our way of handling arguments and options to a library's isn't worth the trouble.

@n2liquid you're welcome to send a pull request! manda brasa, valeu :)

slang800:
[...]

PR #249, opened 2015-11-05
add touch(1)

blockloop:
Add the `touch` command based on the [manpage](http://linux.die.net/man/1/touch)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules/
+tmp/
--- a/README.md
+++ b/README.md
@@ -530,6 +530,29 @@ Notable exceptions:
 + There is no "quiet" option since default behavior is to run silent.
 
 
+### touch([options ,] file)
+Available options:
+
++ `'-a'`: Change only the access time
++ `'-c'`: Do not create any files
++ `'-m'`: Change only the modification time
++ `'-d DATE'`: Parse DATE and use it instead of current time
++ `'-r FILE'`: Use FILE's times instead of current time
+
+Examples:
+
+```javascript
+touch('source.js');
+touch('-c', '/path/to/some/dir/source.js');
+touch({ '-r': FILE }, '/path/to/some/dir/source.js');
+```
+
+Update the access and modification times of each FILE to the current time.
+A FILE argument that does not exist is created empty, unless -c is supplied.
+This is a partial implementation of *[touch(1)](http://linux.die.net/man/1/touch)*.
+
+
+
 ## Non-Unix commands
 
 
--- a/src/touch.js
+++ b/src/touch.js
@@ -1,0 +1,112 @@
+var common = require('./common');
+var fs = require('fs');
+
+//@
+//@ ### touch([options ,] file)
+//@ Available options:
+//@
+//@ + `'-a'`: Change only the access time
+//@ + `'-c'`: Do not create any files
+//@ + `'-m'`: Change only the modification time
+//@ + `'-d DATE'`: Parse DATE and use it instead of current time
+//@ + `'-r FILE'`: Use FILE's times instead of current time
+//@
+//@ Examples:
+//@
+//@ ```javascript
+//@ touch('source.js');
+//@ touch('-c', '/path/to/some/dir/source.js');
+//@ touch({ '-r': FILE }, '/path/to/some/dir/source.js');
+//@ ```
+//@
+//@ Update the access and modification times of each FILE to the current time.
+//@ A FILE argument that does not exist is created empty, unless -c is supplied.
+//@ This is a partial implementation of *[touch(1)](http://linux.die.net/man/1/touch)*.
+//@
+//
+function _touch(opts, files) {
+  opts = common.parseOptions(opts, {
+    'a': 'atime_only',
+    'c': 'no_create',
+    'd': 'date',
+    'm': 'mtime_only',
+    'r': 'reference',
+  });
+
+  if (!files) {
+    common.error('no paths given');
+  }
+
+
+  if (Array.isArray(files)) {
+    files.forEach(function(f) {
+      touchFile(opts, f);
+    });
+  } else if (typeof files === 'string') {
+    touchFile(opts, files);
+  } else {
+    common.error('file arg should be a string file path or an Array of string file paths');
+  }
+
+}
+
+function touchFile(opts, file) {
+  var stat = tryStatFile(file);
+
+  if (stat && stat.isDirectory()) {
+    // don't error just exit
+    return;
+  }
+
+  // if the file doesn't already exist and the user has specified --no-create then
+  // this script is finished
+  if (!stat && opts.no_create) {
+    return;
+  }
+
+  // open the file and then close it. this will create it if it doesn't exist but will
+  // not truncate the file
+  fs.closeSync(fs.openSync(file, 'a'));
+
+  //
+  // Set timestamps
+  //
+
+  // setup some defaults
+  var now = new Date();
+  var mtime = opts.date || now;
+  var atime = opts.date || now;
+
+  // use reference file
+  if (opts.reference) {
+    var refStat = tryStatFile(opts.reference);
+    if (!refStat) {
+      common.error('failed to get attributess of ' + opts.reference);
+    }
+    mtime = refStat.mtime;
+    atime = refStat.atime;
+  } else if (opts.date) {
+    mtime = opts.date;
+    atime = opts.date;
+  }
+
+  if (opts.atime_only && opts.mtime_only) {
+    // keep the new values of mtime and atime like GNU
+  } else if (opts.atime_only) {
+    mtime = stat.mtime;
+  } else if (opts.mtime_only) {
+    atime = stat.atime;
+  }
+
+  fs.utimesSync(file, atime, mtime);
+}
+
+module.exports = _touch;
+
+function tryStatFile(filePath) {
+  try {
+    return fs.statSync(filePath);
+  } catch (e) {
+    return null;
+  }
+}
--- a/test/touch.js
+++ b/test/touch.js
@@ -1,0 +1,87 @@
+var shell = require('../shell.js');
+var assert = require('assert');
+var fs = require('fs');
+var crypto = require('crypto');
+
+shell.config.silent = true;
+shell.rm('-rf', 'tmp');
+shell.mkdir('tmp');
+
+// should handle args
+shell.touch();
+assert.ok(shell.error());
+
+shell.touch(1);
+assert.ok(shell.error());
+
+// exits without error when trying to touch a directory
+shell.touch('tmp/');
+assert.ok(!shell.error());
+shell.touch('tmp');
+assert.ok(!shell.error());
+
+// creates new files
+var testFile = tmpFile();
+shell.touch(testFile);
+assert(fs.existsSync(testFile));
+
+// does not create a file if told not to
+var testFile = tmpFile(true);
+shell.touch('-c', testFile);
+assert.ok(!fs.existsSync(testFile));
+
+// errors if reference file is not found
+var testFile = tmpFile();
+var refFile = tmpFile(true);
+shell.touch({'-r': refFile}, testFile);
+assert.ok(shell.error());
+
+// uses a reference file for mtime
+var testFile = tmpFile(false);
+var testFile2 = tmpFile();
+var testFile2Stat = resetUtimes(testFile2);
+
+shell.touch({'-r': testFile2}, testFile);
+var testFileStat = resetUtimes(testFile);
+assert.strictEqual(testFileStat.mtime.getTime(), testFile2Stat.mtime.getTime());
+
+// sets mtime
+var testFile = tmpFile();
+var oldStat = resetUtimes(testFile);
+shell.touch(testFile);
+assert(oldStat.mtime < fs.statSync(testFile).mtime);
+// sets atime
+assert(oldStat.atime < fs.statSync(testFile).atime);
+
+// does not sets mtime if told not to
+var testFile = tmpFile();
+var oldStat = resetUtimes(testFile);
+shell.touch('-a', testFile);
+assert.equal(oldStat.mtime.getTime(), fs.statSync(testFile).mtime.getTime());
+
+// does not sets atime if told not to
+var testFile = tmpFile();
+var oldStat = resetUtimes(testFile);
+shell.touch('-m', testFile);
+assert.equal(oldStat.atime.getTime(), fs.statSync(testFile).atime.getTime());
+
+function resetUtimes(f) {
+  var d = new Date();
+  d.setYear(2000);
+  fs.utimesSync(f, d, d);
+  return fs.statSync(f);
+}
+
+function tmpFile(noCreate) {
+  var str = crypto.randomBytes(Math.ceil(10 / 2)).toString('hex');
+  var file = 'tmp/' + str;
+  if (!noCreate) {
+    fs.closeSync(fs.openSync(file, 'a'));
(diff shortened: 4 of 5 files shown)


marvindanig:
I wish this were merged on priority. But does this work on Windows?

marvindanig:
You're using `fs.existsSync` at line no. 39 which is probably [deprecated](https://github.com/nodejs/node/issues/1592). Check.

blockloop:
**About working on Windows**: I don't use any OS specific code here so I can say with a high confidence that it should on Windows. I don't have any way of proving that, because I only have OSX and Linux.

**About fs.existsSync deprecation**: I had forgotten about that. Mostly because I don't really agree with it. Regardless, I won't debate the API so I removed it and wrote a manual check using statSync. 

I'd like for it to be merged too, but I don't see much activity on this repo so I don't expect it to.

TimothyGu:
Looks good to me.

src/touch.js
```diff
@@ -0,0 +1,89 @@
...
+	} else if (opts.mtime) {
+		atime = fs.statSync(file).atime;
+	}
+
+	fs.utimesSync(file, atime, mtime);
+}
+
+function fileExists(filePath) {
+	try {
+		fs.statSync(filePath, fs.R_OK | fs.W_OK);
```
TimothyGu: `statSync` only takes one argument. I'd go with:

```js
function fileExists(filePath) {
	try {
		(fs.existsSync || fs.accessSync || fs.statSync)(filePath, fs.R_OK | fs.W_OK);
		return true;
	} catch (e) {
		return false;
	}
}
```

src/touch.js
```diff
@@ -0,0 +1,89 @@
...
+
+	// if the file doesn't already exist and the user has specified --no-create then
+	// this script is finished
+	if (!fileExists(file) && opts.no_create) {
+		return;
+	}
+
+	// open the file and then close it. this will create it if it doesn't exist but will
+	// not wipe out it's contents
+	fs.closeSync(fs.openSync(file, 'a+'));
```
TimothyGu: I'd go for the actual flags used by `touch` from coreutils: `O_WRONLY | O_CREAT` (they are available in `require('constants')`).

src/touch.js
```diff
@@ -0,0 +1,89 @@
...
+		atime = stat.atime;
+	} else if (opts.date) {
+		mtime = opts.date;
+		atime = opts.date;
+	}
+
+	if (opts.atime) {
+		mtime = fs.statSync(file).mtime;
+	} else if (opts.mtime) {
+		atime = fs.statSync(file).atime;
```
TimothyGu: Here you are `stat`ing way too much than necessary. It gives a higher chance of race conditions as another process could remove the file in between stats.

src/touch.js
blockloop: I'm not sure I follow what this accomplishes or fixes. Can you elaborate a bit?

src/touch.js
blockloop: Good idea. I'll update.

src/touch.js
blockloop: True. I should stat once and stash the result in a variable. Will fix.

TimothyGu:
Seems like I forgot to post this, but I already fixed the problems described above:

```diff
diff --git a/src/touch.js b/src/touch.js
index eee2237..bb7f5d2 100644
--- a/src/touch.js
+++ b/src/touch.js
@@ -1,5 +1,8 @@
 var common = require('./common');
 var fs = require('fs');
+var constants = require('constants');
+var O_CREAT = constants.O_CREAT || 0;
+var O_WRONLY = constants.O_WRONLY || 0;
 
 module.exports = _touch;
 
@@ -36,19 +39,21 @@ function _touch(opts, file) {
                common.error('cannot touch directory ' + file);
        }
 
-       if (fileExists(file) && fs.statSync(file).isDirectory()) {
+       var stat = tryToStat(file);
+
+       if (stat && stat.isDirectory()) {
                common.error('cannot touch directory ' + file);
        }
 
        // if the file doesn't already exist and the user has specified --no-create then
        // this script is finished
-       if (!fileExists(file) && opts.no_create) {
+       if (!stat && opts.no_create) {
                return;
        }
 
        // open the file and then close it. this will create it if it doesn't exist but will
        // not wipe out it's contents
-       fs.closeSync(fs.openSync(file, 'a+'));
+       fs.closeSync(fs.openSync(file, O_WRONLY | O_CREAT));
 
        //
        // Set timestamps
@@ -61,7 +66,6 @@ function _touch(opts, file) {
 
        // use reference file
        if (opts.reference) {
-               var stat = fs.statSync(file);
                mtime = stat.mtime;
                atime = stat.atime;
        } else if (opts.date) {
@@ -70,20 +74,19 @@ function _touch(opts, file) {
        }
 
        if (opts.atime) {
-               mtime = fs.statSync(file).mtime;
+               mtime = stat.mtime;
        } else if (opts.mtime) {
-               atime = fs.statSync(file).atime;
+               atime = stat.atime;
        }
 
        fs.utimesSync(file, atime, mtime);
 }
 
-function fileExists(filePath) {
+function tryToStat(filePath) {
        try {
-               fs.statSync(filePath, fs.R_OK | fs.W_OK);
-               return true;
+               return fs.statSync(filePath);
        } catch (e) {
-               return false;
+               return undefined;
        }
 }
 
```

blockloop:
I want to know what the purpose of this is and why it's necessary:

```diff
-       fs.closeSync(fs.openSync(file, 'a+'));
+       fs.closeSync(fs.openSync(file, O_WRONLY | O_CREAT));
```

TimothyGu:
@blockloop it is to imitate [touch](http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/touch.c#n135)

blockloop:
I'm not sure that this specific piece will make much of a difference. The `a+` flag [translates to similar flags](https://github.com/nodejs/node/blob/master/lib/fs.js#L528) which accomplish the goal without having to `require` another module.

nfischer:
@blockloop thanks for the link! The link suggests that the `a` and `w` flags translate even more closely. Would it be better to use `a` over `a+`? I think we should keep dependencies at a minimum (see #122).

@TimothyGu could you provide an example where the behavior would differ between the `a` flag and `O_WRONLY | O_CREAT`?

blockloop:
I think the point I'm trying to make is that the purpose of this line of code is to `Try to open FILE, creating it if necessary` (as the comment in the GNU source says) and either of these flags appear solve that problem. The only other caveat is that we don't want to truncate the file and that is covered.

src/touch.js
```diff
@@ -0,0 +1,89 @@
+var common = require('./common');
+var fs = require('fs');
+
+module.exports = _touch;
```
nfischer: This should be immediately after the closing curly brace of `_touch()` if possible

src/touch.js
```diff
@@ -0,0 +1,89 @@
...
+//@
+//@ ```javascript
+//@ touch('file.txt');
+//@ ```
+//@
+//@ Creates a file if it does not exist. Sets the file's mtime and ctime to NOW if the file
+//@ does exist
+function _touch(opts, file) {
+	opts = common.parseOptions(opts, {
+		'a': 'atime',
```
nfischer: Perhaps rename to `atime_only`? That would make more logical sense given the man page:

> ```
> -a     change only the access time
> ```

src/touch.js
blockloop: Are you suggestiong that I move the exports below the definition of the `touch` function?

src/touch.js
```diff
@@ -0,0 +1,89 @@
...
+		'd': 'date',
+		'm': 'mtime',
+		'r': 'reference',
+	});
+
+	if (!file) {
+		common.error('no paths given');
+	}
+
+	if (typeof file !== 'string') {
```
nfischer: Could we allow this to support a `file` argument that is either a string or an array of strings? Similar to how [`ls()`](https://github.com/shelljs/shelljs/blob/master/src/ls.js#L41) does it.

src/touch.js
```diff
@@ -0,0 +1,89 @@
...
+		common.error('file arg should be a single file path');
+	}
+
+	// if the string ends with "/" then stop
+	if (file.indexOf('/', file.length - 1) !== -1) {
+		common.error('cannot touch directory ' + file);
```
nfischer: I ran the following and saw no errors:

```
$ mkdir foo/; touch foo/
```

Could we change this to just skip the file instead? That seems more in-line with POSIX behavior.

src/touch.js
nfischer: Yes, precisely. That's more along the lines of the style of the other commands

nfischer:
Would you be able to reformat the files to use a two-space indent instead of tabs? That'll keep things stylistically similar.

src/touch.js
blockloop: It's against my style, but I favor consistency so if that's the case then I suppose I can change it.

test/touch.js
```diff
@@ -0,0 +1,83 @@
...
+assert(oldStat.atime.getTime() === fs.statSync(testFile).atime.getTime());
+
+
+
+function resetUtimes(f) {
+	var d = new Date();
+	d.setYear(2000);
+	fs.utimesSync(f, d, d);
+	return fs.statSync(f);
+}
```
nfischer: Please end the file with `shell.exit(123)` (it's required for the test runner)

test/touch.js
```diff
@@ -0,0 +1,83 @@
...
+testFile = 'tmp/test999';
+shell.touch(testFile);
+assert(fs.existsSync(testFile));
+
+testFile = 'tmp/test';
+var testFile2 = 'tmp/test2';
+var testFile2Stat = resetUtimes(testFile);
+shell.touch({'-f': testFile2}, testFile);
+var testFileStat = resetUtimes(testFile);
+assert(testFileStat.mtime.getTime() === testFile2Stat.mtime.getTime());
```
nfischer: Change to `assert.equal()` (unless `===` is particularly important here)

TimothyGu:
@nfischer practically, no, but I prefer to keep the syscalls exactly the same as coreutils. It doesn't matter much either way so I'm not going to insist on this.

blockloop:
I applied all of the discussed changes. I also added an `.editorconfig` file to avoid indent issues

blockloop:
Since I changed the indentation you'll want to [append ?w=1](https://github.com/blog/967-github-secrets#whitespace) to see the actual changes in the most recent diff. I wish I would have changed the whitespaces as a separate commit.

.gitignore
```diff
@@ -1 +1,2 @@
 node_modules/
+tmp/
```
nfischer: Is this necessary? It'd be best to remove any temp directories created during tests if possible.

README.md
```diff
@@ -522,6 +522,26 @@ Notable exceptions:
   given to the umask.
 + There is no "quiet" option since default behavior is to run silent.
 
+### touch([options ,] file)
+Available options:
```
nfischer: Can you make these modifications to `src/test.js` and then run `scripts/generate-docs.js`? Also, please remove the long options (i.e. `--atime`) since these aren't supported.

README.md
```diff
@@ -522,6 +522,26 @@ Notable exceptions:
...
+
++ `a`, `--atime`: Change only the access time
++ `c`, `--no-create`: Do not create any files
++ `d`, `--date`: Parse STRING and use it instead of current time
++ `m`, `--mtime`: Change only the modification time
++ `r`, `--reference`: Use this file's times instead of current time
```
nfischer: Add an argument to `-r`, perhaps like `-r FILE`

.gitignore
blockloop: Since the tests aren't under mocha or any test framework there's no guaranteed way to cleanup the directory without wrapping the whole file in a try/catch/finally. If any test fails then the cleanup will not happen.

blockloop:
Anything else need changing?

nfischer:
Could you rebase off master and squash your commits into one?

I fixed a bug with the README, so this should fix the merge conflict.

blockloop:
Done

nfischer:
LGTM :+1: @ariporad what do you think?

Although I don't think this handles every possible edge case, I think this handles enough of them that it's ready to be a feature. If there are any other significant edge cases, they'll be observed by users, and can be added as features/bug fixes down the road.

.editorconfig
```diff
@@ -0,0 +1,7 @@
+# EditorConfig is awesome: http://EditorConfig.org
```
ariporad: Can you move this into a separate PR? I'd like to try and keep them small and focused. (I can show you some git-fu for how to do that if you'd like, because I _*totally*_ understand git :P).

blockloop: Sure thing. I was afraid you would ask for that :)

> On Jan 11 2016, at 7:45 pm, Ari Porad &lt;notifications@github.com&gt;
wrote:  

>

> In [.editorconfig](https://github.com/shelljs/shelljs/pull/249#discussion_r4
9406493):

>  
>  
>     > @@ -0,0 +1,7 @@

>     > +# EditorConfig is awesome: http://EditorConfig.org

>  
>

> Can you move this into a separate PR? I'd like to try and keep them small
and focused. (I can show you some git-fu for how to do that if you'd like,
because I __totally__ understand git :P).

>

> —  
Reply to this email directly or [view it on GitHub](https://github.com/shelljs
/shelljs/pull/249/files#r49406493).

ariporad: Thanks!

ariporad:
Other than the comments I made (mostly regarding the `.editorconfig`), LGTM!

.editorconfig
blockloop: Done

> On Jan 11 2016, at 7:46 pm, Ari Porad &lt;notifications@github.com&gt;
wrote:  

>

> In [.editorconfig](https://github.com/shelljs/shelljs/pull/249#discussion_r4
9406606):

>  
>  
>     > @@ -0,0 +1,7 @@

>     > +# EditorConfig is awesome: http://EditorConfig.org

>  
>

> Thanks!

>

> —  
Reply to this email directly or [view it on GitHub](https://github.com/shelljs
/shelljs/pull/249/files#r49406606).

nfischer:
The CI failed the first time, but seems to be passing now. I'll merge for now, and we can investigate more closely if the break occurs again.

marvindanig:
:+1: Great to see this feature merged back right in. 

But why isn't the shelljs bumped up from 0.5.3 to something to reflect these changes?

nfischer:
@marvindanig You raise a great point! I've been more concerned with merging new changes and bug fixes, and less concerned with making a release. I think we will bump the version number once for when we make our next release. If you need the version number bumped right now (like if you're checking `package.json` in the git repo), then I can do that.

marvindanig:
:+1:  Nya, I'm good. Just that I didn't realize about this feature being available until I came back to this repo again (today). Otherwise `npm check updates` would have reflected if it was versioned up. 

Looking at the rate at which this library is used I think it would good to bake in semver right away.

ariporad:
Yeah, we're going to use 100% semver staring with he next release. (Because we're <1.0.0, it will be a minor bump).



On Mon, Jan 18, 2016 at 8:22 PM -0800, "Marvin Danig" <notifications@github.com> wrote:












  Nya, I'm good. Just that I didn't realize about this feature being available until I came back to this repo again (today). Otherwise npm check updates would have reflected if it was versioned up. 



Looking at the rate at which this library is used I think it would good to bake in semver right away.

nfischer:
^ Yup, I was thinking the same thing, and hopefully we'll hit v1.0 within the month (if we can add enough features).

Issue #245, opened 2015-10-19
Dir glob breaks when in the middle of path


nickdima:
If you have a path like `some_dir/another_dir/**/*.ext` when you run `common.expand` on it the resulted filepaths will be relative to `another_dir` instead of containing the full path.

nickdima:
Maybe [here](https://github.com/shelljs/shelljs/blob/master/src/common.js#L106) it should be:

```javascript
expanded.push(root + file);
```

ddbruce:
I had this issue too (see [this](https://github.com/ddbruce/thermopi/blob/master/app.js)). Fixed it by modifying common.js as @nickdima said.

nfischer:
Sounds like this could also be related to #225, which recommends using the glob module.

nfischer:
@nickdima @ddbruce Does #275 fix this issue as well?

nfischer:
@nickdima @ddbruce I think this is fixed now by #359. Please reopen if this is still an issue.

PR #257, opened 2015-12-03
Fix wildcard use in middle of filename

ddbruce:
This is a fix for #245.


ariporad:
@ddbruce: Thanks so much for your contribution, but unfortunately, once #275 lands, we've completely replaced the globbing with the [`glob`](http://npm.im/glob) module, so your patch won't really work. So I have to close this.

I'm sorry.

PR #258, opened 2015-12-03
Use rechoir and interpret for scripts.

knpwrs:
This adds support for babel, iced coffeescript, typescript, jsx (maybe somebody is generating HTML in a script, eh?), [etc](https://github.com/js-cli/js-interpret/blob/a62691b31731f0006e8e04cd06cfaf4c47afa722/index.js).

This can be modified to still execute out-of-process which may be desirable because of the `require('../global')` statement. Alternatively, we could require non-globally and leave everything in the `shjs` file.

### DIFF
--- a/bin/shjs
+++ b/bin/shjs
@@ -32,24 +32,8 @@ for (var i = 0, l = args.length; i < l; i++) {
   }
 }
 
-if (scriptName.match(/\.coffee$/)) {
-  //
-  // CoffeeScript
-  //
-  if (which('coffee')) {
-    exec('coffee "' + scriptName + '" ' + args.join(' '), function(code) {
-      process.exit(code);
-    });
-  } else {
-    console.log('ShellJS: CoffeeScript interpreter not found');
-    console.log();
-    process.exit(1);
-  }
-} else {
-  //
-  // JavaScript
-  //
-  exec('node "' + scriptName + '" ' + args.join(' '), function(code) {
-    process.exit(code);
-  });
-}
+var path = require('path');
+var extensions = require('interpret').extensions;
+var rechoir = require('rechoir');
+rechoir.prepare(extensions, scriptName);
+require(require.resolve(path.join(process.cwd(), scriptName)));
--- a/package.json
+++ b/package.json
@@ -24,7 +24,10 @@
   "bin": {
     "shjs": "./bin/shjs"
   },
-  "dependencies": {},
+  "dependencies": {
+    "interpret": "^1.0.0",
+    "rechoir": "^0.6.2"
+  },
   "devDependencies": {
     "coffee-script": "^1.10.0",
     "jshint": "~2.1.11"
--- a/test/shjs.js
+++ b/test/shjs.js
@@ -4,10 +4,9 @@ var assert = require('assert');
 
 function runScript(name) {
   // prefix with 'node ' for Windows, don't prefix for OSX/Linux
-  return shell.exec((process.platform === 'win32' ? 'node ' : '') +
-                    path.resolve(__dirname, '../bin/shjs') +
-                    ' ' +
-                    path.resolve(__dirname, 'resources', 'shjs', name), { silent: true });
+  var cmd = (process.platform === 'win32' ? 'node' : '') + path.resolve(__dirname, '../bin/shjs');
+  var script = path.join('resources', 'shjs', name);
+  return shell.exec(cmd + ' ' + script, { silent: true });
 }
 
 // Exit Codes


nfischer:
Looks like a good addition. Does anyone else in the community want support for these languages?

TimothyGu:
The notion of being able to automatically use Babel to transpile the script sounds fairly appealing to me.

nfischer:
@ariporad @arturadib what do you guys think? Is it worth the dependency? I have yet to test if this handles exit codes properly and things such as that (which I think are more important than supporting arbitrary languages)

arturadib:
agreed @nfischer , if nothing breaks like exit codes, that's fine. I don't think we have tests for `shjs` so maybe add those in with consideration for what behavior we need to preserve first, then see if this passes those?

nfischer:
Sounds like a good idea. I might be able to write up some basic unit tests in bash for shjs, just to make sure that stderr and exit code are preserved (since exec() used to have issues with those)

ariporad:
Once #304 lands, we can see if this breaks shjs.

ariporad:
@knpwrs: Could you rebase off master? Thanks!

knpwrs:
The three-way merge is looking a little hairy:

image

Are you guys looking to preserve anything in particular? Or should I just take my changes?

nfischer:
@knpwrs most of your stuff you should keep. Make sure this is compatible with file names with spaces and node binary paths with spaces, that's all

knpwrs:
Travis failed on one version of node. Appveyor failed a test that I was trying to fix locally on my Mac. I don't currently have access to a Windows machine to debug.

nfischer:
@knpwrs This passes travis. I think someone wrote a bad test for `touch()`. I'll investigate

nfischer:
@knpwrs I dug through the results on my windows 7 box. I see errors that look like `Error: Cannot find module 'C:\Users\Username\C:\path\to\shelljs\test...`. Notice that `C:\` is in there twice. That might be specific to my system, but it's a bug that I wasn't able to quickly get past. My guess is that one of the dependencies doesn't have good Windows support for how you're trying to use it.

If you'd like to debug, feel free to use appveyor as you wish (you can push commits with helpful logging statements, and remove them later). Otherwise, I vote to close the issue, since it isn't worth losing Windows support.

ariporad:
@knpwrs: What's the status on this? I'd really like to get this merged, I miss not having ES6 in my shelljs scripts.

knpwrs:
I've just been super busy. My only real option for testing on Windows is a VM, but I don't think I'll really be able to get anywhere this week.

TimothyGu:
I can take a look.

TimothyGu:
Alright this PR has more than a bunch of issues.

- #355 is needed for me to get the tests working without modifications
- it seems like test/shjs.js wasn't rebased correctly: `__dirname` and a bunch of other fixes are not present

After fixing all of the above, there is only one change needed to fix the issue reported by @nfischer, which is changing `path.join` to `path.resolve` in shjs.

The incremental diff is attached.

```diff
diff --git a/bin/shjs b/bin/shjs
index 9ad0e1a..75ca58b 100755
--- a/bin/shjs
+++ b/bin/shjs
@@ -36,4 +36,4 @@ var path = require('path');
 var extensions = require('interpret').extensions;
 var rechoir = require('rechoir');
 rechoir.prepare(extensions, scriptName);
-require(require.resolve(path.join(process.cwd(), scriptName)));
+require(require.resolve(path.resolve(process.cwd(), scriptName)));
diff --git a/test/shjs.js b/test/shjs.js
index 02380a6..8b95faa 100644
--- a/test/shjs.js
+++ b/test/shjs.js
@@ -4,8 +4,8 @@ var assert = require('assert');

 function runScript(name) {
   // prefix with 'node ' for Windows, don't prefix for OSX/Linux
-  var cmd = (process.platform === 'win32' ? 'node' : '') + path.resolve(__dirname, '../bin/shjs');
-  var script = path.join('resources', 'shjs', name);
+  var cmd = (process.platform === 'win32' ? 'node ' : '') + path.resolve(__dirname, '../bin/shjs');
+  var script = path.resolve(__dirname, 'resources', 'shjs', name);
   return shell.exec(cmd + ' ' + script, { silent: true });
 }


```

nfischer:
@TimothyGu thanks! Awesome work!

nfischer:
@TimothyGu @knpwrs Could we get this rebased? I think most of the blocking issues have been resolved. Thanks!

nfischer:
I refactored this into an up-to-date PR (see #384). It seems to be passing CI so far. If that one is good, I think we should go with that and close this one.

nfischer:
#384 is merged, so I'm closing this. Thanks @knpwrs for the PR!

PR #168, opened 2014-11-27
Fix returning process exit codes

eluck:
This should fix issue #133

[Review on Reviewable](https://reviewable.io/reviews/arturadib/shelljs/168)

Issue #133, opened 2014-05-31
shjs command always exits with zero code


ndelitski:
Lets write a simple test ``test.js``
```js
process.exit(1)
```
then after i type ``shjs test.js || echo 1`` console gives me no output, but ``node test.js || echo 1`` ended with 1 code as expected. Tested in Linux and Windows. 

Maybe I do not catch up something... But obviously ``shjs`` must persist exit code from script it running and i see this logic in ``shjs`` file, but in `./bin/shjs` new child node process doesn't pass exit code to main process. Is the failed child process should not ruin parent process with error exit code? Maybe because it is asynchronous...

So I wait until the subprocess is completed and manualy close main process with the same exit code then all works perfect.
```js
exec('node ' + scriptName + ' ' + args.join(' '), {async: true}, function(code) {
      exit(code);
});
```

mastern2k3:
bump,
I'm using shjs with TeamCity, and this behavior is hiding errors, fixing this would be great..

ndelitski:
any progress so far?

nfischer:
This should now be fixed in master

jonaskello:
I'm also running in to this. Could we have a new npm package with this fix?

ariporad:
@jonaskello: We're working at a new release, we want to do a few more things.

patrixd:
I think this is happening again in the las version of `bin/shjs`.

PR #264, opened 2015-12-23
shjs script will now exit with proper status code, fixing silent failures

nfischer:
This is a fix for #133. This is based off the fix in #168, but this is up to date, and changes less (doesn't affect `package.json` and doesn't fiddle with how coffeescript is found in the path).

I've tested this on my Ubuntu 15.10 system, and it behaves as expected. To test the behavior, try something such as:

```Bash
$ cat input.js
require('shelljs/global');
exit(123);
$ shjs input.js; echo $? # the current version outputs '0'
123
```

This demonstrates that `shjs` now respects the exit code produced by `node` when running the script, and returns it as expected.

### DIFF
--- a/bin/shjs
+++ b/bin/shjs
@@ -37,7 +37,9 @@ if (scriptName.match(/\.coffee$/)) {
   // CoffeeScript
   //
   if (which('coffee')) {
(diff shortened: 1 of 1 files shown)


nfischer:
Closing this, since #252 was first and does the same thing.

Issue #209, opened 2015-06-12
Where is standard error going to?


n2liquid:
If I run:

    console.log(sh.exec('echo WHOA THERE, AN ERROR CONDITION... 1>&2'));
    console.log(sh.error());

I get:

    WHOA THERE, AN ERROR CONDITION...
    { code: 0, output: '' }
    null

So how do I get messages from standard error in JavaScript?

nfischer:
I think there might be a few problems going on here that might need updating. I get very different behavior depending on which commands I run:

```Bash
$ cat test.js
require('shelljs/global');
exec('echo hello world >&2');

$ shjs test.js >/dev/null # no stdout

$ shjs test.js 2>/dev/null # no stderr
hello world

$ node test.js >/dev/null # no stdout, using node
hello world

$ node test.js 2>/dev/null # no stderr, using node
```

This suggests that something is at least broken in the `shjs` script if not somewhere else as well. I suspect that this is related, and points toward a bug related to the `exec()` function (`shjs` is essentially a wrapper for a call to `exec('node ' + script_name + args);` (see [here](https://github.com/shelljs/shelljs/blob/master/bin/shjs#L50)).

I suspect it should just be modified to support capturing a process's stderr in addition to its stdout, and have both of these returned alongside the return code.

nfischer:
This was bugging me, so I added this to #268. @n2liquid does this help you achieve what you wanted?

nfischer:
Closing this, since #268 was merged. If this doesn't work as expected, please reopen this issue and let us know.

PR #268, opened 2015-12-29
Commands that have multiple errors now produce cleaner log output

### LINKED ISSUE #267: Poor output for commands with multiple errors
The following code snippet demonstrates poor error message handling:

```
$ # assume that fake1, fake2, and fake3 do not exist
$ cat test.js
require('shelljs/global');
ls('fake1', 'fake2', 'fake3');

$ shjs test.js
ls: no such file or directory: fake1

ls: no such file or directory: fake1
ls: no such file or directory: fake2

ls: no such file or directory: fake1
ls: no such file or directory: fake2
ls: no such file or directory: fake3
```

As you can see, the error message is concatenated and re-printed each time. I think the cleaner behavior is to concatenate and only print when the command exits.

nfischer:
This is a fix for #267. The new behavior can be seen to be as such:

```
$ # assume that fake1, fake2, and fake3 do not exist
$ cat test.js
require('shelljs/global');
ls('fake1', 'fake2', 'fake3');

$ shjs test.js
ls: no such file or directory: fake1
ls: no such file or directory: fake2
ls: no such file or directory: fake3
```

This is comparable to the Bash behavior:

```
$ ls fake1 fake2 fake3
ls: cannot access fake1: No such file or directory
ls: cannot access fake2: No such file or directory
ls: cannot access fake3: No such file or directory
```

Fixes #267, #209.

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -142,7 +142,8 @@ function execSync(cmd, opts) {
   // True if successful, false if not
   var obj = {
     code: code,
-    output: stdout,
+    output: stdout, // deprecated
+    stdout: stdout,
     stderr: stderr
   };
   return obj;
@@ -150,7 +151,7 @@ function execSync(cmd, opts) {
 
 // Wrapper around exec() to enable echoing output to console in real time
 function execAsync(cmd, opts, callback) {
-  var output = '';
+  var stdout = '';
   var stderr = '';
 
   var options = common.extend({
@@ -159,11 +160,11 @@ function execAsync(cmd, opts, callback) {
 
   var c = child.exec(cmd, {env: process.env, maxBuffer: 20*1024*1024}, function(err) {
     if (callback)
-      callback(err ? err.code : 0, output, stderr);
+      callback(err ? err.code : 0, stdout, stderr);
   });
 
   c.stdout.on('data', function(data) {
-    output += data;
+    stdout += data;
     if (!options.silent)
       process.stdout.write(data);
   });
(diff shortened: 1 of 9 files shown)


nfischer:
Update: I updated the tests to support the new error output format (which has no trailing newline). I slightly modified the tests by moving the `numLines` function into a separate `utils.js` module that is imported (reduces duplication of code).

I added **support for stderr**. Error messages for commands will now be printed to stderr by default (which is more consistent with unix utilities).

Also, **`exec()` will now respect a process's stderr, and will return a new `stderr` field** (in addition to `code` and `output`). This aims to provide a solution to #209. `output` now refers to only a process's stdout, not all output. This breaks the current behavior, which combines stdout and stderr into `output`, but it's more consistent with unix, so I think this is the way to go. Both stdout and stderr are outputted to the terminal by default, which is consistent with current behavior.

Because stdout and stderr are no longer combined together into `output`, this fixes the issue raised in [my comment](https://github.com/shelljs/shelljs/issues/209#issuecomment-167904161), where `shjs` commands print everything to stdout.

nfischer:
@ariporad This should be ready for review now. This deprecates the old `output` property and replaces it with `stdout` (although `output` will still contain all `stdout`). I think using the name `stdout` is more clear, which is probably the better way to go.

src/exec.js
```diff
@@ -98,8 +113,9 @@ function execSync(cmd, opts) {
     // sleepFile is used as a dummy I/O op to mitigate unnecessary CPU usage
     // (tried many I/O sync ops, writeFileSync() seems to be only one that is effective in reducing
     // CPU usage, though apparently not so much on Windows)
-    while (!fs.existsSync(codeFile)) { updateStdout(); fs.writeFileSync(sleepFile, 'a'); }
-    while (!fs.existsSync(stdoutFile)) { updateStdout(); fs.writeFileSync(sleepFile, 'a'); }
+    while (!fs.existsSync(codeFile)) { updateStream(stdoutFile); fs.writeFileSync(sleepFile, 'a'); }
```
ariporad: I think this is a duplicate line.

ariporad:
I made one comment, but other than that, LGTM!

src/exec.js
nfischer: Which does it duplicate? That's meant to replace the original `!fs.existsSync(codeFile)` line. It differs from the next line, which uses `!fs.existsSync(stdoutFile)`

ariporad: Oh yeah, you're right.

ariporad:
LGTM!

PR #269, opened 2015-12-30
Implementing error codes and adding the errorCode() function

nfischer:
1. This implements error codes for the builtin commands. Each command will now have an associated error code (`common.status.code`). I changed many of the commands to return the code that the corresponding Linux command would return (ex. `ls('asdfasdf')` returns 2).

 2. **This fixes a bug with `exec()`** as well. Previously, you would see the following results:

 ```
 $ cat test.js
 require('shelljs/global');
 exec('exit 5'); // or any code besides 1, 2, or 126+
 if (error())
   echo('caught the error');
 else
   echo('this error slips by');

 $ shjs test.js
 this error slips by

 ```

 Any non-zero exit code means that `error()` will return a truthy value.

 3. If `config.fatal` is set, then the **process will exit with the last return code that was set, emulating Bash's behavior.**

 4. This adds the **`errorCode()` function to access the last error code**, which is somewhat of a substitute for Bash's `$?` variable. This is good if a command's output may vary, but its return codes can be relied upon to stay consistent (or if output is inconvenient to type). It can be used like so:

 ```Javascript
 require('shelljs/global');
 var assert = require('assert');
 ls('missing_file1');
 var str1 = error(),
     code1 = errorCode();
 ls('missing_file2');
 var str2 = error(),
     code2 = errorCode();
 assert.ok(str1 !== str2);
 assert.ok(code1 === code2);
 ```

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -11,6 +11,7 @@ exports.config = config;
 
 var state = {
   error: null,
+  errorCode: 0,
   currentCmd: 'shell.js',
   previousDir: null,
   tempDir: null
@@ -26,8 +27,19 @@ function log() {
 }
 exports.log = log;
 
+var DEFAULT_ERROR_CODE = 1;
+
 // Shows error message. Throws unless _continue or config.fatal are true
-function error(msg, _continue) {
+function error(msg, _code, _continue) {
+  // Adjust values if `_continue` is passed in but `_code` isn't
+  if (typeof _code === 'boolean') {
+    _continue = _code;
+    _code = DEFAULT_ERROR_CODE;
+  }
+
+  if (typeof _code !== 'number')
+    _code = DEFAULT_ERROR_CODE;
+
   if (state.error === null)
     state.error = '';
   var log_entry = state.currentCmd + ': ' + msg;
@@ -36,11 +48,15 @@ function error(msg, _continue) {
   else
     state.error += '\n' + log_entry;
 
+  // Override previous return code, but prefer a previous non-default code if it exists
+  if (state.errorCode === 0 || _code !== DEFAULT_ERROR_CODE)
+    state.errorCode = _code;
+
   if (msg.length > 0)
     log(log_entry);
 
   if (config.fatal)
-    process.exit(1);
+    process.exit(state.errorCode);
 
   if (!_continue)
     throw '';
@@ -191,6 +207,7 @@ function wrap(cmd, fn, options) {
 
     state.currentCmd = cmd;
     state.error = null;
+    state.errorCode = 0;
 
     try {
       var args = [].slice.call(arguments, 0);
--- a/src/exec.js
+++ b/src/exec.js
@@ -137,9 +137,8 @@ function execSync(cmd, opts) {
   try { common.unlinkSync(codeFile); } catch(e) {}
   try { common.unlinkSync(sleepFile); } catch(e) {}
 
-  // some shell return codes are defined as errors, per http://tldp.org/LDP/abs/html/exitcodes.html
-  if (code === 1 || code === 2 || code >= 126)  {
-      common.error('', true); // unix/shell doesn't really give an error message after non-zero exit codes
+  if (code !== 0)  {
+    common.error('', code, true); // unix/shell doesn't really give an error message after non-zero exit codes
   }
   // True if successful, false if not
   var obj = {
--- a/README.md
+++ b/README.md
@@ -575,6 +575,10 @@ Follows Python's [tempfile algorithm](http://docs.python.org/library/tempfile.ht
 Tests if error occurred in the last command. Returns `null` if no error occurred,
 otherwise returns string explaining the error
 
+### errorCode()
+Returns status code for the last command. Returns 0 upon success, otherwise
+returns a nonzero integer
+
 
 ## Configuration
 
--- a/shell.js
+++ b/shell.js
@@ -122,8 +122,10 @@ exports.tempdir = common.wrap('tempdir', _tempDir);
 
 
 //@include ./src/error
-var _error = require('./src/error');
+var _error = require('./src/error').error;
 exports.error = _error;
+var _errorCode = require('./src/error').errorCode;
+exports.errorCode = _errorCode;
 
 
 
--- a/src/error.js
+++ b/src/error.js
@@ -4,7 +4,16 @@ var common = require('./common');
 //@ ### error()
 //@ Tests if error occurred in the last command. Returns `null` if no error occurred,
 //@ otherwise returns string explaining the error
-function error() {
+function _error() {
   return common.state.error;
 }
-module.exports = error;
+exports.error = _error;
+
+//@
+//@ ### errorCode()
+//@ Returns status code for the last command. Returns 0 upon success, otherwise
+//@ returns a nonzero integer
+function _errorCode() {
+  return common.state.errorCode;
+}
+exports.errorCode = _errorCode;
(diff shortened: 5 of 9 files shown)


nfischer:
I've modified some of the calls to `error()` to explicitly use code `1` where appropriate (that's the default code). This makes for a messier diff, since it results in many files being modified.

Because the default error code is `1`, I can actually go back and revert these lines, and they should have the same behavior. This cleans up the diff, but might make things more confusing (since it relies on that being the default). Let me know if this is preferred (this is actually the way I'm leaning).

src/common.js
```diff
@@ -25,17 +26,32 @@ function log() {
...
+    _continue = _code;
+    _code = DEFAULT_RETURN_CODE;
+  }
+
+  if (typeof _code !== 'number')
+    _code = DEFAULT_RETURN_CODE;
+
   if (state.error === null)
     state.error = '';
   state.error += state.currentCmd + ': ' + msg + '\n';
```
ariporad: This should be _code != 0

src/common.js
```diff
@@ -25,17 +26,32 @@ function log() {
...
 
+  // Override previous return code
+  if (state.errorCode === 0 || _code !== DEFAULT_RETURN_CODE)
```
ariporad: This should be _code != 0

src/exec.js
```diff
@@ -118,8 +118,8 @@ function execSync(cmd, opts) {
   try { common.unlinkSync(sleepFile); } catch(e) {}
 
   // some shell return codes are defined as errors, per http://tldp.org/LDP/abs/html/exitcodes.html
```
ariporad: This is no longer needed

ariporad:
Looks Great, Just a few minor things I pointed out, and two bigger ones too:
1. Why not just have error() return the code, and get rid of errorCode? Any non-zero value is truthy, so it would still work, and that would make it equivalent to bash's $?.
2. Can you split this into two commits? One for the implimentation, and one for switching over all the functions.

Otherwise, looks great!

src/common.js
nfischer: I wrote it like this for two reasons:

 1. This prefers non-default return codes. If a command has two errors (and `fatal` === false, so it doesn't exit yet), and one is a non-default return code, then it prefers that value overall (which probably tells the user more useful information).
 2. This gives the user a way to send warning messages (output like an error, but don't change the status code) by providing `_code === 0`

I thought about changing it to `if (state.errorCode === 0 || state.errorCode === DEFAULT_RETURN_CODE)` instead, but that would allow `error('...', 0)` to wipe out a previous non-zero code, which it probably shouldn't do.

What do you think?

src/exec.js
nfischer: Done

nfischer:
Thanks for the feedback! I responded inline to the minor comments. In response to the major comments:

 1. I thought about that with `error()`. I agree that the conditionals `if (error())` and `if (errorCode())` will behave the same. My concern is that `error()` already exists and provides a useful functionality in returning the error string. I'm not sure if anyone relies on it for its string value. One option is to rename `error()` to `errorMsg()` and `errorCode()` will become `error()`, but that may still break some scripts.
 2. I'll split this up and rebase in a moment.

Also, what do you think about camel case for `errorCode()`? I was also thinking of naming it `errorcode()`. This would be the only function in the API that is camel case.

We might also consider adding a `warn()` function that outputs to `stderr` without changing the error code. This would be very simple to implement with things the way they are written.

nfischer:
@ariporad I updated this and rebased off master. I reverted most of the calls to `error(...1, true)`, but left my changes for the ones that change the status code to a non-default value. I can go back if and put those in a separate commit if you still prefer that (although this is enough to clean up the diff considerably).

If it looks good, feel free to merge. If not, leave some comments and I'll address them.

nfischer:
@ariporad I think this is ready for review now

ariporad:
@nfisher: Sorry for the delay.

I still think that we should just make `error()` return the error code. `exec().stderr` is there if you need the error string. This is what breaking changes are for.

And we should defiantly use cammelCase. What do we use now?

nfischer:
Right now everything is lower case. "tempdir" is an example.

I agree that works well for exec (and is more intuitive), but I think someone may still want this feature for a non-exec function. Maybe we name this to error() and rename that to errorMsg()?

Or we refactor things *heavily* so that every command returns an object that has a stderr field. I'm working right now to have each command return objects instead of simple strings (which should allow for pipes), and could add the stderr field. This isn't my preferred method though, since I think relying on error messages is fairly fragile (especially since we just fixed bugs related to error messages)

ariporad:
@nfischer: I think that we should wait till me and you and @arturadib (If he wants) to get together in gitter. It kind of revolves around how much we want to mess with shellljs. (I personally want to do a large refactor, and to standardize everything. Then we can just to a breaking change).

nfischer:
:+1:

nfischer:
Rebased off master

ariporad:
@nfischer: Since we're not meeting with Artur, I think we should just refactor everything to return an object with .stderr/.stdout, and have `error()` return the exit code.

nfischer:
My only concern is that it would break a lot of backwards compatibility. I agree it's a cleaner API, but do you think enough people are relying on the old behavior?

I can do that refactor either way, though :+1:

ariporad:
@nfischer: Hmm... My comment didn't go through.

I meant to say (like 4 days ago) that I think breaking changes are there for a reason.

nfischer:
Hmm ok. It'll take me a couple days to sort out the bugs I ran into from the `ShellString` refactor. I'll try to prioritize that and make a separate PR for that, and then we can add error codes on top of that.

ariporad:
@nfischer: What's the status on this? Also, could you rebase off master?

nfischer:
@ariporad still in progress. I can rebase off master, but I think you'd prefer a different interface to expose return codes, so I'm waiting until I make some other changes. I'd at least like a stderr field being returned before we get rid of the current functionality of error()

ariporad:
@nfischer: bump.

nfischer:
@ariporad I think we should wait until #360 is merged, and then revisit alternate implementations. The way I see it it is that we can either do:

 - each ShellString has a `code` attribute (this wouldn't work well for commands like `cp()` that don't return anything)
 - `error()` returns the code, and we rely on the `stderr` attribute to get the error message (this also doesn't work well for commands like `cp()`)

ariporad:
@nfischer: Ok, since #360 is merged, can we move forward with having `error` return the error code?

nfischer:
Yup. I was working on it a bit yesterday. It's been a busy week, but I should have something out by Friday. I'll open this as a new PR.

nfischer:
I'm working on refactoring this today. I should have this out by tonight in a new PR.

levithomason:
Bump, this PR is tagged as blocking 1st release of shx.  Anything I can do to help?

nfischer:
I'll wrap this up later this week, just a little in over my head right now.

@levithomason could you investigate a fix for how shelljs commands (that have fatal errors) don't output to the console sometimes? the issue was introduced when we converted `config.fatal` to throw an exception

levithomason:
Sure thing, I'll review this issue and see what I can find.

nfischer:
@levithomason I'll try to make a bug report that more clearly documents the problem.

And I'll get out a primitive implementation of error codes today (finally got some spare time).

levithomason:
Sounds good, I pulled this branch and perused around but needed more context. Will wait to hear back from you.

nfischer:
@levithomason the bug isn't in this branch. This branch is pretty ancient at this point (I'm refactoring everything in a new branch).

I'll ping you in the bug report, and describe it better there.

nfischer:
I'm closing this, since #394 is a better replacement

PR #270, opened 2016-01-05
Add ability to override execAsync's maxBuffer default of 20MiB

pieterjandesmedt:
A maxBuffer of 20 MiB is not enough when you use for example a `cut` command on a multi-gigabyte file.


nfischer:
It might help if you add this change to the documentation. https://github.com/pieterjandesmedt/shelljs/blob/master/src/exec.js#L167

Also, if you could add an example to the docs, that would probably be good as well.

pieterjandesmedt:
`maxBuffer` option documentation and example added to README and source code

nfischer:
:+1:

pieterjandesmedt:
No problem. Are you going to merge these?

nfischer:
I unfortunately don't have access to the repo. I think only @arturadib does. I'm just trying to review the PRs and issues that I can, so that it's easier to merge things when the project starts getting maintained.

I'm maintaining my own fork (https://github.com/nfischer/shelljs) right now, since this project doesn't seem to be maintained. If you open a PR from this branch to my fork, I'll double check and then merge it.

Be warned, my fork has some extra features that aren't yet merged into this repo, but I think they're useful features that will be appreciated (and will hopefully be merged upstream).

nfischer:
@pieterjandesmedt could you rebase this off master? This should be something we can merge fairly soon.

pieterjandesmedt:
Like this? #284

Issue #101, opened 2014-01-22
cp -r doesn't match linux when destination doesn't exist


gdw2:
In linux, `new_dir` and `existing_dir` are now copies (`new_dir` didn't exist before):
```
$ mkdir existing_dir
$ touch existing_dir/foo
$ cp -rf existing_dir new_dir
```
With shelljs, `new_dir/existing_dir` and `existing_dir` are copies.  I would expect the linux behavior.  
```
cp('-rf', 'existing_dir', 'new_dir')
```

nfischer:
This should be fixed by #271. Would anyone here be able to confirm that this PR fixes it? As far as I can tell, it seems to be good.

PR #271, opened 2016-01-06
Fix cp to match unix behavior

### LINKED ISSUE #256: cp and rename directory with -r doesn't match unix behavior
As shown below, if I do `cp -r a b` with shelljs where `a` is a directory and `b` doesn't exist, shelljs creates `b/a` directories and the copies the contents of `a` into `b/a`.  Instead, like in a unix shell, it should create `b` directory and copy the contents of `a` into `b`.

```
> mkdir a
> touch a/file.txt
> node
> require('shelljs/global')
{}
> cp('-r', 'a', 'b');
undefined
> 
(To exit, press ^C again or type .exit)
> 
> ls b/
a
> rm -rvf b
b/a/file.txt
b/a
b/
> cp -rvp a b
a -> b
a/file.txt -> b/file.txt
```

freitagbr:
Fixes #256, #101.

### DIFF
--- a/test/cp.js
+++ b/test/cp.js
@@ -125,14 +125,14 @@ assert.equal(shell.cat('resources/cp/dir_a/z'), shell.cat('tmp/cp/dir_a/z')); //
 
 //recursive, creates dest dir since it's only one level deep (see Github issue #44)
 shell.rm('-rf', 'tmp/*');
-shell.cp('-r', 'resources/issue44/*', 'tmp/dir2');
+shell.cp('-r', 'resources/issue44', 'tmp/dir2');
 assert.equal(shell.error(), null);
 assert.equal(shell.ls('-R', 'resources/issue44') + '', shell.ls('-R', 'tmp/dir2') + '');
 assert.equal(shell.cat('resources/issue44/main.js'), shell.cat('tmp/dir2/main.js'));
 
 //recursive, does *not* create dest dir since it's too deep (see Github issue #44)
 shell.rm('-rf', 'tmp/*');
-shell.cp('-r', 'resources/issue44/*', 'tmp/dir2/dir3');
+shell.cp('-r', 'resources/issue44', 'tmp/dir2/dir3');
 assert.ok(shell.error());
 assert.equal(fs.existsSync('tmp/dir2'), false);
 
--- a/test/cp.js
+++ b/test/cp.js
@@ -136,6 +136,12 @@ shell.cp('-r', 'resources/issue44/*', 'tmp/dir2/dir3');
 assert.ok(shell.error());
 assert.equal(fs.existsSync('tmp/dir2'), false);
 
+//recursive, creates dest dir, implicitly copies contents of source dir
+shell.rm('-rf', 'tmp/*');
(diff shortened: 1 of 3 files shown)


nfischer:
It might be helpful to add unit tests for this behavior. When I ran a quick test, it seemed to work as expected.

nfischer:
@freitagbr Could you rebase off master?

freitagbr:
@nfischer Done.

nfischer:
Looks like some of the tests are still failing. If we could get these fixed (or have the tests adjusted, if appropriate), that'd be good

test/cp.js
```diff
@@ -125,17 +125,23 @@ assert.equal(shell.cat('resources/cp/dir_a/z'), shell.cat('tmp/cp/dir_a/z')); //
 
 //recursive, creates dest dir since it's only one level deep (see Github issue #44)
 shell.rm('-rf', 'tmp/*');
-shell.cp('-r', 'resources/issue44/*', 'tmp/dir2');
+shell.cp('-r', 'resources/issue44', 'tmp/dir2');
```
nfischer: Should these test cases be changed? What was wrong with the previous behavior?

test/cp.js
```diff
@@ -125,17 +125,23 @@ assert.equal(shell.cat('resources/cp/dir_a/z'), shell.cat('tmp/cp/dir_a/z')); //
...
 assert.equal(shell.error(), null);
 assert.equal(shell.ls('-R', 'resources/issue44') + '', shell.ls('-R', 'tmp/dir2') + '');
 assert.equal(shell.cat('resources/issue44/main.js'), shell.cat('tmp/dir2/main.js'));
 
 //recursive, does *not* create dest dir since it's too deep (see Github issue #44)
 shell.rm('-rf', 'tmp/*');
-shell.cp('-r', 'resources/issue44/*', 'tmp/dir2/dir3');
+shell.cp('-r', 'resources/issue44', 'tmp/dir2/dir3');
```
nfischer: Same question applies here

test/cp.js
freitagbr: In \*nix, `cp -r resources/issue44/* tmp/dir2` is not valid. If there is only one file in `resources/issue44`, it will be copied into `tmp` and be renamed `dir2`. If there is more than one file in `resources/issue44`, the command will fail.

The intention here is to copy all of the files in `resources/issue44` to `tmp/dir2`, where `dir2` will be created. This is achieved with the new command (and in *nix as well).

test/cp.js
freitagbr: Again, `cp -r resources/issue44/* tmp/dir2/dir3` is not valid.

test/cp.js
nfischer: Awesome explanation! The logic sounds right to me. It sounds like this is indeed a bug in the test. Thanks for fixing!

nfischer:
LGTM! @ariporad

test/cp.js
freitagbr: Definitely!

PR #273, opened 2016-01-08
Implementing cd('-') to behave like Bash's "cd -"

nfischer:
I added a `previousDir` field to `common.state` to allow the `cd()` function to have this option. I also added unit tests for this behavior.

Usage is as follows:

```Javascript
// Setup
mkdir('foo');
mkdir('bar');
cd('foo');
cd('../bar');
echo(pwd()); // outputs "/path/to/bar"
cd('-'); // jump back
echo(pwd()); // outputs "/path/to/foo"
cd('-'); // jump back again
echo(pwd()); // outputs "/path/to/bar"
```

### DIFF
--- a/src/cd.js
+++ b/src/cd.js
@@ -8,12 +8,20 @@ function _cd(options, dir) {
   if (!dir)
     common.error('directory not specified');
 
+  if (dir === '-') {
+    if (!common.state.previousDir)
+      common.error('could not find previous directory');
+    else
+      dir = common.state.previousDir;
+  }
+
   if (!fs.existsSync(dir))
     common.error('no such file or directory: ' + dir);
 
   if (!fs.statSync(dir).isDirectory())
     common.error('not a directory: ' + dir);
 
+  common.state.previousDir = process.cwd();
   process.chdir(dir);
 }
 module.exports = _cd;
--- a/src/common.js
+++ b/src/common.js
@@ -12,6 +12,7 @@ exports.config = config;
 var state = {
   error: null,
   currentCmd: 'shell.js',
+  previousDir: null,
   tempDir: null
 };
 exports.state = state;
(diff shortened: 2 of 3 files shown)


nfischer:
This can already be done by something like:

```Javascript
var prevDir = pwd();
cd('foo');
// do something
cd(prevDir); // jump back
```

This PR just makes things a little easier if you're porting `sh` to `shelljs`, and makes things a little more convenient for that use-case. This isn't a critical feature, but it doesn't cause too much harm, so I figured I'd implement it.

src/cd.js
```diff
@@ -8,12 +8,21 @@ function _cd(options, dir) {
   if (!dir)
     common.error('directory not specified');
 
+  if (dir === '-') {
+    var previous_dir = true;
+    if (!common.state.previousDir)
+      common.error('OLDPWD not set');
```
ariporad: Maybe make this error message a little more descriptive? `OLDPWD` takes a minute to understand.

src/common.js
```diff
@@ -12,6 +12,7 @@ exports.config = config;
 var state = {
   error: null,
   currentCmd: 'shell.js',
+  previousDir: null,
```
ariporad: Shouldn't this default to `pwd()`?

ariporad:
@nfischer: Just a few small things, otherwise great job!

src/cd.js
nfischer: Sure thing. I just copied the error message from bash. I agree, it probably should be more descriptive.

src/common.js
nfischer: In bash, `OLDPWD` (the env variable that holds the previous directory) is initially unset. So running `$ cd -` in a new bash shell is an error, which is why I'm initializing to null and checking the value.

Either way, most people aren't going to run `cd('-')` without first having been in some directory, so it's probably not a big concern to have either behavior.

ariporad: Ok, starting with null is fine.

nfischer:
@ariporad this should be good now

ariporad:
LGTM!

PR #275, opened 2016-01-08
Use glob for globbing

### LINKED ISSUE #225: ls globbing does not behave like shell, consider using glob.sync
ls() globbing does not behave like the shell. For instance, this doesn't work:

/opt/stagecoach/apps/*/current/public

The * does not match path components in the middle of a path.

The same thing works, however, if I use the glob module's "glob.sync" method instead.

I suggest using glob.sync rather than a homegrown implementation of ls.

This would impact the options currently supported by ls however.

ariporad:
Fixes #225

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,9 @@
   "bin": {
     "shjs": "./bin/shjs"
   },
-  "dependencies": {},
+  "dependencies": {
+    "glob": "^6.0.3"
+  },
   "devDependencies": {
     "coffee-script": "^1.10.0",
     "jshint": "~2.1.11"
--- a/src/common.js
+++ b/src/common.js
@@ -1,6 +1,6 @@
 var os = require('os');
 var fs = require('fs');
-var _ls = require('./ls');
+var glob = require('glob');
 
 // Module globals
 var config = {
@@ -106,34 +106,15 @@ exports.parseOptions = parseOptions;
 // For example:
 //   expand(['file*.js']) = ['file1.js', 'file2.js', ...]
 //   (if the files 'file1.js', 'file2.js', etc, exist in the current dir)
-function expand(list) {
-  var expanded = [];
-  list.forEach(function(listEl) {
-    // Wildcard present on directory names ?
-    if(listEl.search(/\*[^\/]*\//) > -1 || listEl.search(/\*\*[^\/]*\//) > -1) {
-      var match = listEl.match(/^([^*]+\/|)(.*)/);
-      var root = match[1];
-      var rest = match[2];
-      var restRegex = rest.replace(/\*\*/g, ".*").replace(/\*/g, "[^\\/]*");
-      restRegex = new RegExp(restRegex);
-
-      _ls('-R', root).filter(function (e) {
-        return restRegex.test(e);
-      }).forEach(function(file) {
-        expanded.push(file);
-      });
-    }
-    // Wildcard present on file names ?
-    else if (listEl.search(/\*/) > -1) {
-      _ls('', listEl).forEach(function(file) {
-        expanded.push(file);
-      });
-    } else {
-      expanded.push(listEl);
-    }
-  });
-  return expanded;
+// It's Just a thin wrapper around glob.sync, to allow passing in an array.
+function expand(list, opts) {
+  if (!Array.isArray(list)) list = [list];
+  opts = opts || {};
+  return list.reduce(function (files, pattern) {
+    return files.concat(glob.sync(pattern, opts));
+  }, []);
 }
+
 exports.expand = expand;
 
 // Normalizes _unlinkSync() across platforms to match Unix behavior, i.e.
--- a/src/ls.js
+++ b/src/ls.js
@@ -1,8 +1,4 @@
-var path = require('path');
-var fs = require('fs');
 var common = require('./common');
-var _cd = require('./cd');
-var _pwd = require('./pwd');
 
 //@
 //@ ### ls([options,] [path, ...])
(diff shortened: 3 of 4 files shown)


TimothyGu:
Doesn't shell.js have [its own](https://github.com/shelljs/shelljs/blob/master/src/common.js#L88-L91) glob expansion function?

ariporad:
Oh yeah, I'll fix that.

ariporad:
Fixed.

nfischer:
@ariporad Could you ping this with a push to trigger the CI?

ariporad:
@nfischer: Done.

boutell:
Thank you!

package.json
```diff
@@ -24,7 +24,9 @@
   "bin": {
     "shjs": "./bin/shjs"
   },
-  "dependencies": {},
+  "dependencies": {
+    "glob": "^6.0.3"
+  },
   "devDependencies": {
```
nfischer: Is it possible to use glob without adding it as a package dependency? I know some of the original maintainers wanted to keep dependencies at a minimum (https://github.com/shelljs/shelljs/issues/122#issuecomment-112498119).

src/ls.js
```diff
@@ -36,91 +37,24 @@ function _ls(options, paths) {
...
-  // (e.g. prevents hidden files to be included unless explicitly told so)
-  function pushFile(file, query) {
-    // hidden file?
-    if (path.basename(file)[0] === '.') {
-      // not explicitly asking for hidden files?
-      if (!options.all && !(path.basename(query)[0] === '.' && path.basename(query).length > 1))
-        return false;
-    }
+  paths = paths.map(function (path) {
+    path += (options.recursive ? '/**/*.*' : '/*.*');
```
nfischer: Does this only select files matching `*.*`? Would this also select a with the name `no_dot_in_name`?

src/ls.js
```diff
@@ -1,5 +1,6 @@
 var path = require('path');
 var fs = require('fs');
+var glob = require('glob');
```
nfischer: Is it convenient to move all the `glob` stuff into `common.js`? That would be more shell-like, since a shell handles all the globbing itself (not the specific utilities).

package.json
nfischer: Nevermind. Just tried it and saw that I do need to install it as a dependency. We might need to ping @arturadib and get his input on this in that case.

ariporad: @nfischer: I think it's not that big a deal. If a user is installing us, then they have npm installed. The whole node world favors small focused modules, which mean dependencies. This means that deps are the norm. It's not out of the ordinary. (As a matter of fact, I'd like to try and split this module up into some smaller ones).

However, if @arturadib wants to chime in, that's more than welcome.

src/ls.js
ariporad: @nfischer: No, because the `*` has to match something.

src/ls.js
ariporad: @nfischer: good point. When I have a spare moment (I'm busy for the rest of today, sorry!), I'll do that.

package.json
nfischer: I agree with the philosophy of using modules where they make sense. I think we probably should take advantage of the `glob` module, since it seems to integrate fairly seemlessly with the existing code.

src/ls.js
nfischer: My concern is that on unix you can do:

```Bash
$ touch a
$ touch b
$ touch c.txt
$ ls * # all files
a   b   c.txt
$ ls *.* # only files with dots in the name
c.txt
```

And it looks like this would not support that behavior. If the code does perform as expected, let me know.

arturadib:
@nfischer @ariporad as you can probably tell from `package.json`, I have generally biased to avoid dependencies. I have no problem with leveraging dependencies – esp for larger projects so you can get to market more quickly etc – but the downside of dependencies is that you lose a lot of control, in particular nuanced features and updates, and more often than not packages are abandoned/unmaintained (*cough* *cough* :)), so you end up blocking your own work until someone lands a feature upstream, etc.

luckily in our case our package is pretty small and the problems low level enough that the Node core lib is mostly sufficient. when I do encounter cases where reinventing the wheel takes more than a dozen lines of code or so, I simply copy and paste the relevant snippet :) like this: https://github.com/shelljs/shelljs/blob/master/src/cp.js#L44-L51

(it's obviously a tradeoff as in this case we also don't benefit from a well maintained project upstream, but again, those are all tradeoffs)

all this being said..... glob does seem like a good, well maintained library, and I think our code is reinventing the wheel a bit too much (and not that well), so let's go ahead and see if this is a good solution going forward.

my main concern at this point is breaking existing users, so let's bump at least the minor when we release, and make sure `glob` is mostly shell-compliant. (I'm not too familiar with `**` behavior on *nix, so we'll need to check that).

thanks!

ariporad:
@arturadib: Three things:

1. Personally, I like lots of small modules, so I apologize in advance for when I forget on my PRs (Just remind me and I'll be more than happy to fix them. I'm not a fan of flame wars). 
2. `glob` is written by \@isaacs, the creator and CEO of `npm` (I'm not mentioning him to reduce his spam). No need to worry about it being abandoned.
3. According to semver, it should either be a patch bump if it doesn't break anything, or else a major bump (however, we're <1.0.0, which means that a instead of a major bump we would do a minor bump).

isaacs:
> (I'm not mentioning him to reduce his spam)

Unfortunately, your clever and thoughtful plan failed.  You need to put the `\` between the @ and the name, or else GitHub will still send the notification.  (Not to worry, I get basically an infinite amount of email, so one more or less makes no substantive difference ;)

The goal of the glob module is to match Bash 4.3 pattern matching as closely as possible, so it's probably a great fit for shelljs's needs.  Since this will certainly change some of your existing edge cases, it seems like it probably should be a major version bump for shelljs, just to be on the safe side.

nfischer:
> all this being said..... glob does seem like a good, well maintained library, and I think our code isreinventing the wheel a bit too much (and not that well), so let's go ahead and see if this is a good solution going forward.

I agree. My vote is in favor of switching to glob. I think if reinventing the wheel is simple enough, it may be best to do it ourselves. In this case however, we're not very closely matching Bash globbing, so the switch is probably smart.

It seems like we all agree we should give `glob` a shot. My main concern with this PR is that I'd like to move all globbing into `common.js`. Globbing is normally handled by the shell, not the utilities, in *nix, so handling all globbing in that file will simplify our code and almost certainly get closer to *nix behavior.

ariporad:
@nfischer: the more I think about it, the more I think that I want to submit a PR moving all the command boilerplate into common.js.

I'm unfortunately extremely busy this week, but over the weekend I should have some time to do that.

src/ls.js
```diff
@@ -36,91 +37,24 @@ function _ls(options, paths) {
...
-        return false;
-    }
+  paths = paths.map(function (path) {
+    path += (options.recursive ? '/**/*.*' : '/*.*');
+    return path.replace(/\\/g, '/');
+  });
 
-    if (common.platform === 'win')
-      file = file.replace(/\\/g, '/');
+  var files = paths.reduce(function (files, path) {
```
nfischer: I think if we change this to `files = common.expand(files, { dot: options.all })`, and update `expand()` to take glob's options as a second parameter, this will eliminate the dependency on glob here.

src/ls.js
ariporad: Done.

ariporad:
@nfischer: Ok, moved all the globbing into common, and fixed listing files with no dots.

ariporad:
Also rebased onto master.

src/common.js
```diff
@@ -106,34 +107,15 @@ exports.parseOptions = parseOptions;
...
-      });
-    } else {
-      expanded.push(listEl);
-    }
-  });
-  return expanded;
+// It's Just a thin wrapper around glob.sync, to allow passing in an array.
+function expand(list, opts) {
+  if (!Array.isArray(list)) list = [list];
+  opts = opts || [];
```
nfischer: Should this be `{}`?

nfischer:
@ariporad looks like there's some lint errors.

Also, do we have any test cases that make sure `ls()` (without the `-A` flag) returns a list without `.hidden_file`? If not, let's make sure that case is covered.

src/common.js
nfischer: @ariporad I think this is my only remaining comment. I can merge once this is addressed

ariporad: Fixed.

ariporad:
@nfischer: Should be fixed now.

nfischer:
@ariporad looks like there's some conflicts now (probably due to the `-l` option for ls). Once those are resolved, I can merge

ariporad:
@nfischer: Hmm... I actually think that there's no way to salvage `-l`. It's really tightly coupled to the existing implementation. I think we'll have to re-implement it.

nfischer:
@ariporad could you push this into a branch on the main repo? I'll take a look at refactoring `-l` support. I'd prefer to not lose any features if it can be avoided.

ariporad:
@nfischer: I did, but it won't work to update this PR.

nfischer:
@ariporad we would have to push to a new branch and open that as a PR to replace this one I think.

ariporad:
@nfischer: Yeah, or I can give you push access to my PR. Up to you.

nfischer:
> Yeah, or I can give you push access to my fork. Up to you. (I'd rather not loose all the discussion).

^ even better idea, actually

ariporad:
@nfischer: Ok, you've got push access.

nfischer:
I'll take a look

ariporad:
@nfischer: Ok, you've got push access.

nfischer:
So it turns out that this PR actually is kind of far from working. It passes CI because `run-tests` doesn't actually run any tests (`ls()` is broken). I have most of the tests passing, but I'm still running into some nasty edge cases. I think it may be more worth-while to do some refactoring first, and then make a simpler change that swaps `common.expand()`'s implementation for one that uses `glob.sync()`.

ariporad:
@nfischer: that's what this does. (Take a look at the changes tab). What do you mean it doesn't work?

nfischer:
@ariporad this completely modifies the behavior of ls() as well. Click on any run of the CI and notice how it says "all tests passed"... without listing any of the tests. Now click on any other PR and see that run-tests lists every test file (i.e. cat.js, cp.js, etc) before the words "all tests passed"

ariporad:
@nfischer: Hmm... Let me look in to fixing that.

nfischer:
@ariporad I don't think you can wipe out all the `ls()` code. I think we would be much better off swapping the implementation of `expand()` with one using `glob`. I can add a call to `expand()` inside `common.wrap()`, which I where I think we should handle **most** expansion (it lifts the burden off utilities like `rm()`, which shouldn't need to worry about `expand()`, and vastly improves bash compatibility.

ariporad:
+1 on doing globbing (I'll take care of it).
It looks like this problem can be simplified to five cases we need to be able to handle:

1. `foo` (where `./foo` is a directory, should be equivalent to `foo/*` [or `foo/**/*` if `-r`]).
2. `foo/*.bar`
3. `foo/**/bar` (where `./foo/<whatever>/bar` is a directory, equivalent to `foo/**/bar/*`)
4. `foo/**/bar` (where `./foo/<whatever>/bar` is a file without an extension, equivalent to `foo/**/bar`)
5. `foo/**/bar/*.baz`

Glob can handle 2, 4, 5, so we need to handle 1 and 3. I think the best way to do this is going to be to (pseudocode):

```javascript
var output = [];
var results = glob(input);

for (var i = 0, len = results.length; i < len; ++i) {
    if (fs.statSync(results[i]).isDirectory()) {
        output.push(glob(results[i] + (options.recursive ? '/**/*' : '*')));    
    } else {
        output.push(results[i]);
    }
}
```

nfischer:
I think we should *not* handle cases 1 and 3 in `common.extend()`. The `glob` package works the way Bash would expand things *at the command line*. I think `common.extend()` should expand things just as the commandline would, and it should do that for every utility. Appending `*` to directory names is a pretty sloppy hack, and it's bound to cause us to deviate from Unix behavior sooner or later.

I'll write up a PR that puts a call to `extend()` inside `common.wrap`, and I'll take out extra `extend()` calls in the other utilities where possible. That way you can focus on swapping out the internals of `common.extend()` with `glob.sync`, and making sure we have compatibility.

ariporad:
@nfischer: Agreed. But don't submit the PR (I've already got 75% of it).

ariporad:
@nfischer: Actually, that isn't going to work, because we need to parse the options and then optionally pass `dot` to `glob`.

So I suppose t

ariporad:
Whoops, I didn't mean to close this. Nor post that last comment.

ariporad:
I'm going to demote this to v0.7.0, because we've got some kinks to work out.

nfischer:
Just for future reference, this was closed in favor of #359

PR #276, opened 2016-01-09
Make to and toEnd chainable

TimothyGu:


nfischer:
This seems like a cool potential feature. Is this possible to emulate in Bash though?

ariporad:
@nfischer: Apparently (TIL!):

```bash
$ echo "Apparently this works\!" > foo > bar > baz
$ echo foo
Apparently this works!
$ echo bar
Apparently this works!
$ echo baz
Apparently this works!
```
Screenshot

ariporad:
@TimothyGu: Two things:
1. Could you please squash your commits? Thanks!
2. Other than that, LGTM!

src/to.js
```diff
@@ -21,7 +21,9 @@ function _to(options, file) {
       common.error('no such file or directory: ' + path.dirname(file));
 
   try {
-    fs.writeFileSync(file, this.toString(), 'utf8');
+    var str = this.toString();
+    fs.writeFileSync(file, str, 'utf8');
+    return str;
```
nfischer: Would `return this;` also work? I'm refactoring the ShellString type with the intention that only strings created with `common.ShellString` will have the `.to()` and `.toEnd()` types. This is so that we can avoid modifying the String prototype.

src/toEnd.js
```diff
@@ -21,7 +21,9 @@ function _toEnd(options, file) {
       common.error('no such file or directory: ' + path.dirname(file));
 
   try {
-    fs.appendFileSync(file, this.toString(), 'utf8');
+    var str = this.toString();
+    fs.appendFileSync(file, str, 'utf8');
+    return str;
```
nfischer: Same as above

nfischer:
I just have one comment that should be addressed before merging

src/to.js
TimothyGu: It should.

TimothyGu:
@nfischer @ariporad both comments addressed. Sorry, the comment 16 days ago got lost in my inbox…

nfischer:
LGTM once this passes tests

ariporad:
Merged! Thanks @TimothyGu

PR #279, opened 2016-01-10
shjs utility will also infer the extension of filenames ending in a '.'

### LINKED ISSUE #278: Allow shjs utility to infer the extension for "filename."
Currently, the `shjs` utility can sometimes infer the file extension of the script name. For example:

```
$ cat input.js
require('shelljs/global');
echo('Hello world');

$ shjs input.js
Hello world

$ shjs input # .js extension will be inferred
Hello world

$ shjs input. # it'd be nice if the .js extension could be inferred here as well
ShellJS: script not found (input.)
```

I often find myself writing out simple shelljs scripts and their bash equivalents, and I usually name them similarly (`input.js` and `input.sh` respectively, for example). Tab complete will complete up to `$ shjs input.`, so it'd be convenient if `shjs` would infer that I want to run the `.js` file.

Since the utility already infers file extensions, this would just be a (perhaps more useful) expansion of that.

nfischer:
`shjs` can now infer the file extension if the file ends in `.`. So the following code now works:

```Bash
$ cat input.js
require('shelljs/global');
echo('Hello world');

$ shjs input.js
Hello world

$ shjs input
Hello world

$ shjs input. # this was previously an error
Hello world
```

This is useful in the case of tab completion. If there's another file (say, `input.sh`), then tab completion will result in "`input.`". This is an often enough scenario that, if we're going to do file extension inference at all, we should probably do it here. Since inference is already supported, I see no reason to not support it in this scenario as well.

This fixes #278

### DIFF
--- a/bin/shjs
+++ b/bin/shjs
@@ -12,10 +12,14 @@ var args,
 env['NODE_PATH'] = __dirname + '/../..';
 
 if (!scriptName.match(/\.js/) && !scriptName.match(/\.coffee/)) {
-  if (test('-f', scriptName + '.js'))
-    scriptName += '.js';
-  if (test('-f', scriptName + '.coffee'))
-    scriptName += '.coffee';
+  var prefix = scriptName;
+  if (prefix[prefix.length-1] === '.')
+    prefix = prefix.slice(0, -1);
+
+  if (test('-f', prefix + '.js'))
+    scriptName = prefix + '.js';
+  else if (test('-f', prefix + '.coffee'))
+    scriptName = prefix + '.coffee';
 }
 
 if (!test('-f', scriptName)) {


ariporad:
Hmm... I don't really think that this is a good idea, it would be a little confusing.

nfischer:
I think the file extension inference is a little confusing in general. This would just make that already-existing-feature more useful (in my opinion). It's not a big concern though, so I'm fine if it doesn't get merged.

ariporad:
Hmm... Anyone else have any opinions on this?

nfischer:
Closing this, since there doesn't seem to be much interest (and it wasn't a big concern of mine anyway)

PR #282, opened 2016-01-11
Fix relative symlinks

### LINKED ISSUE #100: Symbolic links are broken!

freitagbr:
Fixes #100

### DIFF
--- a/src/ln.js
+++ b/src/ln.js
@@ -29,9 +29,11 @@ function _ln(options, source, dest) {
   }
 
   source = String(source);
+  var sourcePath = path.normalize(source).replace(RegExp(path.sep + '$'), '');
+  var isAbsolute = (path.resolve(source) === sourcePath);
   dest = path.resolve(process.cwd(), String(dest));
 
-  if (!fs.existsSync(source) && !path.isAbsolute(source) && !fs.existsSync(path.dirname(dest) + '/' + source)) {
+  if ((isAbsolute && !fs.existsSync(sourcePath)) || !fs.existsSync(path.resolve(process.cwd(), path.dirname(dest), source))) {
     common.error('Source file does not exist', true);
   }
 
--- a/src/ln.js
+++ b/src/ln.js
@@ -28,10 +28,10 @@ function _ln(options, source, dest) {
     common.error('Missing <source> and/or <dest>');
   }
 
-  source = path.resolve(process.cwd(), String(source));
+  source = String(source);
   dest = path.resolve(process.cwd(), String(dest));
 
-  if (!fs.existsSync(source)) {
+  if (!fs.existsSync(source) && !path.isAbsolute(source) && !fs.existsSync(path.dirname(dest) + '/' + source)) {
     common.error('Source file does not exist', true);
   }
 
--- a/src/ln.js
+++ b/src/ln.js
@@ -33,10 +33,6 @@ function _ln(options, source, dest) {
   var isAbsolute = (path.resolve(source) === sourcePath);
   dest = path.resolve(process.cwd(), String(dest));
 
-  if ((isAbsolute && !fs.existsSync(sourcePath)) || !fs.existsSync(path.resolve(process.cwd(), path.dirname(dest), source))) {
-    common.error('Source file does not exist', true);
-  }
-
   if (fs.existsSync(dest)) {
     if (!options.force) {
       common.error('Destination file exists', true);
@@ -46,8 +42,14 @@ function _ln(options, source, dest) {
   }
 
   if (options.symlink) {
+    if ((isAbsolute && !fs.existsSync(sourcePath)) || !fs.existsSync(path.resolve(process.cwd(), path.dirname(dest), source))) {
+      common.error('Source file does not exist', true);
+    }
     fs.symlinkSync(source, dest, os.platform() === "win32" ? "junction" : null);
   } else {
+    if (!fs.existsSync(source)) {
+      common.error('Source file does not exist', true);
+    }
     fs.linkSync(source, dest, os.platform() === "win32" ? "junction" : null);
   }
 }
--- a/test/ln.js
+++ b/test/ln.js
@@ -31,6 +31,15 @@ assert.ok(shell.error());
 shell.ln('tmp/noexist', 'tmp/linkfile1');
 assert.ok(shell.error());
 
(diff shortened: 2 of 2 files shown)


freitagbr:
I based this PR on Schoonology's fix (PR #235), removing the warning that is displayed when a symbolic link is created.

nfischer:
@freitagbr Could you check that this works under node v0.10? It seems to be failing on the CI.

nfischer:
Also, would you be able to squash the commits? See [here](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/)

freitagbr:
The Node.js v0.10 `path` module does not have the `isAbsolute` method (it was added in Node v0.11). That is what is causing the build to fail.

nfischer:
The usual method we use is to  check for its existence, and if it doesn't exist, use a workaround. Would that be feasible?

freitagbr:
As long as the source string does not start with a `/`, it is absolute. Simply checking `source.indexOf('/') !== 0` should do the same thing, correct?

nfischer:
Will that play well with Windows? I think that should be sufficient for *nix though (unless `~` is supported)

freitagbr:
This should work as intended, though `~` is still not supported. It is trivial to add support, and I think it would be worthwhile. Thoughts?

nfischer:
I'm working on a PR for the tilde (currently unsupported), but there are some other big changes I want to get merged first. So don't add support for it just yet. Once we do, we'll have to add an "is absolute path" method that handles the tilde

test/ln.js
```diff
@@ -84,7 +84,7 @@ assert.equal(
 );
 
 // Abspath regression
-shell.ln('-sf', 'tmp/file1', path.resolve('tmp/abspath'));
+shell.ln('-sf', 'file1', path.resolve('tmp/abspath'));
```
nfischer: :+1:

test/ln.js
```diff
@@ -96,4 +96,21 @@ assert.equal(
...
+assert.equal(
+  fs.readFileSync('tmp/new/file1.txt').toString(),
+  fs.readFileSync('tmp/new/file2.txt').toString()
+);
+fs.writeFileSync('tmp/new/file1.txt', 'new content txt');
+assert.equal(
+  fs.readFileSync('tmp/new/file2.txt').toString(),
+  'new content txt'
+);
+
```
nfischer: Could you please add a test that uses hard links? the semantics for hard links are a little different.

```
$ mkdir -p tmp/
$ echo 'not in tmp' > ./file1.txt
$ echo 'within tmp' > ./tmp/file1.txt
$ ln -sf file1.txt tmp/softlink.txt
$ ln -f file1.txt tmp/hardlink.txt
$ cat tmp/softlink.txt
within tmp
$ cat tmp/hardlink.txt
not in tmp
```

nfischer:
Could you also add some invalid test cases toward the top?

```
# symlinks
mkdir -p tmp/
touch tmp/file1.txt
ln -sf tmp/file1.txt tmp/file2.txt
# make sure there was an error

mkdir -p tmp/
touch file1.txt
ln -sf file1.txt tmp/file2.txt
# make sure there was an error

# hardlinks
mkdir -p tmp/
touch tmp/file1.txt
ln -f file1.txt tmp/file2.txt
# make sure there was an error
```

ariporad:
@freitagbr: Any chance I could ask you to add the small changes? I'd really like to include this in the next release. Thanks!

freitagbr:
@ariporad Will do! I'll have this done by the end of the day.

nfischer:
:+1: @freitagbr just ping this thread when you have the tests merged in and I'll review

ariporad:
@freitagbr: Thanks, but no rush!


Ari

test/ln.js
freitagbr: Looks like there already is a test for hard links, line 62 in `test/ln.js`.

nfischer: :+1:

nfischer:
@freitagbr then my only other comment is the invalid test cases (https://github.com/shelljs/shelljs/pull/282#issuecomment-174239702). Once those are in, this'll be good!

src/ln.js
```diff
@@ -44,8 +42,14 @@ function _ln(options, source, dest) {
...
 
   if (options.symlink) {
+    if ((isAbsolute && !fs.existsSync(sourcePath)) || !fs.existsSync(path.resolve(process.cwd(), path.dirname(dest), source))) {
+      common.error('Source file does not exist', true);
+    }
     fs.symlinkSync(source, dest, os.platform() === "win32" ? "junction" : null);
   } else {
+    if (!fs.existsSync(source)) {
+      common.error('Source file does not exist', true);
+    }
```
freitagbr: In testing, I found that the source file was not checked properly for hard links.This correctly checks if the source exists for hard links.

nfischer:
@freitagbr nice work!

LGTM

ariporad:
LGTM!

ariporad:
:+1:

PR #283, opened 2016-01-11
chore(travis): add OS X to Travis CI

nfischer:
Just adding OS X to the CI, since this should be supported (and it was very easy to add). If this PR fails, then it'll point out some bugs that we have to track down.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,23 @@
-language: node_js
-node_js:
-  - "0.10"
-  - "0.11"
-  - "0.12"
-  - "4.2.4"
+language: c++
+env:
+  - NODE_VERSION="0.10"
+  - NODE_VERSION="0.11"
+  - NODE_VERSION="0.12"
+  - NODE_VERSION="4"
+  - NODE_VERSION="5"
+
+# keep this blank to make sure there are no before_install steps
+before_install:
+
+install:
+  - rm -rf ~/.nvm
+  - git clone https://github.com/creationix/nvm.git ~/.nvm
+  - source ~/.nvm/nvm.sh
+  - nvm install $NODE_VERSION
+  - node --version
+  - npm install
+os:
+  - linux
+  - osx
+script:
+  - npm test


ariporad:
It looks like travis doesn't fully support node on osx out of the box:
travis-ci/travis-ci#2311.

nfischer:
Yup, saw that as soon as I pushed. I'm going to try the steps here (https://github.com/travis-ci/travis-ci/issues/2311#issuecomment-121773546)

nfischer:
@ariporad This should be ready for review now. Commits are squashed.

This wound up being more changes to the yaml file than I anticipated. It should work very similarly to the previous behavior though. I followed the recommendation from [this comment](https://github.com/travis-ci/travis-ci/issues/2311#issuecomment-121773546). I changed language to `c++` because that gets rid of the automatic `nvm` commands (and works under both linux and os x).

This makes things look a bit uglier in the yaml file, but perhaps the benefit of testing under os x is worth it.

Also, does this follow the preferred convention for commit messages?

.travis.yml
```diff
@@ -1,6 +1,22 @@
...
-  - "0.10"
-  - "0.11"
-  - "0.12"
-  - "4.2.4"
+language: c++
+env:
+  - NODE_VERSION="0.10"
+  - NODE_VERSION="0.11"
+  - NODE_VERSION="0.12"
+  - NODE_VERSION="4.2.4"
```
ariporad: Two things:
1. I'd say just change this to `4`
2. Can you add `5`, and/or `stable`?

ariporad:
@nfischer: I made a comment or two, but otherwise looks good. And the commit message is great!

.travis.yml
nfischer: I think it's at 4.2.4 because that's the latest LTS (@arturadib set that up). I'll add a more recent version as well though.

nfischer:
:+1:

nfischer:
@ariporad
I updated travis. The CI will still test under 4.2.4 (I think that's the last LTS) and will now test under 5.0. I think this should be sufficient coverage for node versions on both linux and OS X, so this is ready for review. Let me know if this needs any other changes.

.travis.yml
ariporad: @nfischer: I know, but change this from `NODE_VERSION="4.2.4"` to just `NODE_VERSION="4"`. That way if they do a patch release it will automatically work.

.travis.yml
```diff
@@ -1,6 +1,23 @@
...
+  - NODE_VERSION="5.0"
```
ariporad: This should just be `5` (The current version of node is `5.4.1`, and that won't match `5.0`.

.travis.yml
```diff
@@ -1,6 +1,23 @@
...
+
```
ariporad: NVM has a special keyword: `stable`, which automagically gets the latest version of node. We should add that too.

.travis.yml
```diff
@@ -1,6 +1,23 @@
-language: node_js
-node_js:
-  - "0.10"
-  - "0.11"
-  - "0.12"
-  - "4.2.4"
+language: c++
+env:
```
ariporad: Question: Do we want to test on `iojs`?

.travis.yml
nfischer: According to the github page for nvm:

> stable: this alias is deprecated, and only truly applies to node v0.12 and earlier. Currently, this is an alias for node.

.travis.yml
nfischer: Ok, I can switch to 5. I wasn't sure if we always wanted to test against the latest release of node (could be somewhat unstable).

.travis.yml
nfischer: @arturadib removed that a couple days ago, so I left it out

.travis.yml
nfischer: @ariporad Finally got around to fixing this. You're right, 4.x.x is a better move going forward. This should be good for review now.

ariporad:
LGTM!

PR #284, opened 2016-01-11
add ability to override execAsync's maxBuffer default of 20MiB

pieterjandesmedt:
exec() maxBuffer override option now in a separate branch.

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -5,6 +5,8 @@ var path = require('path');
 var fs = require('fs');
 var child = require('child_process');
 
+var DEFAULT_MAXBUFFER_SIZE = 20*1024*1024;
+
 // Hack to run child_process.exec() synchronously (sync avoids callback hell)
 // Uses a custom wait loop that checks for a flag file, created when the child process is done.
 // (Can't do a wait loop that checks for internal Node variables/messages as
@@ -20,6 +22,9 @@ function execSync(cmd, opts) {
   var options = common.extend({
     silent: common.config.silent
   }, opts);
+
+  var maxBuffer = parseInt(opts.maxBuffer, 10) || DEFAULT_MAXBUFFER_SIZE;
+  if (isNaN(maxBuffer)) maxBuffer = DEFAULT_MAXBUFFER_SIZE;
 
   var previousStdoutContent = '';
   // Echoes stdout changes from running process, if not silent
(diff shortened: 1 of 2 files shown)


nfischer:
Marking this as an update of #270 (which was closed).

nfischer:
Does this also work for `execSync`?

pieterjandesmedt:
Good catch - option added to execSync as well.

nfischer:
Do we also need to make the change on lines like https://github.com/pieterjandesmedt/shelljs/blob/exec-maxBuffer-override/src/exec.js#L90 where the value is hardcoded?

Also, could you move `20*1024*1024` into a constant at the top of the file? Something like `DEFAULT_BUF_SIZE`.

pieterjandesmedt:
Done & done.

nfischer:
Let me check this later today, and I'll give you an LGTM if I don't find any issues.

do you know of any way that we could continue using small buffers but simply open up a new buffer when it gets full? This would be even better, since then we could have an option to effectively never crash from filling up the buffer

pieterjandesmedt:
AFAIK, there's no easy way to increase the buffer size once the child process is running...

src/exec.js
```diff
@@ -21,6 +23,9 @@ function execSync(cmd, opts) {
     silent: common.config.silent
   }, opts);
 
+  var maxBuffer = parseInt(opts.maxBuffer, 10) || DEFAULT_MAXBUFFER_SIZE;
+  if (isNaN(maxBuffer)) maxBuffer = DEFAULT_MAXBUFFER_SIZE;
```
nfischer: Is there a way to combine these checks into one `if`? That way we only have to assign the default value once, which is probably cleaner.

src/exec.js
```diff
@@ -137,7 +143,10 @@ function execAsync(cmd, opts, callback) {
     silent: common.config.silent
   }, opts);
 
-  var c = child.exec(cmd, {env: process.env, maxBuffer: 20*1024*1024}, function(err) {
+  var maxBuffer = parseInt(opts.maxBuffer, 10) || DEFAULT_MAXBUFFER_SIZE;
+  if (isNaN(maxBuffer)) maxBuffer = DEFAULT_MAXBUFFER_SIZE;
```
nfischer: Same as my comment above

src/exec.js
```diff
@@ -178,6 +188,15 @@ function execAsync(cmd, opts, callback) {
 //@   console.log('Exit code:', code);
 //@   console.log('Program output:', output);
 //@ });
+//@
+//@ var child = exec('cut -d\\; -f1 some_huge_file.csv', {
```
nfischer: Could we use a separator other than `;` here? It would just make the example cleaner to use something that doesn't need to be escaped. Not a big concern though.

src/exec.js
```diff
@@ -178,6 +188,15 @@ function execAsync(cmd, opts, callback) {
...
+//@   async: true,
+//@   silent: true,
+//@     maxBuffer: 1024*1024*1024 // 1 GiB
```
nfischer: Indentation

nfischer:
Overall, this looks great! Just a few comments above to address. I tested this on Linux, and saw that this helped avoid crashes when the output buffer was exceeded, so it works as intended.

Also, please squash your commits and rebase off the upstream master, and this should be good to merge.

ariporad:
@pieterjandesmedt: Could you add the changes and rebase onto master please? I'd really like to get this into the next release.

ariporad:
@pieterjandesmedt: bump

pieterjandesmedt:
Unsure how to proceed... I squashed the commits an created PR #333. You decide :-)

ariporad:
@pieterjandesmedt: As much as I hate to reject a PR, it looks like this has been fixed by #335, so I have to close this. Sorry.

pieterjandesmedt:
No problem, as long as it's fixed, I'm happy :-)

PR #288, opened 2016-01-14
chore: update github URL in package.json

nfischer:
The package.json still used the old `arturadib` github URL. This is now updated for the new `shelljs` URL. This will save someone an HTTP redirect.


ariporad:
LGTM!

PR #290, opened 2016-01-14
Search PATHEXT instead of 3 hardcoded values

isiahmeadows:
Redo of #134.

Since it was only a few lines, it was easy to rewrite against master.

### DIFF
--- a/src/which.js
+++ b/src/which.js
@@ -2,6 +2,10 @@ var common = require('./common');
 var fs = require('fs');
 var path = require('path');
 
+// XP's system default value for PATHEXT system variable, just in case it's not
+// set on Windows.
+var XP_DEFAULT_PATHEXT = '.com;.exe;.bat;.cmd;.vbs;.vbe;.js;.jse;.wsf;.wsh';
+
 // Cross-platform method for splitting environment PATH variables
 function splitPath(p) {
   if (!p)
@@ -26,7 +30,8 @@ function checkPath(path) {
 //@ var nodeExec = which('node');
 //@ ```
 //@
-//@ Searches for `command` in the system's PATH. On Windows looks for `.exe`, `.cmd`, and `.bat` extensions.
+//@ Searches for `command` in the system's PATH. On Windows, this uses the
+//@ `PATHEXT` variable to append the extension if it's not already executable.
 //@ Returns string containing the absolute path to the command.
 function _which(options, cmd) {
   if (!cmd)
@@ -43,26 +48,37 @@ function _which(options, cmd) {
       if (where)
         return; // already found it
 
-      var attempt = path.resolve(dir + '/' + cmd);
+      var attempt = path.resolve(dir, cmd);
 
       if (common.platform === 'win') {
+        attempt = attempt.toUpperCase();
+
+        // In case the PATHEXT variable is somehow not set (e.g.
+        // child_process.spawn with an empty environment), use the XP default.
+        var pathExtEnv = process.env.PATHEXT || XP_DEFAULT_PATHEXT;
+        var pathExtArray = splitPath(pathExtEnv.toUpperCase());
+        var i;
+
+        // If the extension is already in PATHEXT, just return that.
+        for (i = 0; i < pathExtArray.length; i++) {
+          var ext = pathExtArray[i];
+          if (attempt.slice(-ext.length) === ext && checkPath(attempt)) {
+            where = attempt;
+            return;
+          }
+        }
+
+        // Cycle through the PATHEXT variable
         var baseAttempt = attempt;
-        attempt = baseAttempt + '.exe';
-        if (checkPath(attempt)) {
-          where = attempt;
-          return;
-        }
-        attempt = baseAttempt + '.bat';
-        if (checkPath(attempt)) {
-          where = attempt;
-          return;
-        }
-        attempt = baseAttempt + '.cmd';
-        if (checkPath(attempt)) {
-          where = attempt;
-          return;
+        for (i = 0; i < pathExtArray.length; i++) {
+          attempt = baseAttempt + pathExtArray[i];
+          if (checkPath(attempt)) {
+            where = attempt;
+            return;
+          }
         }
       } else {
+        // Assume it's Unix-like
         if (checkPath(attempt)) {
           where = attempt;
           return;
--- a/test/which.js
+++ b/test/which.js
@@ -16,15 +16,24 @@ shell.which();
 assert.ok(shell.error());
 
 var result = shell.which('asdfasdfasdfasdfasdf'); // what are the odds...
-assert.equal(shell.error(), null);
-assert.equal(result, null);
+assert.ok(!shell.error());
+assert.ok(!result);
 
 //
 // Valids
 //
 
-var result = shell.which('node');
-assert.equal(shell.error(), null);
-assert.equal(fs.existsSync(result), true);
+var node = shell.which('node');
+assert.ok(!shell.error());
+assert.ok(fs.existsSync(node));
+
+if (process.platform === 'win32') {
+    // This should be equivalent on Windows
+    var nodeExe = shell.which('node.exe');
+    assert.ok(!shell.error());
+    // If the paths are equal, then this file *should* exist, since that's
+    // already been checked.
+    assert.equal(node, nodeExe);
+}
 
 shell.exit(123);
--- a/README.md
+++ b/README.md
@@ -373,7 +373,8 @@ Examples:
 var nodeExec = which('node');
 ```
 
-Searches for `command` in the system's PATH. On Windows looks for `.exe`, `.cmd`, and `.bat` extensions.
+Searches for `command` in the system's PATH. On Windows, this uses the
+`PATHEXT` variable to append the extension if it's not already executable.
 Returns string containing the absolute path to the command.
 
 


src/which.js
```diff
@@ -26,8 +29,9 @@ function checkPath(path) {
 //@ var nodeExec = which('node');
 //@ ```
 //@
-//@ Searches for `command` in the system's PATH. On Windows looks for `.exe`, `.cmd`, and `.bat` extensions.
-//@ Returns string containing the absolute path to the command.
```
nfischer: Why was this line removed?

nfischer:
@isiahmeadows could you `generate-docs`?

src/which.js
```diff
@@ -2,6 +2,9 @@ var common = require('./common');
 var fs = require('fs');
 var path = require('path');
 
+// XP's system default value for PATHEXT system variable
```
BYK: Might be good to mentioned that WinXP lacks this thing in the comment. Also not sure if supporting Windows XP is a good idea. It is quite old and unsupported even by Microsoft.

src/which.js
```diff
@@ -47,20 +51,16 @@ function _which(options, cmd) {
...
-          where = attempt;
-          return;
-        }
-        attempt = baseAttempt + '.cmd';
-        if (checkPath(attempt)) {
-          where = attempt;
-          return;
+        var pathExtEnv = process.env.PATHEXT || XP_DEFAULT_PATHEXT;
+        var pathExtArray = splitPath(pathExtEnv);
+        var len = pathExtArray.length;
```
BYK: Don't think array length caching makes any difference so you can just use `pathExtArray.length` in the `for` part without any performance penalty.

BYK:
LGTM

FYI: Windows also has a built-in `where` command.

src/which.js
nfischer: I don't see much harm in the default value. I agree that WinXP doesn't need to be supported by us, since the OS is end-of-life by Microsoft. However, just in case it's perhaps possible to delete environmental variables in Windows (is this even possible?), It'd be good to keep some fall-back value.

src/which.js
nfischer: I don't know if it's part of the standard or not. Either way is fine with me, since the meaning is very clear (and performance isn't super critical yet).

src/which.js
BYK: You can delete environmental variables but if you delete this one I'm guessing shelljs would be the least of your problems :D

BYK: Still, a default value is a good fallback so we should keep it.

src/which.js
BYK: There are posts over the Internet saying it is simply unnecessary unless you're dealing with very old IE versions. Both ways are fine.

src/which.js
nfischer: ^ Good point. But yes, I also vote we keep it for no other reason than extra safety measure.

src/which.js
```diff
@@ -26,8 +29,9 @@ function checkPath(path) {
 //@ var nodeExec = which('node');
 //@ ```
 //@
-//@ Searches for `command` in the system's PATH. On Windows looks for `.exe`, `.cmd`, and `.bat` extensions.
-//@ Returns string containing the absolute path to the command.
+//@ Searches for `command` in the system's PATH. On Windows, looks for extensions in the PATHEXT
```
nfischer: *Totally optional suggestion*

Perhaps we don't need this much info in the README? The README is supposed to give a brief overview of  commands, and this might be too much detail. I think it's sufficient to say something to the effect of *"On Windows, this uses the `PATHEXT` environment variable to append the extension"* here.

I think it's better to have this comment down below in the function body.

isiahmeadows: Will fix.

src/which.js
isiahmeadows: By accident.

src/which.js
isiahmeadows: I'm clarifying the comment as it's a precautionary measure.

src/which.js
isiahmeadows: Nowadays, I normally don't cache the length unless I'm writing multiple loops of the same array in a performance sensitive section of code. This was originally a simple copy-paste of a 1 1/2 year old PR, where I was still in the habit.

(Cheap checks first are often faster in my experience unless you're dealing with large arrays. But that doesn't apply here.)

isiahmeadows:
Do note that I use Linux, my Windows installation is rather broken ATM, and my computer is too old/underpowered to run a VM, so I don't know how well this runs, and I kinda feel guilty not being able to test this. :frowning: 

But feel free to test this yourself. I did add a test. It's also a trivial enough patch that it *should* work.

(@nfischer If you all got Appveyor to run for PRs as well, that'd be awesome, probably even for you, considering this supports Windows. :wink:)

I submitted the original PR, #134, about 1 1/2 years ago, when my Windows installation actually worked (mine needs *way* more troubleshooting than I have time for ATM - my standard user directory requires Admin privileges to even *read*).

nfischer:
I'll test this out on my Windows machine today and see how it runs. Thanks again!

test/which.js
```diff
@@ -23,8 +23,16 @@ assert.equal(result, null);
...
 assert.equal(shell.error(), null);
-assert.equal(fs.existsSync(result), true);
+assert.equal(fs.existsSync(node), true);
+
+if (process.platform === 'win32') {
+    // This should be equivalent on Windows
+    var nodeExe = shell.which('node.exe');
+    assert.equal(shell.error(), null);
+    assert.equal(fs.existsSync(nodeExe), true);
+    assert.equal(node, node.exe);
```
nfischer: typeo: s/node.exe/nodeExe

test/which.js
```diff
@@ -23,8 +23,16 @@ assert.equal(result, null);
...
+var node = shell.which('node');
 assert.equal(shell.error(), null);
-assert.equal(fs.existsSync(result), true);
+assert.equal(fs.existsSync(node), true);
+
+if (process.platform === 'win32') {
+    // This should be equivalent on Windows
+    var nodeExe = shell.which('node.exe');
+    assert.equal(shell.error(), null);
+    assert.equal(fs.existsSync(nodeExe), true);
```
nfischer: nit: change to `assert.ok(fs.existsSync(node))`

test/which.js
```diff
@@ -23,8 +23,16 @@ assert.equal(result, null);
 // Valids
 //
 
-var result = shell.which('node');
+var node = shell.which('node');
 assert.equal(shell.error(), null);
-assert.equal(fs.existsSync(result), true);
+assert.equal(fs.existsSync(node), true);
```
nfischer: nit: change to `assert.ok(fs.existsSync(node))`

src/which.js
```diff
@@ -46,23 +51,32 @@ function _which(options, cmd) {
...
+        // In case the PATHEXT variable is somehow not set (e.g.
+        // child_process.spawn with an empty environment), use the XP default.
+        var pathExtEnv = process.env.PATHEXT || XP_DEFAULT_PATHEXT;
+        var pathExtArray = splitPath(pathExtEnv);
+        var i;
+
+        // If the extension is already in PATHEXT, just return that.
+        for (i = 0; i < pathExtArray.length; i++) {
+          var ext = pathExtArray[i];
+          if (attempt.slice(-ext.length) === ext && checkPath(attempt)) {
```
nfischer: This does a case-sensitive comparison. What we'd really like is a case-insensitive comparison for Windows. Could we perhaps use a regex with the `/i` option?

test/which.js
```diff
@@ -23,8 +23,16 @@ assert.equal(result, null);
...
+
+if (process.platform === 'win32') {
+    // This should be equivalent on Windows
+    var nodeExe = shell.which('node.exe');
+    assert.equal(shell.error(), null);
```
nfischer: nit: `assert.ok(!shell.error())`

nfischer:
The nits are all totally optional. They're just for stylistic consistency (which we aren't very good at anyway, as you can probably see by other lines in this test that differ from the style I'm suggesting).

The big issue is the comment about case-insensitive comparison (and also see the typo I pointed out). Once that gets updated, I'll test again and give the thumbs up.

src/which.js
BYK: I'd prefer running `.toUpperCase()` on the extracted extension.

nfischer: I prefer case insensitive comparison. There's no guarantee (as far as I can tell) that the PATHEXT values will **always** be uppercase, and it's a little excessive to call .toUpperCase() on both sides. The regex should be very simple. Whichever is easiest is fine with me though

isiahmeadows: Oops...will fix when I get home...

test/which.js
isiahmeadows: Yeah...I think yours looks better, anyways.

test/which.js
isiahmeadows: Oops...will fix when I get home...

nfischer:
@isiahmeadows Could you rebase this off master? We now have Windows CI set up with appveyor, and it'd be great to verify this under that (although I'm fairly certain it'll pass after the rebase).

isiahmeadows:
I like that GitHub has this "Update Branch" button where I could do that here. :smile:

ariporad:
What? They added such a button‽‽‽

Thanks! Ari
On Thu, Jan 28, 2016 at 3:38 PM, shelljs/shelljs
<reply@reply.github.com>
wrote:
I like that GitHub has this "Update Branch" button where I could do that here.:smile: [https://assets-cdn.github.com/images/icons/emoji/unicode/1f604.png]

—
Reply to this email directly or view it on GitHub
[https://github.com/shelljs/shelljs/pull/290#issuecomment-176482283] .[https://github.com/notifications/beacon/ABu7pJGYEXZGHWuh2J1oFc_U2wgsCAwAks5pep3-gaJpZM4HFNsu.gif]

isiahmeadows:
That's what I just used. And it doesn't appear to be passing... I'll fix it when I get home. I've been a bit busy lately.

isiahmeadows:
This latest push won't fix the build, but it'll give me a better idea what happened.

isiahmeadows:
@nfischer It's surprising that the AppVeyor build isn't required...I'm guessing that was an oversight? (I just noticed this.)

test/which.js
isiahmeadows: I removed this because it's redundant after the initial check (if this fails after the last, it's not a bug in this).

isiahmeadows:
I've rebased against `master` to remove all the superflouous commits.

isiahmeadows:
@ariporad I think this is a recent addition, where if you don't have merge conflicts, you can just update your remote branch through GitHub itself instead of locally having to do it. It's just like with merging branches.

nfischer:
@isiahmeadows is checkpath() case insensitive? Could you add a

nfischer:
LGTM (in case lgtm.co missed it)

ariporad:
Thanks @isiahmeadows!

And yeah that was an oversight, At the time we added it, our windows build was broken.

isiahmeadows:
@ariporad Okay.

PR #293, opened 2016-01-19
feat: add tilde expansion to expand()

nfischer:
This adds tilde expansion to the expand() function. Arguments starting with '~/'
will have the tilde expanded to the user's home directory, as with Bash.

Also, if an argument is simply `~`, then it will expand to the home directory as well.

```Javascript
> cd('~'); // Go to home directory
> ls('~/*.js'); // Look for all .js files
> cd('~/..'); // Go to /home, /Users, or C:\Users
```

This does not support all of Bash's tilde expansion (not sure if this would be difficult). For example:

```Bash
$ ls ~otheruser # look into otheruser's home folder
$ # Nothing like this is supported in ShellJS
```

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -53,6 +53,18 @@ function ShellString(str) {
 }
 exports.ShellString = ShellString;
 
+// Return the home directory in a platform-agnostic way, with consideration for
+// older versions of node
+function getUserHome() {
+  var result;
+  if (os.homedir)
+    result = os.homedir(); // node 3+
+  else
+    result = process.env[(process.platform === 'win32') ? 'USERPROFILE' : 'HOME'];
+  return result;
+}
+exports.getUserHome = getUserHome;
+
 // Returns {'alice': true, 'bob': false} when passed a dictionary, e.g.:
 //   parseOptions('-a', {'a':'alice', 'b':'bob'});
 function parseOptions(str, map) {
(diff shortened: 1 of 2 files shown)


nfischer:
This works right now with the old `expand()` function. If we rework this function to use the `glob` module (#275), this should still be compatible.

This still needs to be tested on Windows.

nfischer:
For future reference, this prefers `os.homedir` when it's available, which I can see in node v3.0.0 (https://github.com/nodejs/node/pull/1791). For earlier versions, this relies on environmental variables, which appears to work well.

src/common.js
```diff
@@ -53,6 +53,20 @@ function ShellString(str) {
 }
 exports.ShellString = ShellString;
 
+// Return the home directory in a platform-agnostic way, with consideration for
+// older versions of node
+function getUserHome() {
```
ariporad: Why not just use [`os-homedir`](https://github.com/sindresorhus/os-homedir) instead? I don't really feel like we need to reinvent the wheel.

nfischer: In terms of implementation, this should be very similar. I opted to just code it myself because it cuts out a dependency and it was only a few lines of code. I think it's easier to keep small snippets in the code base.

ariporad:
Did this ever get tested on Windows?

src/common.js
ariporad: Ok

nfischer:
@ariporad good catch! I did test it on my Windows 7 machine. It passes the `test/cd.js` unit tests. I also ran a simple shell script myself, and it behaves exactly as I would expect.

```Javascript
cd('~');
echo(pwd());
cd('~/Documents'); // still works, evven with a forward slash
echo(pwd());
cd('~');
echo(pwd());
```

nfischer:
That was testing on Windows 7 64 bit, Node v5, running through `cmd.exe`

ariporad:
LGTM!

Issue #294, opened 2016-01-19
run-tests.js is broken for cmd.exe


nfischer:
I was trying out `run-tests.js` on my Windows machine, and ran into *lots* of problems. Since we're trying to get good support for Windows, I figured I'd make this a formal issue so that others can chime in and confirm the behavior I'm seeing.

The end-goal of this issue is to open up `cmd.exe` on Windows, `cd` to the shelljs directory, type `npm test`, and see that all the tests pass. To me this seems like a worthy goal.

With that said, if someone else sees behavior different from what I'm seeing, please verify that you're running everything through `cmd.exe`, like I am. I already know this can run on some other shells (Bash, for instance), but I'd like to see this work on this particular shell.

Initial problems I've seen:

 - must use `node path/to/jshint` in the `run-tests` script, instead of relying on the shebang
 - must expand wildcards ourselves, not relying on the shell, since [cmd relies on the application to expand them](https://superuser.com/questions/460598/is-there-any-way-to-get-the-windows-cmd-shell-to-expand-wildcard-paths), and jshint doesn't appear to be doing this
 - `exec()` will start in the current working directory if run through `cmd.exe`. If we pass the `cwd` argument to `child_process.exec()`, then this will fix the issue only for admin mode (not regular user mode). I think this is only possible to fix for admin mode. This means that all our exec calls will need `pwd()` prefixed to the paths.

I'll add more to the list in comments as I find the issues.

nfischer:
I'll open a PR later that at least gets jshint working properly (and should work on both Windows and Linux).

mtscout6:
Have you considered getting a CI setup with [AppVeyor](https://www.appveyor.com/)? It's a free service like travis that runs on Windows machines. It also supports GitHub integration like Travis for Pull Requests so you know they work with both Windows and Linux based systems.

nfischer:
Yes! We're working on setting up AppVeyor (although this issue sounds like it needs to be fixed before AppVeyor can be useful). If you have any experience writing an AppVeyor configure file, help would be appreciated :+1:

mtscout6:
I did one for [react-bootstrap](https://github.com/react-bootstrap/react-bootstrap) which is:

https://github.com/react-bootstrap/react-bootstrap/blob/master/appveyor.yml

```

version: '{build}'

# Install scripts. (runs after repo cloning)
install:
  # Get the latest stable version of io.js
  - ps: Install-Product node ''
  - npm -g install npm@latest
[...]

PR #295, opened 2016-01-19
fix: jshint works on Windows

nfischer:
`run-tests.js` previously relied on shell wildcard expansion, and did not specify
full paths. This uses common.expand to handle globbing internally, and specifies
the full path, so jshint can find all the files.

This **does not** entirely fix `run-tests`, but this does fix the jshint portion, so we'll at least get it to process javascript files. A full fix would be much more intensive, since `exec()` can't start from the current working directory. So this is only a partial fix for #294.

Also, this adds in lint-checking the `scripts/` folder (and fixes the lint errors I found!). Now the only significant folder this skips is the `src/` folder (which I'm working on providing support for).

### DIFF
--- a/scripts/generate-docs.js
+++ b/scripts/generate-docs.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+/* globals cd, echo, grep, sed */
 require('../global');
 
 echo('Appending docs to README.md');
--- a/scripts/run-tests.js
+++ b/scripts/run-tests.js
@@ -1,14 +1,14 @@
 #!/usr/bin/env node
+/* globals cd, echo, exec, exit, ls, pwd, test */
 require('../global');
-
-var path = require('path');
+var common = require('../src/common');
 
 var failed = false;
 
 //
 // Lint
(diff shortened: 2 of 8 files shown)


nfischer:
Update: I fixed the lint errors in `src/`, so now this will go through and do lint checking on `src/*.js` as well.

nfischer:
@ariporad is this safe to merge?

ariporad:
LGTM!

ariporad:
@nfischer: This gave me an idea (if you like it, I'll open a separate issue): What if we wrote an `ESLint` plugin/preset for shelljs, which automatically included the needed globals. (Then we could switch the project itself to eslint).

nfischer:
@ariporad I don't have much experience with eslint. That sounds like it'd be useful though (since it can be somewhat annoying to declare in a comment which globals are used). I think this project relies on jshint partially because it uses shelljs as a dependency. So far, it seems fairly easy to stick with jshint, so I say we don't prioritize the switch too much.

Issue #296, opened 2016-01-19
Failing tests on Windows


BYK:
There are failing tests on Windows. My system is Node 4.2.4 with Windows 10:

```
BYK@ ~▶Documents▶Projects▶shelljs⛕ master$ npm test

> shelljs@0.5.3 test C:\Users\BYK\Documents\Projects\shelljs
> node scripts/run-tests

'.' is not recognized as an internal or external command,
operable program or batch file.
*** JSHINT FAILED! (return code != 0)

Running test: cat.js
Running test: cd.js
Running test: chmod.js

assert.js:89
  throw new assert.AssertionError({
  ^
AssertionError: 438 == 493
    at Object.<anonymous> (C:\Users\BYK\Documents\Projects\shelljs\test\chmod.js:23:8)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:136:18)
    at node.js:963:3
*** TEST FAILED! (missing exit code "123")

Running test: common.js
Running test: config.js
Running test: cp.js

assert.js:89
  throw new assert.AssertionError({
  ^
AssertionError: 0 == 1
    at Object.<anonymous> (C:\Users\BYK\Documents\Projects\shelljs\test\cp.js:145:8)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:136:18)
    at node.js:963:3
*** TEST FAILED! (missing exit code "123")

Running test: dirs.js
Running test: echo.js
Running test: env.js
Running test: exec.js
Running test: find.js
Running test: grep.js
Running test: ln.js

assert.js:89
  throw new assert.AssertionError({
  ^
AssertionError: false == true
    at Object.<anonymous> (C:\Users\BYK\Documents\Projects\shelljs\test\ln.js:51:1)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:136:18)
    at node.js:963:3
*** TEST FAILED! (missing exit code "123")

[...]

PR #297, opened 2016-01-19
Fix tests on Windows

### LINKED ISSUE #294: run-tests.js is broken for cmd.exe

BYK:
Fixes #294 partially and #296.

### DIFF
--- a/src/ln.js
+++ b/src/ln.js
@@ -50,7 +50,7 @@ function _ln(options, source, dest) {
     if (!fs.existsSync(source)) {
       common.error('Source file does not exist', true);
     }
-    fs.linkSync(source, dest, os.platform() === "win32" ? "junction" : null);
+    fs.linkSync(source, dest);
   }
 }
 module.exports = _ln;
--- a/test/cp.js
+++ b/test/cp.js
@@ -1,4 +1,5 @@
 var shell = require('..');
+var common = require('../src/common');
 
 var assert = require('assert'),
     fs = require('fs'),
@@ -139,11 +140,14 @@ shell.cp('-r', 'resources/cp/dir_a', 'tmp/dest');
 assert.equal(shell.error(), null);
 assert.equal(fs.existsSync('tmp/dest/z'), true);
 
-//preserve mode bits
-shell.rm('-rf', 'tmp/*');
-var execBit = parseInt('001', 8);
-assert.equal(fs.statSync('resources/cp-mode-bits/executable').mode & execBit, execBit);
-shell.cp('resources/cp-mode-bits/executable', 'tmp/executable');
-assert.equal(fs.statSync('resources/cp-mode-bits/executable').mode, fs.statSync('tmp/executable').mode);
+// On Windows, permission bits are quite different so skip those tests for now
+if (common.platform !== 'win') {
+    //preserve mode bits
+    shell.rm('-rf', 'tmp/*');
+    var execBit = parseInt('001', 8);
+    assert.equal(fs.statSync('resources/cp-mode-bits/executable').mode & execBit, execBit);
+    shell.cp('resources/cp-mode-bits/executable', 'tmp/executable');
+    assert.equal(fs.statSync('resources/cp-mode-bits/executable').mode, fs.statSync('tmp/executable').mode);
+}
 
 shell.exit(123);
--- a/test/pushd.js
+++ b/test/pushd.js
@@ -190,31 +190,31 @@ assert.equal(process.cwd(), trail[0]);
 reset(); trail = shell.pushd('resources/pushd');
 assert.equal(shell.error(), null);
 assert.equal(trail.length, 2);
-assert.equal(path.relative(root, trail[0]), 'resources/pushd');
+assert.equal(path.relative(root, trail[0]), path.join('resources', 'pushd'));
 assert.equal(trail[1], root);
 assert.equal(process.cwd(), trail[0]);
 trail = shell.pushd();
 assert.equal(shell.error(), null);
 assert.equal(trail.length, 2);
 assert.equal(trail[0], root);
-assert.equal(path.relative(root, trail[1]), 'resources/pushd');
+assert.equal(path.relative(root, trail[1]), path.join('resources', 'pushd'));
 assert.equal(process.cwd(), trail[0]);
 
 // Push without arguments should swap top two directories when stack length is > 2
 trail = shell.pushd('resources/pushd/a');
 assert.equal(shell.error(), null);
 assert.equal(trail.length, 3);
-assert.equal(path.relative(root, trail[0]), 'resources/pushd/a');
+assert.equal(path.relative(root, trail[0]), path.join('resources', 'pushd', 'a'));
 assert.equal(trail[1], root);
-assert.equal(path.relative(root, trail[2]), 'resources/pushd');
+assert.equal(path.relative(root, trail[2]), path.join('resources', 'pushd'));
 assert.equal(process.cwd(), trail[0]);
 
 trail = shell.pushd();
 assert.equal(shell.error(), null);
 assert.equal(trail.length, 3);
 assert.equal(trail[0], root);
-assert.equal(path.relative(root, trail[1]), 'resources/pushd/a');
-assert.equal(path.relative(root, trail[2]), 'resources/pushd');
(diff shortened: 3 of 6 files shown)


src/ln.js
```diff
@@ -47,7 +47,7 @@ function _ln(options, source, dest) {
   if (options.symlink) {
     fs.symlinkSync(source, dest, os.platform() === "win32" ? "junction" : null);
   } else {
-    fs.linkSync(source, dest, os.platform() === "win32" ? "junction" : null);
+    fs.linkSync(source, dest);
```
BYK: This was a bug since the last option doesn't exist in `linkSync` at all.

src/ln.js
```diff
@@ -47,7 +47,7 @@ function _ln(options, source, dest) {
   if (options.symlink) {
     fs.symlinkSync(source, dest, os.platform() === "win32" ? "junction" : null);
```
BYK: We should only use `"junction"` if we know the source is a directory. Gonna handle this separately since we never test symlinking for directories.

BYK:
This commit is mostly "if Windows, skip these tests" but unfortunately Windows is not POSIX compatible so certain things don't even work on Windows (`chmod` for instance). I'll try to make symlinking work properly first and then see if I can get `chmod` working.

Tests for `test -L` is also tricky because `git` on Windows checks symlinks as normal files with symlinked filename as the content making it impossible to use the links in `resources` directory on Windows. Once/if `ln -s` is fixed, we may use that to create links and then test them.

src/ln.js
nfischer: Looks like the docs agree :+1:

test/test.js
```diff
@@ -63,24 +64,27 @@ assert.equal(shell.error(), null);
...
-assert.equal(result, true);//true
-
-var result = shell.test('-L', 'resources/badlink');
-assert.equal(shell.error(), null);
-assert.equal(result, true);//true
-
-var result = shell.test('-L', 'resources/404');
-assert.equal(shell.error(), null);
-assert.equal(result, false);//false
+// Windows is weird with links so skip these tests
```
nfischer: Is there no way to support links on Windows? If we can't find a way to make `ln` behave properly on Windows, then we'll need to document it as not running on Windows. It'd be preferable if we could fix the behavior and get at least some links to work.

test/pushd.js
```diff
@@ -190,31 +190,31 @@ assert.equal(process.cwd(), trail[0]);
...
 assert.equal(process.cwd(), trail[0]);
 
 trail = shell.pushd();
 assert.equal(shell.error(), null);
 assert.equal(trail.length, 3);
 assert.equal(trail[0], root);
-assert.equal(path.relative(root, trail[1]), 'resources/pushd/a');
-assert.equal(path.relative(root, trail[2]), 'resources/pushd');
+assert.equal(path.relative(root, trail[1]), path.join('resources', 'pushd', 'a'));
+assert.equal(path.relative(root, trail[2]), path.join('resources', 'pushd'));
```
nfischer: :+1: nice!

test/cp.js
```diff
@@ -139,11 +140,14 @@ shell.cp('-r', 'resources/cp/dir_a', 'tmp/dest');
...
 assert.equal(fs.existsSync('tmp/dest/z'), true);
 
-//preserve mode bits
-shell.rm('-rf', 'tmp/*');
-var execBit = parseInt('001', 8);
-assert.equal(fs.statSync('resources/cp-mode-bits/executable').mode & execBit, execBit);
-shell.cp('resources/cp-mode-bits/executable', 'tmp/executable');
-assert.equal(fs.statSync('resources/cp-mode-bits/executable').mode, fs.statSync('tmp/executable').mode);
+// On Windows, permission bits are quite different so skip those tests for now
+if (common.platform !== 'win') {
```
nfischer: What's different about permissions on Windows? I know executable permissions are a bit different, but do read/write permissions work as expected?

test/test.js
BYK: Hey, see my other comments - especially https://github.com/shelljs/shelljs/pull/297#discussion_r50105148

This is a bug, long existed in shelljs but it is large enough to deserve a separate bug and PR IMO.

test/cp.js
BYK: >What's different about permissions on Windows? I know executable permissions are a bit different, but do read/write permissions work as expected?

Please see the following:
- https://github.com/nodejs/node-v0.x-archive/issues/4812
- http://stackoverflow.com/a/23330010/90297

test/test.js
nfischer: I agree, if you could open a github issue for the bug, that would be nice to have documented. And if we can get that fixed, even better!

BYK: Done: #301.

I'm willing to take this on. Shouldn't be too hard.

test/cp.js
nfischer: I see. So it sounds like something we should document as not being well supported on Windows. If we could update the docs to specify which features of chmod work on Windows, that'd be great.

BYK: Agreed and ticketed: #302

test/test.js
nfischer: Awesome! :+1:

nfischer:
@BYK looks like there's some merge conflicts. Could you resolve?

@ariporad Do you think we should merge, or wait until `chmod` and `ln` have their Windows-incompatibility documented? This blocks #299, which is a priority.

mtscout6:
I'd advocate for merging #299 first to ensure these fixed tests work in AppVeyor before merging.

BYK:
@nfischer done

nfischer:
@BYK could you rebase off master one more time? That will make sure AppVeyor runs checks on the code

BYK:
@nfischer done :)

nfischer:
@BYK could you modify `test/shjs.js`? I think we need to change:

> ```Javascript
>   return shell.exec(path.resolve(__dirname, '../bin/shjs') +
>                     ' ' +
>                     path.resolve(__dirname, 'resources', 'shjs', name), { silent: true });
> ```

into:

> ```Javascript
>   return shell.exec((process.platform === 'win32' ? 'node ' : '') + // call by path for unix, call with 'node' for Windows
>                     path.resolve(__dirname, '../bin/shjs') +
>                     ' ' +
>                     path.resolve(__dirname, 'resources', 'shjs', name), { silent: true });
> ```

I think the issue is that Windows can't call the .js file directly, while we expect OSX/Linux to be able to.

nfischer:
@BYK Also, awesome job fixing the other tests! It's great to finally have (almost) all the tests passing on Windows.

nfischer:
@BYK or you can just rebase off of the branch in #315 (or off master if it gets merged). This contains the fix I wrote above for the `shjs` test.

nfischer:
I'm going to go ahead and merge this, since I tested it in a separate branch, and I tested it rebased off master and it passes appveyor CI now that #315 is merged :+1:

nfischer:
@BYK Thanks again for all your help! This rocks!

BYK:
@nfischer - glad I could help. I'm planning to help with other Windows issues too. Just don't have a much time. May be over the weekend :)

PR #303, opened 2016-01-24
fix: regexes are more consistent with sed and grep

nfischer:
`sed()` will now convert search strings to regex form, so `'a*'` will now work like `/a*/`. Also, new tests for grep and sed ensure that `*` is not expanded for filename globbing.

While the filename globbing isn't currently an issue, I anticipate that it could be a problem if we don't handle `sed()` and `grep()` carefully when doing filename globbing as part of the `wrap()` function (which I think is the way this project should go). These tests are an extra guard to make sure we don't make any mistakes.

So this should probably be merged before #275 to be safe.

### DIFF
--- a/src/sed.js
+++ b/src/sed.js
@@ -28,13 +28,20 @@ function _sed(options, regex, replacement, file) {
   else
     common.error('invalid replacement string');
 
+  // Convert all search strings to RegExp
+  if (typeof regex === 'string')
+    regex = RegExp(regex);
+
   if (!file)
     common.error('no file given');
 
   if (!fs.existsSync(file))
     common.error('no such file or directory: ' + file);
 
-  var result = fs.readFileSync(file, 'utf8').replace(regex, replacement);
+  var result = fs.readFileSync(file, 'utf8').split('\n').map(function (line) {
+    return line.replace(regex, replacement);
+  }).join('\n');
+
   if (options.inplace)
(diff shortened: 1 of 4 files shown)


src/sed.js
```diff
@@ -28,13 +28,21 @@ function _sed(options, regex, replacement, file) {
...
   if (!file)
     common.error('no file given');
 
   if (!fs.existsSync(file))
     common.error('no such file or directory: ' + file);
 
-  var result = fs.readFileSync(file, 'utf8').replace(regex, replacement);
+  var result_list = fs.readFileSync(file, 'utf8').split('\n').map(function (line) {
+    return line.replace(regex, replacement);
+  });
```
ariporad: [Slight Nitpick]: Maybe just simplify to:
```javascript
var result = fs.readFileSync(file, 'utf8').split('\n').map(function (line) {
    return line.replace(regex, replacement);
}).join('\n');
```

ariporad:
One small nitpick, but otherwise LGTM!

src/sed.js
nfischer: I originally wrote it like that. I switched it because I thought we might later want to modify sed to return a list (similar what ls() does) instead of return a multi line string. I can revert this back though

nfischer:
@ariporad updated to address your comment. Feel free to merge once tests pass.

ariporad:
Merged

PR #304, opened 2016-01-24
test(shjs): add tests for shjs

### LINKED ISSUE #280: Add tests for the shjs utility
It might be useful to add some tests for the `shjs` utility itself. This would include making sure it respects stdout/stderr and exit codes. This is less of a priority now, but would be more useful if #268 (stderr) and #252 (exit codes) are merged.

This would involve creating shelljs scripts and executing them with the `shjs` command, and making sure results are as expected. I don't think we need a large number of test cases, but we should have a few just to make sure we don't get regressions in the utility.

ariporad:
Fixes  #280.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   },
   "dependencies": {},
   "devDependencies": {
+    "coffee-script": "^1.10.0",
     "jshint": "~2.1.11"
   },
   "optionalDependencies": {},
--- a/test/resources/shjs/exit-codes.js
+++ b/test/resources/shjs/exit-codes.js
@@ -1,0 +1,1 @@
+process.exit(42);
--- a/test/shjs.js
+++ b/test/shjs.js
@@ -1,0 +1,29 @@
+var shell = require('..');
+var path = require('path');
+var assert = require('assert');
+
+function runScript(name) {
+  return shell.exec(path.resolve(__dirname, '../bin/shjs') +
+                    ' ' +
+                    path.resolve(__dirname, 'resources', 'shjs', name), { silent: true });
+}
+
+// Exit Codes
+assert.equal(runScript('exit-codes.js').code, 42, 'exit code works');
+assert.equal(runScript('exit-0.js').code, 0, 'exiting 0 works');
+
+// Stdout/Stderr
+var stdioRet = runScript('stdout-stderr.js');
+assert.equal(stdioRet.stdout, 'stdout: OK!\n', 'stdout works');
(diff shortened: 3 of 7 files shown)


ariporad:
@nfischer: ready for review.

test/resources/shjs/exit-codes.js
```diff
@@ -0,0 +1 @@
+process.exit(42);
```
nfischer: Can we add a test with return code `0`?

ariporad: Sure

package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {},
   "devDependencies": {
+    "coffee-script": "^1.10.0",
```
nfischer: This dependency seems reasonable to me :+1:

test/shjs.js
```diff
@@ -0,0 +1,19 @@
...
+function runScript(name) {
+  return shell.exec(path.resolve(__dirname, '../bin/shjs') +
+                    ' ' +
+                    path.resolve(__dirname, 'resources', 'shjs', name), { silent: true });
+}
+
+assert(runScript('exit-codes.js'), 42, 'exit code works');
+
+var stdioRet = runScript('stdout-stderr.js');
+assert(stdioRet.stdout, 'stdout: OK!\n', 'stdout works');
```
nfischer: s/assert/assert.equal

Also, I don't think it takes a third parameter

test/shjs.js
```diff
@@ -0,0 +1,19 @@
...
+assert(stdioRet.stderr, 'stderr: OK!\n', 'stderr works');
+
+assert(runScript('coffeescript.coffee').stdout, 'CoffeeScript: OK!\n');
```
nfischer: assert.equal, like above. Also, perhaps we could check its error code to make sure it's 0?

test/resources/shjs/exit-codes.js
nfischer: (or just see my comment below)

nfischer:
Could we add a test to make sure it correctly infers the file extension? Perhaps `shjs path/to/resource/file` for a file named `path/to/resource/file.js`?

nfischer:
Other than that, LGTM

test/shjs.js
ariporad: Yes it does. (All assert methods do), it's a error message if the assert fails.

ariporad:
@nfischer: Should have fixed all those, feel free to re-review/merge.

nfischer:
Awesome!

PR #306, opened 2016-01-24
feat(cd): cd() (no args) changes to home directory

nfischer:
`cd()` used to be a error. Now it changes to the home directory instead. This is to match unix behavior:

```Bash
$ pwd # prints whatever the cwd is
$ cd
$ pwd # prints the home directory path
```


ariporad:
LGTM!

PR #307, opened 2016-01-24
travis: Mark as not using `sudo` and do not test 0.11

TimothyGu:


TimothyGu:
This makes Travis running on Linux significantly more consistent (all finishing within 20 seconds) and faster (with two fewer builds). Read more about `sudo: false` in Travis CI's documentation: https://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F

nfischer:
@TimothyGu why remove v0.11? I believe we want to continue support for that if possible

TimothyGu:
0.11 is an unstable version that is not supported or updated anymore. 0.12 is fully compatible with 0.11 and as it is in maintenance mode, no breaking changes will be committed there.

TimothyGu:
And we *should* test 0.8 too since it is listed as supported in package.json.

nfischer:
> 0.12 is fully compatible with 0.11

Good point. @ariporad any thoughts? Slightly faster CI, which is nice. And it sounds like it should be safe.

For 0.8 issue, I think it's because jshint is not supported. We should probably remove 0.8 from the package.json in that case. @TimothyGu thanks for pointing that out!

TimothyGu:
Yeah we should remove 0.8 now. It's last release is over 1.5 years old and its first 3.5 yrs old.

ariporad:
LGTM!

PR #308, opened 2016-01-24
chore(package): remove v0.8 from engines list

nfischer:
We don't really guarantee our stuff works on v0.8, since travis doesn't run on that version (this is because jshint requires v0.10+). So this just removes it from package.json. This doesn't prevent v0.8 users from using shelljs, just removes the promise of support.

See #307 for a discussion about this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,6 @@
   },
   "optionalDependencies": {},
   "engines": {
-    "node": ">=0.8.0"
+    "node": ">=0.10.0"
   }
 }


ariporad:
I don't think we support anything less than <0.12, do we? (I think travis only builds 0.12 and up)

nfischer:
@ariporad the travis config uses `nvm` to install the version of node. `nvm` supports each of those versions.

This can be verified by looking at the travis output where it says `node --version`.

ariporad:
@nfischer: no, I know, but is *our* build set up to build on node 0.10? (Sorry, on mobile, so I can't check)

nfischer:
@ariporad Yeah, our travis config uses v0.10, 0.11, 0.12, 4, and 5

ariporad:
Ok then. LGTM!

Issue #309, opened 2016-01-25
Gen-docs is broken


nfischer:
Oops! It turns out that #303 modified the behavior of `sed()` in a way that didn't play nicely with `scripts/generate-docs.js`. The docs script uses a regex `/## Command reference(.|\n)*/`, which was previously used to mean "replace everything from '## Command reference' until the very end of the file."

Unix `sed`, however, does not operate on multiple lines, so it only replaces the first line, leaving all the old docs concatenated onto the end of the new docs.

ariporad:
Fixed.

PR #310, opened 2016-01-25
fix(gen-docs): fix issue where docs are generated wrong

nfischer:
This fixes an issue (#309) where the gen-docs script's sed() command would improperly
update the docs.

As a bonus, Travis now checks that docs are generated properly (or, more specifically, that no files have been changed by running the tests and re-generating docs, which is exactly what we want). This prevents breaking changes in `sed()` from going unnoticed. This will also prevent people from modifying doc-comments in the source files without generating docs, or by modifying the README manually without changing comments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,5 @@ os:
   - osx
 script:
   - npm test
+  - node scripts/generate-docs.js
+  - git diff --quiet # make sure no files have changed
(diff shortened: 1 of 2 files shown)


nfischer:
@ariporad could you take a look at this when you get the chance?

nfischer:
Also, we could consider switching to only one `replace()` (no `sed()`). The nice thing about keeping `sed()` there is that this tests it out to catch any unintended change, so I'm inclined to keep `sed()` in there (efficiency isn't a concern anyway).

ariporad:
LGTM

PR #315, opened 2016-01-27
fix(windows): fix shjs commands for windows

nfischer:
Prefix `shjs` commands with`node ` for Windows, since Windows doesn't use shebangs.

Blocks #297


nfischer:
@ariporad I think it's ok to merge this with the broken tests, since it fixes one of them (and then #297) should be safe to merge.

ariporad:
LGTM!

PR #317, opened 2016-01-29
chore: switch from JSHint to ESLint

ariporad:
This PR switches us from JSHint to ESLint.

ESLint checks both lint errors and style.

We're using [AirBnB's Style Guide](https://github.com/airbnb/javascript) as a starting point.

Since there are about 1700 lint errors, which I couldn't possibly hope to do myself, this PR is going to be done slightly differently. While this PR will remain open, Me and @nfischer will accept PRs against ***this*** branch fixing a small number of lint errors (Please do one file per PR). Then, once the number of lint errors is 0, we'll squash everything and merge into master.

Current Lint Problem Count for this branch: `794`


nfischer:
@ariporad is this up to date with master? It looks like it's failing Windows CI.

Also, it looks like something is undefined in `run-tests.js`

ariporad:
@nfischer: It's up to date with master, but the lint stage is failing ***badly***

ariporad:
@nfischer: Oh, actually, `spawnSync` isn't available on node v0.10. Fixed.

nfischer:
@ariporad I changed the milestone, since I don't think we should push off 0.6 just so we can switch lint tools

.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
```
BYK: We should prefer `.eslintrc.json` as the file name. "raw" `.eslintrc` files are deprecated.

BYK:
I'm willing to help with this. Have you tried using the `--fix` option to get some of these fixed autoamtically?

.eslintrc
ariporad: Good catch

ariporad:
@BYK: Oops... I forgot about that.

ariporad:
@BYK: Fixed.

BYK:
Adding the following to the rule set helps a lot:

    "spaced-comment": [2, "always", { "markers": ["@"], "exceptions": ["@"] }]

BYK:
Looks like this is the problem:

    /home/travis/build/shelljs/shelljs/test/which.js
    18:5  error  "result" is already defined  no-redeclare

BYK:
FYI I didn't try to fix all warnings so the warnings are expected. `no-param-reassign` is extensively violated and requires some fundamental changes in a few places so I didn't bother fixing them.

Also I'm not sure AirBnb style guide was a good choice since it looks a bit too browser oriented.

BYK:
>(Please do one file per PR)

lol - sorry for not following this :D

ariporad:
Don't worry, it didn't end up making a difference.

TimothyGu:
Umm… is there a reason why this is high priority?

nfischer:
I feel like this should be a low priority issue. Lengthy chores probably shouldn't be high priority. I think only in-demand features and bug fixes (and issues blocking those) should be high priority

ariporad:
Oh, yeah, that was a mistake. Fixed.

nfischer:
@ariporad any further work on this? This might be nice to start transitioning to using.

PR #321, opened 2016-01-29
Rewrite .gitignore to be more comprehensive

BYK:
Ignores more Node/NPM files and adds ignore rules for the following:

- Linux
- OS X
- Vim
- JetBrains IDEs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,19 @@
-node_modules/
-tmp/
+# Project
+npm-debug.log*
+node_modules
+tmp
+
+# Linux
+*~
+
+# JetBrains
+.idea
+
+# Vim
+## swap
+[._]*.s[a-w][a-z]
+[._]s[a-w][a-z]
+
+# OS X
+.DS_Store
+


nfischer:
I think it would be more useful to have one PR that ignores basic ignore files, like in swap files and such as well. I can write that out, but you're more than welcome to add them yourself

nfischer:
I think adding these should be good enough:

```
# Mac files
.DS_Store

# Vim files
[._]*.s[a-w][a-z]
*~
```

BYK:
Well, since you asked for it, I did a much more comprehensive edit (a rewrite at this point) on the file. Updated the commit and PR messages accordingly. What  do you think?

nfischer:
@BYK this might be a little too much for my taste. I think less is more. We haven't had an issue of accidentally committing editor files, so it may be best to put this off until the problem arises.

BYK:
@nfischer - alright, removed a bunch of stuff. What do you think? :)

ariporad:
@BYK: Hmm... What about [`gitignore.io`](http://gitignore.io)?

That would probably be more comprehensive, but I'm not really super strict either way. So LGTM. @nfischer: What do you think? I'm happy either way, so feel free to merge.

nfischer:
LGTM

PR #322, opened 2016-01-29
Fix symlinking on Windows

### LINKED ISSUE #301: `link -s` is broken for files on Windows
Currently we simply use `junction` as the linking method on Windows do avoid any linking permissions that Windows has (requires admin permissions to link a file). This though only works for directories so we need to check the source item type and use the proper link method accordingly.

BYK:
Fixes symlink creation on Windows systems and adds a test
for symlinking directories since it needs special treatment
on Windows.

Fixes #301.

### DIFF
--- a/src/ln.js
+++ b/src/ln.js
@@ -1,7 +1,6 @@
 var fs = require('fs');
 var path = require('path');
 var common = require('./common');
-var os = require('os');
 
 //@
 //@ ### ln([options,] source, dest)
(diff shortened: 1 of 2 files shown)


BYK:
Gonna fox POSIX issues.

BYK:
Okay, should be fixed now.

nfischer:
I haven't tested this on my windows machine, but it looks like it won't break linux/mac behavior, and it definitely looks like an improvement for Windows behavior. Since it passes all the CI, LGTM.

BYK:
Rebased

ariporad:
LGTM!

ariporad:
Thanks @BYK!

Issue #323, opened 2016-01-29
Feature request: Metadata with `ls`


gvn:
Hi there. First off, I'm a big fan of this library!

Unless I'm missing something, there's no way to read the modification date of a file (among other things). I'm having to go back into node's `fs` module for that. It seems like this would be very useful. It strikes me that it could be added in an option for `ls`.

EG: `ls('-l', 'path')` might return an array of objects with file metadata such as permissions, owner, group, size, and modification date.

nfischer:
Sounds good to me. I think I'll open a PR for ls -l later today, and it'll be based off this design

gvn:
Awesome! I'm happy to give it a code review if you'd like. I could probably also write a PR soon-ish, but my guess is you'd be way faster at adding it. Just let me know! :smiley:

nfischer:
@gvn I'll tag you in the PR once it's done :+1:

nfischer:
@gvn I tagged you in the PR

PR #324, opened 2016-01-30
feat(ls): add -l  option

nfischer:
The `-l` option will now cause `ls()` to return an object containing file stats.
These objects will also have a toString() method that formats it into something
analogous to `ls -l`'s output format.

This is useful in the case of `ls('-l', 'dir/').join('\n')`, which will now have formatting similar to unix's `ls -l`.

The mechanics for `-l` are based off #137. This resolves feature request #323.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,11 @@ Available options:
 
 + `-R`: recursive
 + `-A`: all files (include files beginning with `.`, except for `.` and `..`)
++ `-d`: list directories themselves, not their contents
++ `-l`: list objects representing each file, each with fields containing `ls
+        -l` output fields. See
+        [fs.Stats](https://nodejs.org/api/fs.html#fs_class_fs_stats)
+        for more info
 
 Examples:
 
@@ -179,6 +184,7 @@ Examples:
 ls('projs/*.js');
 ls('-R', '/users/me', '/tmp');
 ls('-R', ['/users/me', '/tmp']); // same as above
+ls('-l', 'file.txt'); // { name: 'file.txt', mode: 33188, nlink: 1, ...}
 ```
 
 Returns array of files in the given path, or in current directory if no path provided.
--- a/src/ls.js
+++ b/src/ls.js
@@ -11,6 +11,11 @@ var _pwd = require('./pwd');
 //@
 //@ + `-R`: recursive
 //@ + `-A`: all files (include files beginning with `.`, except for `.` and `..`)
+//@ + `-d`: list directories themselves, not their contents
+//@ + `-l`: list objects representing each file, each with fields containing `ls
+//@         -l` output fields. See
+//@         [fs.Stats](https://nodejs.org/api/fs.html#fs_class_fs_stats)
+//@         for more info
 //@
 //@ Examples:
 //@
@@ -18,6 +23,7 @@ var _pwd = require('./pwd');
 //@ ls('projs/*.js');
 //@ ls('-R', '/users/me', '/tmp');
 //@ ls('-R', ['/users/me', '/tmp']); // same as above
+//@ ls('-l', 'file.txt'); // { name: 'file.txt', mode: 33188, nlink: 1, ...}
 //@ ```
 //@
 //@ Returns array of files in the given path, or in current directory if no path provided.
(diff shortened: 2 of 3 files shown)


nfischer:
@gvn @ariporad this should be ready for review.

I believe Windows does not support `uid` and `gid` for `fs.statSync()` so I'm merely ignoring these values in the tests.

Here's a comparison of output from `ls -l` and `ls('-l', ...)`:

```
$ ls -l tmp/54.js
-rw-rw-r-- 1 nate nate 54 Jan 29 15:58 tmp/54.js

> ls('-l', 'tmp/54.js').toString();
33204 1 1000 1000 54 Fri Jan 29 2016 15:58:26 GMT-0800 (PST) tmp/54.js
```

The dates are javascript dates. The uid and gid are ID numbers, not username/groupname ([this npm module](https://github.com/jandre/node-userid) uses C code to translate ID to name). The modes are in whatever representation `fs.stat()` returns them. Although the appearance isn't identical, I think using the JS types will benefit users more.

src/ls.js
```diff
@@ -128,3 +149,16 @@ function _ls(options, paths) {
   return list;
 }
 module.exports = _ls;
+
+
+function ls_stat(path) {
+  var stats = fs.statSync(path);
+  // Note: this object will contain more information than .toString() returns
+  var long_stats = Object.create(stats);
+  long_stats.name = path;
```
gvn: I'd suggest making this value just the filename instead of the entire path, and then adding another property called `path` with the full path. Seems a little clearer that way IMO.

src/ls.js
```diff
@@ -128,3 +149,16 @@ function _ls(options, paths) {
   return list;
 }
 module.exports = _ls;
+
+
+function ls_stat(path) {
+  var stats = fs.statSync(path);
+  // Note: this object will contain more information than .toString() returns
+  var long_stats = Object.create(stats);
```
gvn: One slight drawback to using `Object.create()` is that the stats are on the object's prototype, which makes them invisible in a console.log of the object. 

It might be better to append the `stats` properties directly to `long_stats` to make things slightly clearer for debugging.

src/ls.js
```diff
@@ -9,6 +9,7 @@ var _pwd = require('./pwd');
 //@ ### ls([options,] path_array)
 //@ Available options:
 //@
+//@ + `-l`: long list format
```
gvn: I'd recommend being a little more verbose here especially if you stick with using Object.create below...

gvn:
I had a couple thoughts, but otherwise this looks great!

Thanks for jumping on this so quickly. :+1:

src/ls.js
nfischer: Good Idea. I'll fix this

src/ls.js
nfischer: Good call. I'll update this to be slightly more verbose.

src/ls.js
nfischer: I see the benefit, but it could be slightly trickier that way. `ls -l` returns output with the relative path to the file, so that's what I'm opting for with this. If users need the specific name, they can use `path.basename()` to get it. If they need the full path, I think `path.resolve()` would do it.

I could use these functions to set the internal properties, but it's harder to use the full path and basename to get the relative path for the `toString()` method.

README.md
```diff
@@ -172,6 +172,9 @@ Available options:
 
 + `-R`: recursive
 + `-A`: all files (include files beginning with `.`, except for `.` and `..`)
++ `-d`: list directories themselves, not their contents
```
nfischer: I forgot to add docs for the `-d` option, so I'm adding them now.

README.md
```diff
@@ -172,6 +172,9 @@ Available options:
...
++ `-l`: return an array of objects instead of strings, each with properties
+        resembling `ls -l` output
```
nfischer: @gvn what do you think about this? I think this level of detail is good (we don't need to specify each field). I think the wording could be improved, however.

ariporad: @nfischer: Agreed, The wording could be improved.

ariporad:
@nfischer: One small comment, then looks good.

README.md
nfischer: @ariporad I updated this with a link, and added an example

ariporad:
LGTM!

src/ls.js
gvn: Ah, I see. In that case `path` or `relativePath` would be more semantic than `name`, but only a by a little bit, so close enough! :wink:

gvn:
This is working great. Thank you @nfischer !

nfischer:
:+1: My pleasure

PR #326, opened 2016-01-31
fix(cat): make behavior more like unix

nfischer:
`cat()` no longer puts `\n`'s in weird places (causing double newlines).

The behavior of `cat()` previously didn't make much sense, since it would stick newlines between files where they don't belong (which unix's `cat` does not actually do).

Example:

```Bash
$ echo -n 'test1' > file1 # this file doesn't end in a newline

$ echo -n 'test2' > file2 # neither does this one

$ cat file1 file2
test1test2
$ # this prompt has no empty line above it, which means that the previous output didn't end in a newline
```

```Javascript
> // Without this fix, ShellJS would behave as follows...
> echo('test1').to('file1'); // no trailing newline
> echo('test2').to('file2'); // no trailing newline
> var s = cat('file1', 'file2');
> assert.equal(s, 'test1\ntest2'); // the newline in the middle shouldn't be there
```


src/cat.js
```diff
@@ -32,12 +32,9 @@ function _cat(options, files) {
     if (!fs.existsSync(file))
       common.error('no such file or directory: ' + file);
 
-    cat += fs.readFileSync(file, 'utf8') + '\n';
+    cat += fs.readFileSync(file, 'utf8');
   });
 
-  if (cat[cat.length-1] === '\n')
```
ariporad: Why is this there? Wouldn't it actually be incorrect.

Example:

```
$ printf "foo\n" > foo # printf, because it properly handles `\n`.
$ printf "bar\n" > bar
$ cat foo bar
foo
bar
$ # There was a newline above this prompt.
```

With this line, there would be no newline at the end.

ariporad:
@nfischer: One comment, otherwise looks good!

src/cat.js
nfischer: @ariporad I'm not sure what you intend to be the correct behavior. I tried your example, and I see a newline above the prompt.

The line I removed would strip a trailing newline that the old code incorrectly appended. The code doesn't append newlines anymore, so it doesn't need to strip a trailing newline.

nfischer:
@ariporad could you clarify your comment?

src/cat.js
ariporad: @nfischer: Whoops... I apparently can't read... Never mind. (I thought you added this).

ariporad:
LGTM!

nfischer:
:+1:

PR #325, opened 2016-01-30
Update cp documentation

freitagbr:
Updates cp documentation to mention that files are not overwritten by default (#210).

Issue #210, opened 2015-06-22
`cp` does not overwrite files by default


lvanengelen:
In POSIX `cp` would just overwrite a file by default, without requiring the `-f` option.  This quite surprised me.  Was this done to make the `cp` behaviour a bit saver? In that case, would it be possible to mention this in the docs?

nfischer:
I agree, this is something that probably should be documented. I think it would be even better to add the `-n` option and make `-f` the default (and simply keep it there for legacy reasons).

nfischer:
This is now fixed in ShellJS v0.6. Thanks for reporting this, @lvanengelen :+1:

PR #328, opened 2016-01-31
fix(mv): add -n option, make -f default behavior

nfischer:
This lets `mv()` have `-f` as the default behavior (the flag exists for legacy reasons). I introduce the `-n` flag, which will allow the opposite behavior. This is the default behavior on Bash.

This also introduces a modification to `parseOptions()`, allowing dictionaries to be specified of the following form:

```Javascript
options = common.parseOptions(options, {
  'f': '!no_force', // the boolean opposite of "no_force"
  'n': 'no_force'
});
```

The empty string will still cause `no_force` to default to false, as it normally would. Options prefixed with `!` have the special meaning that the appearance of that commandline flag sets the option to `false` instead of `true`. So passing the `-f` flag will cause `options.no_force` to be `false`.

If someone passes in a flag like `-fn`, this will set `options.no_force = true`, taking whichever is last, which is consistent with most core-utils. `-nf` will likewise set `options.no_force = false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,12 +246,13 @@ Removes files. The wildcard `*` is accepted.
 ### mv([options ,] source_array, dest')
 Available options:
 
-+ `-f`: force
++ `-f`: force (default behavior)
++ `-n`: no-clobber
 
 Examples:
 
 ```javascript
-mv('-f', 'file', 'dir/');
+mv('-n', 'file', 'dir/');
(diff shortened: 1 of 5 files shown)


nfischer:
This is related to #210 (although this is the `mv` equivalent), and the discussion in #325. If #325 is refactored to change default behavior, then this blocks #325.

ariporad:
LGTM!

PR #329, opened 2016-02-01
feat(set): add new set() command

nfischer:
Add new set() command as a wrapper for `config` variables. This takes the `-e`, `-v`, `+e`, and `+v` flags.

This is based off (and replaces) #149, which introduced the `logCmd` option. This also adds `set()` as a nice wrapper for interfacing with the variables.

`config.verbose` will log each command in a unix-y kind of format. For example, `ls('dir/')` logs as `ls dir/` (strips parens, strips surrounding quotes). This is mostly just because it was easier to implement this way.

I left the docs for `config.fatal` (and added docs for `config.verbose`), although we might also consider deprecating those and encouraging the use of the equivalent `set()` commands instead (to trim down on documentation). One issue is that `config.silent` does not have an analogous [`set` option](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html).

This also fixes a formatting error in `touch.js`'s comment block.

### DIFF
--- a/src/set.js
+++ b/src/set.js
@@ -1,0 +1,49 @@
+var common = require('./common');
+
+//@
+//@ ### set(options)
+//@ Available options:
+//@
+//@ + `+/-e`: exit upon error (`config.fatal`)
+//@ + `+/-v`: verbose: show all commands (`config.verbose`)
+//@
+//@ Examples:
+//@
+//@ ```javascript
+//@ set('-e'); // exit upon first error
+//@ set('+e'); // this undoes a "set('-e')"
+//@ ```
+//@
+//@ Sets global configuration variables
+function _set(options) {
+  if (!options) {
+    var args = [].slice.call(arguments, 0);
+    if (args.length < 2)
+      common.error('must provide an argument');
+    options = args[1];
+  }
+  var negate = (options[0] === '+');
+  if (negate) {
+    options = '-' + options.slice(1); // parseOptions needs a '-' prefix
+  }
+  options = common.parseOptions(options, {
+    'e': 'fatal',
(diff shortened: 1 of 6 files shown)


src/set.js
```diff
@@ -0,0 +1,49 @@
...
+//@ Sets global configuration variables
+function _set(options) {
+  if (!options) {
+    var args = [].slice.call(arguments, 0);
+    if (args.length < 2)
+      common.error('must provide an argument');
+    options = args[1];
+  }
+  var negate = /^\+/.test(options);
+  if (negate) {
```
ariporad: Why not just always prepend '-' here, and register th

nfischer: Yeah, I can edit it to do that. I still need to get the `negate` boolean anyway.

ariporad: @nfischer: No, I realized that what I meant wouldn't work. Never mind.

src/set.js
```diff
@@ -0,0 +1,49 @@
...
+//@
+//@ Sets global configuration variables
+function _set(options) {
+  if (!options) {
+    var args = [].slice.call(arguments, 0);
+    if (args.length < 2)
+      common.error('must provide an argument');
+    options = args[1];
+  }
+  var negate = /^\+/.test(options);
```
ariporad: Small perf nitpick: This should just be `options[0] === '+'` (Way more performant).

nfischer: Good catch! Thanks, I'll update

nfischer:
@ariporad this should be ready

ariporad:
Oops, LGTM!

nfischer:
I'll merge once CI passes

PR #330, opened 2016-02-01
Use bash by default for exec(), if it's available.

### LINKED ISSUE #281: Switch `exec()` to use bash by default
This should be possible via the `shell` option (see [the docs](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)). I've written out some preliminary code for Linux, and it seems to work as expected. I'll fix Windows compatibility and send out a PR.

Edit: `bash` should be the new default on Linux, `cmd.exe` should remain the default on Windows.

nfischer:
This fixes #281. This uses `which()` to check for the existence of an executable named `bash`. If it exists, it adds this to the `shell` option of `exec()` and `execSync()`, otherwise it leaves this blank. This should work for Windows users that have bash installed. The previous behavior was to use `sh` by default on unix, and `cmd.exe` on Windows.

To test this, you can execute:

```Javascript
exec('echo $0'); // outputs the path of the shell, reliably
```

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -1,5 +1,6 @@
 var common = require('./common');
 var _tempDir = require('./tempdir');
+var _which = require('./which');
 var _pwd = require('./pwd');
 var path = require('path');
 var fs = require('fs');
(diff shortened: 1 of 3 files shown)


nfischer:
I'll be using appveyor to debug the issue for Windows. I'll ping this thread and squash out commits once this is ready for review.

nfischer:
@ariporad This should be ready for review now. It basically tries a "best effort" attempt at setting Bash to be the default shell on unix systems (I think bash is not allowed as a shell for `exec()` on windows systems). This exposes the `shell` option to users, so if they know their system can handle something other than `bash` or `cmd.exe`, they can specify that.

src/exec.js
```diff
@@ -64,13 +65,21 @@ function execSync(cmd, opts) {
     maxBuffer: 20*1024*1024
   };
 
+  if (opts.shell)
+    execOptions.shell = opts.shell; // use bash if it exists
+
   var script;
 
+  var execOptString = '{env: process.env, maxBuffer: ' +
```
ariporad: I feel like this should just be an object that we `JSON.stringify`.

test/exec.js
```diff
@@ -79,6 +80,36 @@ assert.equal(shell.error(), null);
 assert.equal(result.code, 0);
 assert.equal(result.stdout, "'+'_'+'\n");
 
+// bash by default on unix
+if (common.platform !== 'win') {
+  var result = shell.exec('echo $0');
+  assert.equal(shell.error(), null);
+  if (shell.which('bash')) {
+    // node v0.10 and earlier don't seem to support the 'shell' option
```
ariporad: I feel like this is a bad idea, if we can't support bash everywhere, we shouldn't support it anywhere.

nfischer: I think supporting bash by default would be nice, but I understand the concern. With that said, if #348 is merged, I think that'll be good enough for most users. Once that's merged, I'm fine with closing this.

nfischer:
Closing this in favor of #348, which allows `bash` (but not by default).

PR #333, opened 2016-02-03
add ability to override execAsync's maxBuffer default of 20MiB

pieterjandesmedt:
- add ability to override execAsync's maxBuffer default of 20MiB
- add maxBuffer override to execSync as well
- added constant DEFAULT_MAXBUFFER_SIZE and made sure maxBuffer is an integer
- fixed indentation and removed delimiter in example

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -5,6 +5,8 @@ var path = require('path');
 var fs = require('fs');
 var child = require('child_process');
 
+var DEFAULT_MAXBUFFER_SIZE = 20*1024*1024;
+
 // Hack to run child_process.exec() synchronously (sync avoids callback hell)
 // Uses a custom wait loop that checks for a flag file, created when the child process is done.
 // (Can't do a wait loop that checks for internal Node variables/messages as
@@ -20,6 +22,9 @@ function execSync(cmd, opts) {
   var options = common.extend({
     silent: common.config.silent
   }, opts);
+
+  var maxBuffer = parseInt(opts.maxBuffer, 10) || DEFAULT_MAXBUFFER_SIZE;
+  if (isNaN(maxBuffer)) maxBuffer = DEFAULT_MAXBUFFER_SIZE;
 
   var previousStdoutContent = '';
   // Echoes stdout changes from running process, if not silent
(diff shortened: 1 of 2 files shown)


src/exec.js
```diff
@@ -21,6 +23,9 @@ function execSync(cmd, opts) {
     silent: common.config.silent
   }, opts);
 
+  var maxBuffer = parseInt(opts.maxBuffer, 10) || DEFAULT_MAXBUFFER_SIZE;
```
ariporad: I don't think we need this line.

src/exec.js
```diff
@@ -137,7 +143,10 @@ function execAsync(cmd, opts, callback) {
     silent: common.config.silent
   }, opts);
 
-  var c = child.exec(cmd, {env: process.env, maxBuffer: 20*1024*1024}, function(err) {
+  var maxBuffer = parseInt(opts.maxBuffer, 10) || DEFAULT_MAXBUFFER_SIZE;
```
ariporad: Why is this twice?

ariporad:
@pieterjandesmedt:Thanks so much for the PR, I have a few small comments, otherwise it looks great!

@nfischer?

nfischer:
This does not resolve any merge conflicts, so CI won't run.

@pieterjandesmedt would you like me to copy your diff to a new PR and resolve the conflicts for you? Also, for future reference, you don't have to make fresh PRs when we ask you to "rebase," you just need to use "git push -f" (new PRs can be confusing)

src/exec.js
pieterjandesmedt: I don't see it... why not? If someone decides to input `{ maxBuffer: "20000" }`, this line will make that work. Or am I missing something?

nfischer: @pieterjandesmedt I think he meant that `"hello"` is invalid input, so we don't need to support it. I'm fine with either catching the mistake and defaulting to a value (like you've done), or with crashing. I actually prefer your method.

ariporad: @pieterjandesmedt: Yeah, I meant that I don't think we need our API to support someone passing in an invalid value. 

But now that I think about it, this line will end up being useful for #244, so forget I ever said anything.

ariporad:
@pieterjandesmedt: As much as I hate to reject a PR, it looks like this has been fixed by #335, so I have to close this. Sorry.

Issue #132, opened 2014-05-21
add timeout option for exec


dcharbonnier:
http://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

manovotny:
Gosh I want this feature so bad right now. :frowning: 

Might have to roll up my sleeves and make a pull request!

HQidea:
+1
and maybe it can also pass other options to child_process.exec[Sync]

lizzie:
+1

nfischer:
:+1: I think this would be a good feature to add. I see a trend that most people want the native options to be exposed, so I suspect we'll modify `exec()` to support all native options. I'll tag this issue in a PR over the next couple weeks.

ariporad:
Ok, this in fixed in #335, so I'm going to close this.

westlakem:
Ok, but how do I specify the timeout?

nfischer:
@westlakem 

https://github.com/shelljs/shelljs/blob/d7b6a1f3784fa6f9d1abcc07fcec6d08e89bf294/test/exec.js#L130

PR #335, opened 2016-02-04
feat(exec): allow all exec options to pass through

### LINKED ISSUE #250: why not support set 'cwd'  when invoke execAsync ?
https://github.com/shelljs/shelljs/blob/v0.5.3/src/exec.js#L140

nfischer:
This allows any option from https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback to be passed through our `shell.exec()` function to the native `exec()`/`execSync()`.

This fixes #250, #132 and replaces #270 #284 #333 #163 

This specifically tests the `cwd`, `maxBuffer`, and `timeout` options, since those were requested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,6 +486,8 @@ Available options (all `false` by default):
 + `async`: Asynchronous execution. If a callback is provided, it will be set to
   `true`, regardless of the passed value.
 + `silent`: Do not echo program output to console.
++ and any option available to NodeJS's
+  [child_process.exec()](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
 
 Examples:
 
--- a/src/exec.js
+++ b/src/exec.js
@@ -5,6 +5,8 @@ var path = require('path');
 var fs = require('fs');
 var child = require('child_process');
 
+var DEFAULT_MAXBUFFER_SIZE = 20*1024*1024;
+
 // Hack to run child_process.exec() synchronously (sync avoids callback hell)
 // Uses a custom wait loop that checks for a flag file, created when the child process is done.
 // (Can't do a wait loop that checks for internal Node variables/messages as
@@ -18,7 +20,7 @@ function execSync(cmd, opts) {
       scriptFile = path.resolve(tempDir+'/'+common.randomFileName()),
       sleepFile = path.resolve(tempDir+'/'+common.randomFileName());
 
-  var options = common.extend({
+  opts = common.extend({
     silent: common.config.silent
   }, opts);
 
@@ -26,7 +28,7 @@ function execSync(cmd, opts) {
       previousStderrContent = '';
   // Echoes stdout and stderr changes from running process, if not silent
   function updateStream(streamFile) {
-    if (options.silent || !fs.existsSync(streamFile))
+    if (opts.silent || !fs.existsSync(streamFile))
       return;
 
     var previousStreamContent,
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -486,6 +486,8 @@ Available options (all `false` by default):
 + `async`: Asynchronous execution. If a callback is provided, it will be set to
   `true`, regardless of the passed value.
 + `silent`: Do not echo program output to console.
++ and any option available to NodeJS's
+  [child_process.exec()](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
```
ariporad: Why is this on a seperate line?

src/exec.js
```diff
@@ -58,19 +60,16 @@ function execSync(cmd, opts) {
...
   var execCommand = '"'+process.execPath+'" '+scriptFile;
-  var execOptions = {
-    env: process.env,
-    cwd: _pwd(),
-    maxBuffer: 20*1024*1024
-  };
-
+  opts.cwd = opts.cwd || _pwd();
+  opts.env = opts.env || process.env;
+  opts.maxBuffer = opts.maxBuffer || DEFAULT_MAXBUFFER_SIZE;
```
ariporad: Why aren't these up on line 23 in the `common.extend`?

src/exec.js
```diff
@@ -160,7 +159,11 @@ function execAsync(cmd, opts, callback) {
     silent: common.config.silent
   }, opts);
 
-  var c = child.exec(cmd, {env: process.env, maxBuffer: 20*1024*1024}, function(err) {
+  opts.env = opts.env || process.env;
+  opts.cwd = opts.cwd || _pwd();
+  opts.maxBuffer = opts.maxBuffer || DEFAULT_MAXBUFFER_SIZE;
```
ariporad: Same comment as above.

ariporad:
@nfischer: A few small comments, otherwise looks good!

src/exec.js
nfischer: `common.extend()` prefers new values over old values. This block prefers old values over new values. What I mean is...

If someone supplies a `cwd` option to `exec()`, it will not be overridden by `_pwd()`. If someone does not provide it, then it defaults to `_pwd()`. Same logic for the other options (a default maxbuffer value was provided before, so I kept that as well). `process.env` is good because it makes sure any `env` changes get passed through to the exec call.

README.md
nfischer: @ariporad I try to format lines to be under 80 characters. This is how vim's built-in formatting did it. Markdown allows you to extend lines (as long as you don't leave one blank line in-between) without it affecting the end-result. You can check "view file" to be sure

src/exec.js
ariporad: @nfischer: But you could do this, and it would have the same effect (I think):

```javascript
opts = common.extend({
  silent: common.config.silent,
  cwd: _pwd(),
  env: process.env,
  maxBuffer: DEFAULT_MAXBUFFER_SIZE,
}, opts);
```

nfischer: @ariporad fixed. If it passes CI, feel free to squash the commits and merge.

README.md
ariporad: SGTM

nfischer:
@ariporad this should be good now

ariporad:
LGTM!

PR #336, opened 2016-02-04
fix(cp): add -n option, make -f default behavior

nfischer:
Fix default behavior for `cp()` for bash compatibility. This replaces #325, and is done in a similar style to #328


ariporad:
LGTM!

PR #339, opened 2016-02-04
chore(.npmignore): update npmignore

ariporad:


ariporad:
@nfischer: Ready for review.

.npmignore
```diff
@@ -1,2 +1,13 @@
 test/
-tmp/
\ No newline at end of file
+tmp/
+scripts/
```
nfischer: I think we can keep this

.npmignore
```diff
@@ -1,2 +1,13 @@
...
+.documentup.json
+.gitignore
+.jshintrc
+.lgtm
+.travis.yml
+appveyor.yml
+RELEASE.md
+
+# This is commented out because I don't see the point of not including it.
+# MAINTAINERS
```
nfischer: I think we can keep the MAINTAINERS as well. We can probably delete these comments

.npmignore
ariporad: @nfischer: I suppose we could, but I can't think of a reason anyone who downloads it from npm would ever need them, so I thought we may as well save them the bytes.

.npmignore
ariporad: Will do.

.npmignore
nfischer: I guess the only reason is so that they can run tests to verify their build works. It might be helpful for debugging someone's problems

ariporad: Hmm... Ok.

ariporad:
@nfischer: should be fixed now.

nfischer:
LGTM

PR #341, opened 2016-02-04
fix(touch): enhance parseOptions and fix touch's -r flag

nfischer:
This improves `parseOptions` to allow options of the form:

```
touch({'-r', 'path/to/file'}, 'other-file.txt');
```

to supply a value to an option. In this example, `-r` has the string value `'path/to/file'`.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -67,46 +67,57 @@ function getUserHome() {
 }
 exports.getUserHome = getUserHome;
 
-// Returns {'alice': true, 'bob': false} when passed a dictionary, e.g.:
+// Returns {'alice': true, 'bob': false} when passed a string and dictionary as follows:
 //   parseOptions('-a', {'a':'alice', 'b':'bob'});
-function parseOptions(str, map) {
+// Returns {'reference': 'string-value', 'bob': false} when passed two dictionaries of the form:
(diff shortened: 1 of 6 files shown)


src/common.js
```diff
@@ -80,33 +82,44 @@ function parseOptions(str, map) {
...
-      else
-        options[opt] = true;
-    } else {
-      error('option not recognized: '+c);
-    }
-  });
+  var optionName;
+  if (typeof opt === 'string') {
+    // e.g. match[1] = 'Rf' for opt = '-Rf'
+    var match = opt.match(/^\-(.+)/);
```
ariporad: This would be much more performant if it just was `opt[0] === '-'`.

ariporad:
One small nit, otherwise looks good!

ariporad:
@nfischer: In the changelog, how should this be described?

nfischer:
@ariporad I replaced the instances of `match()` with suitable alternatives

ariporad:
@nfischer: Fantastic!

nfischer:
In the changelog, you can just say something like "feat: now supports valued options (see `touch -r` for an example)."

ariporad:
Will Do!

ariporad:
LGTM!

PR #346, opened 2016-02-09
feat(touch): supports multiple files

nfischer:
Does what it says. From the (updated) docs:

> touch([options,] file [, file ...])
> touch([options,] file_array)

This is consistent with the API for `ls()`, `cp()`, `mv()`, `rm()`, etc.

### DIFF
--- a/src/touch.js
+++ b/src/touch.js
@@ -2,7 +2,8 @@ var common = require('./common');
 var fs = require('fs');
 
 //@
-//@ ### touch([options,] file)
+//@ ### touch([options,] file [, file ...])
+//@ ### touch([options,] file_array)
 //@ Available options:
 //@
 //@ + `-a`: Change only the access time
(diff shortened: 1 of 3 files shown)


src/touch.js
```diff
@@ -31,20 +32,18 @@ function _touch(opts, files) {
...
-  if (Array.isArray(files)) {
-    files.forEach(function(f) {
-      touchFile(opts, f);
-    });
-  } else if (typeof files === 'string') {
-    touchFile(opts, files);
-  } else {
+  if (!files)
+    common.error('no files given');
+  else if (Array.isArray(files))
```
ariporad: Why is this here?

nfischer: This is to do a no-op. For what it's worth, I think #352 will make this line obsolete, so we might want to wait until that's merged

ariporad: @nfischer: Ok, #352 is merged.

nfischer: I'll rebase and update this tonight or tomorrow. Thanks!

nfischer:
@ariporad this should be fixed now

ariporad:
LGTM! Will merge once Travis is done.

ariporad:
LGTM!

Issue #208, opened 2015-05-20
exec failed to return


qiluo:
I'm trying to call

```
exec('cmd <<< "dir"')
```

the command failed to return, don't know if it's related to `cmd <<<`

nfischer:
I tried running the following commands, and saw this output:

```
$ node
> require('shelljs/global');
{}
> exec('cat <<< "dir"');
/bin/sh: 1: Syntax error: redirection unexpected
{ code: 2,
  output: '/bin/sh: 1: Syntax error: redirection unexpected\n' }
```

This suggests that `exec()` uses `sh`, not `bash`. When I try emulating this in `sh`, I see comparable output:

```sh
$ sh # make sure I'm using sh instead of bash
$ cat <<< "dir"
sh: 1: Syntax error: redirection unexpected
```

So this sounds like you're using syntax that isn't supported by `sh`, and therefore won't be supported by `exec()`.

It might be worth it to modify `exec()` to call bash instead of sh, but this would change the project's behavior, so it should only be done with some hesitation.

ariporad:
@nfischer: I think that we should switch to bash, it would likely be more intuitive.

nfischer:
@ariporad I agree, I think bash would be better. I think it's ok for now to put this off, since we're at least matching `sh`, but we may want to move forward and switch to bash in the near future.

nfischer:
I opened #281 to discuss switching shells

nfischer:
@qiluo this should be fixed by #335, which is in v0.6. See the `shell` option for Node's [child_process.exec](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback).

nfischer:
Whoops! Looks like #335 doesn't quite fix this. I'm opening #348 to actually resolve the issue like I suggested (and it adds a test to make sure).

nfischer:
#348 is now merged, so this should fix the issue. You should be able to use `exec('echo foo', {shell: '/bin/bash'})` to do what you need (assuming that's your path to bash, of course).

PR #348, opened 2016-02-09
fix(exec): now actually supports shell option

### LINKED ISSUE #138: shell builtin
stupid question: can I call shell-builtin ?
Im trying exec with history
getting:
'history' is not recognized as an internal or external command,
operable program or batch file.
Exit code: 1
Program output: 'history' is not recognized as an internal or external command,
operable program or batch file.

nfischer:
Looks like #335 doesn't work perfectly with all options. This is because `execSync` does 2 calls to `exec()`, and some options need to be passed to both calls.

This should clean things up a bit, and fix the issue for the `shell` option, which would actually fix #138 and #208.

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -65,11 +65,14 @@ function execSync(cmd, opts) {
   var execCommand = '"'+process.execPath+'" '+scriptFile;
   var script;
 
+  opts.cwd = path.resolve(opts.cwd);
+  var optString = JSON.stringify(opts);
+
   if (typeof child.execSync === 'function') {
     script = [
       "var child = require('child_process')",
       "  , fs = require('fs');",
-      "var childProcess = child.exec('"+escape(cmd)+"', {env: process.env, maxBuffer: "+opts.maxBuffer+"}, function(err) {",
+      "var childProcess = child.exec('"+escape(cmd)+"', "+optString+", function(err) {",
       "  fs.writeFileSync('"+escape(codeFile)+"', err ? err.code.toString() : '0');",
       "});",
       "var stdoutStream = fs.createWriteStream('"+escape(stdoutFile)+"');",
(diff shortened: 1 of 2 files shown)


src/exec.js
```diff
@@ -65,11 +65,18 @@ function execSync(cmd, opts) {
   var execCommand = '"'+process.execPath+'" '+scriptFile;
   var script;
 
+  var optArray = [];
```
ariporad: I feel like this should just be an object that we `JSON.stringify`.

nfischer: That would probably be better. I'll look into that.

nfischer:
@ariporad This should be good now

test/exec.js
```diff
@@ -107,6 +107,31 @@ if (process.version >= 'v0.11') { // this option doesn't work on v0.10
   assert.ok(shell.error());
 }
 
+// check process.env works
+assert.ok(!shell.env.FOO);
+shell.env.FOO = 'Hello world';
+result = shell.exec('echo $FOO');
```
ariporad: `cmd.exe` doesn't support variables prefixed with a `$`, what you want here is `%FOO%`.

ariporad:
LGTM, except for AppVeyor, which is failing.

test/exec.js
nfischer: Thanks! Good catch

nfischer:
@ariporad it should be working now

src/exec.js
```diff
@@ -65,11 +65,14 @@ function execSync(cmd, opts) {
   var execCommand = '"'+process.execPath+'" '+scriptFile;
   var script;
 
+  opts.cwd = path.resolve(opts.cwd);
+  var optString = JSON.stringify(opts);
```
TimothyGu: This isn't safe against Unicode newlines (namely, U+2028 and U+2029): http://stackoverflow.com/a/9168133/1937836

Something like https://www.npmjs.com/package/js-stringify should fix this issue

PR #349, opened 2016-02-09
feat(grep): add -l option

nfischer:
`grep -l` will only print filenames of matching files.

Sort-of replaces #116. I can add `-s` support if we want to completely close that PR.


ariporad:
LGTM!

PR #352, opened 2016-02-11
feat(glob): glob support for (almost) all commands

### LINKED ISSUE #343: Glob commands by default
It'd be nice if any command could support globs by default. For example:

```Javascript
touch('a.txt');
touch('b.txt');
touch('*.txt'); // this doesn't create a file, it touches the two other files
```

This has the concern that calling `glob.sync()` might have a performance hit. However, this would be very consistent with Bash's behavior, which is a win. We might want to limit this to only *some* commands. For example:

```Javascript
touch('pwd');
which('p*'); // This should technically output the path of `pwd`, but this is not very common
// this would be an example that we might not want to support, for performance reasons
```

I think this should apply to the majority of commands, however.

nfischer:
Fixes #343. This performs glob expansion on commands, where it makes sense. This doesn't glob for:

 - `pwd()` (it doesn't take any arguments)
 - `test()` (globbing here is very unreliable in bash)
 - `exec()` (this actually **does** glob-expand, but that's done in the shell itself (`/bin/sh`), not by this function
 - `set()` (globbing doesn't make sense)
 - `tempdir()` (it doesn't take any arguments)

Some other commands we might want to not permit to glob (because it isn't useful, and I suspect globbing slightly hampers performance) would be:

 - `which()`
 - `mkdir()`

I had to modify some test cases that didn't actually have correct behavior (the `ls('path/*')` cases were often wrong). I also added a lot of new test cases for globs on commands that didn't previously have them.

I think this should be merged before #275, because it will help make sure that `glob.sync` doesn't break behavior.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -127,10 +127,17 @@ exports.parseOptions = parseOptions;
 //   expand(['file*.js']) = ['file1.js', 'file2.js', ...]
 //   (if the files 'file1.js', 'file2.js', etc, exist in the current dir)
 function expand(list) {
+  if (!Array.isArray(list)) {
+    throw new TypeError('must be an array');
+  }
   var expanded = [];
   list.forEach(function(listEl) {
+    // Don't expand non-strings
+    if (typeof listEl !== 'string') {
+      expanded.push(listEl);
+    }
     // Wildcard present on directory names ?
-    if(listEl.search(/\*[^\/]*\//) > -1 || listEl.search(/\*\*[^\/]*\//) > -1) {
+    else if(listEl.search(/\*[^\/]*\//) > -1 || listEl.search(/\*\*[^\/]*\//) > -1) {
       var match = listEl.match(/^([^*]+\/|)(.*)/);
       var root = match[1];
       var rest = match[2];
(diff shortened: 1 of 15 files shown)


src/common.js
```diff
@@ -206,6 +218,34 @@ exports.extend = extend;
 
 // Common wrapper for all Unix-like commands
 function wrap(cmd, fn, options) {
+  // -1 means "don't glob"
+  // 0 means "glob the options" (probably don't want this)
+  var indexOfExpandableArg = {
```
ariporad: This should be passed in as an option to common.wrap.

ariporad:
One nitpick, otherwise is good!

src/common.js
nfischer: Yeah, I was thinking about that approach as well. It's probably cleaner, so I'll modify it to do that instead. I'll get rid of the `-1` stuff too.

nfischer:
@ariporad This should have your comments resolved. What do you think about globbing for `which()` and `mkdir()`? If we want to support these, then we should probably make unit tests for them. Otherwise, I can remove globbing support for them.

Also, I'll probably add something about "globbing is supported for all commands"

ariporad:
@nfischer: Yes for `mkdir`, no for `which`.

nfischer:
@ariporad `which()` does not glob, and `mkdir()` does glob. This should be ready for review

ariporad:
LGTM!

PR #354, opened 2016-02-11
fix(exec): temp files are now cleaned up

### LINKED ISSUE #353: exec() doesn't clean up all temp files
I've noticed on my system I get lots of temp files in `/tmp/` (I'm on Ubuntu 15.10). I've tested (I just run `npm test`, which I know makes at least one call to `exec()` per test) under the following versions of node:

 - v0.10: everything gets cleaned up
 - v0.11, v0.12, v4, v5: `npm test` leaves 1 code file, 2 script files, 2 stderr files, and 2 stdout files

v0.11 is when `execSync()` got introduced, so this could be related.

nfischer:
Fixes #353 

The issue had to do with an exception being thrown, causing `execSync()` to skip the cleanup. This seems to fix the issue for each version of node we support

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -94,7 +94,16 @@ function execSync(cmd, opts) {
     }
 
     // Welcome to the future
-    child.execSync(execCommand, opts);
+    try {
+      child.execSync(execCommand, opts);
+    } catch (e) {
+      // Clean up immediately if we have an exception
+      try { common.unlinkSync(scriptFile); } catch(e) {}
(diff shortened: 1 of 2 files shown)


nfischer:
This also fixes an issue where `shell.error()` would sometimes indicate success, even if the process exited with non-zero status (like 12, for example).

nfischer:
@ariporad this should be ready for review

ariporad:
LGTM!

PR #355, opened 2016-02-14
Consistently use LF line endings

TimothyGu:
Fixes cat, ls, and sed tests on Windows when Git line ending configuration is set to auto.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf


nfischer:
@TimothyGu could you please also add this file name to npmignore? Other than that, LGTM

nfischer:
Tested this on my system, and it seems harmless, so I'll merge. Thanks, @TimothyGu!

PR #357, opened 2016-02-15
config.fatal now throws an exception

jrmclaurin:
Hi everyone!

This PR adds a 'config.throw' option to throw errors from shell commands. Config.fatal is useful, but it doesn't log stack traces, which makes deep errors hard to find. This 'throw' option is useful in synchronous scripts where you have try / catch handling, or you want the default behavior of printing a full stack trace and exiting.

Thanks!!

### DIFF
--- a/test/exec.js
+++ b/test/exec.js
@@ -19,20 +19,14 @@ assert.ok(result.code > 0);
 
 // Test 'fatal' mode for exec, temporarily overriding process.exit
 var old_fatal = shell.config.fatal;
-var old_exit = process.exit;
-
-var exitcode = 9999;
-process.exit = function (_exitcode) {
-    exitcode = _exitcode;
-};
 
 shell.config.fatal = true;
 
-var result = shell.exec('asdfasdf'); // could not find command
-assert.equal(exitcode, 1);
+assert.throws(function() {
+  shell.exec('asdfasdf'); // could not find command
+}, /exec: internal error/);
 
 shell.config.fatal = old_fatal;
-process.exit = old_exit;
 
 //
 // Valids
--- a/test/set.js
+++ b/test/set.js
@@ -25,9 +25,11 @@ assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n'
 
 // set -e
 var result = shell.exec('node -e \"require(\'../global\'); set(\'-e\'); ls(\'file_doesnt_exist\'); echo(1234);\"');
-assert.equal(result.code, 1);
+var nodeVersion = process.versions.node.split('.').map(function(str) { return parseInt(str, 10); });
+var uncaughtErrorExitCode = (nodeVersion[0] === 0 && nodeVersion[1] < 11) ? 8 : 1;
+assert.equal(result.code, uncaughtErrorExitCode);
 assert.equal(result.stdout, '');
-assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n');
+assert(result.stderr.indexOf('Error: ls: no such file or directory: file_doesnt_exist') >= 0);
 
 // set -v
 var result = shell.exec('node -e \"require(\'../global\'); set(\'-v\'); ls(\'file_doesnt_exist\'); echo(1234);\"');
@@ -37,9 +39,9 @@ assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n'
 
 // set -ev
 var result = shell.exec('node -e \"require(\'../global\'); set(\'-ev\'); ls(\'file_doesnt_exist\'); echo(1234);\"');
-assert.equal(result.code, 1);
+assert.equal(result.code, uncaughtErrorExitCode);
 assert.equal(result.stdout, 'ls file_doesnt_exist\n');
-assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n');
+assert(result.stderr.indexOf('Error: ls: no such file or directory: file_doesnt_exist') >= 0);
 
 // set -e, set +e
 var result = shell.exec('node -e \"require(\'../global\'); set(\'-e\'); set(\'+e\'); ls(\'file_doesnt_exist\'); echo(1234);\"');
--- a/README.md
+++ b/README.md
@@ -635,12 +635,13 @@ Example:
 ```javascript
 require('shelljs/global');
 config.fatal = true; // or set('-e');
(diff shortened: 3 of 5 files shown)


nfischer:
@jrmclaurin did you make your README changes in the file that the README is generated from? I think it's shell.js

@ariporad what do you think of this? I don't like having 2 options for basically the same thing. Maybe config.throw is better behavior? So we could change config.error to throw an exception, and let config.error & config.silent to silently exit with an error code. Just a thought, not sure if this would be an issue

jrmclaurin:
Hi @nfischer, I hadn't noticed that README was generated, so I updated it directly. Sorry about that.

nfischer:
Could you move the changes to the file its generated from? Otherwise CI won't pass. The issue is that regenerating docs will just wipe out your changes, which isn't what we want

ariporad:
I say we make config.fatal just do this. I can't see any reason why that 
would
be bad.

nfischer:
@jrmclaurin would you be able to refactor config.fatal?

jrmclaurin:
@nfischer Sure thing, I'll do that refactor, and put the docs in the right place. Thanks!!

nfischer:
Awesome! Thanks, @jrmclaurin! If you can, please squash your commits when finished (or we can do that for you, if you prefer). This should be a good addition

jrmclaurin:
OK @nfischer, I refactored to change the behavior of config.fatal, and update unit tests to match. I believe I squashed the commits correctly, but let me know if I did anything wrong. Thanks everyone for your open-mindedness about this issue!

jrmclaurin:
I forgot about @nfischer's suggestion to exit silently if both config.fatal and config.silent are set. This is a bit of conjecture, but if someone sets both config.silent and config.fatal, they probably want config.silent to eliminate console noise when everything's OK, but they'll expect config.fatal to still throw when an error happens. So should we leave it like it is (fatal always throws), and anyone who wants "silent fatal" behavior can wrap their code in try / catch and swallow errors? Making config.fatal stronger than config.silent seems like the safer choice.

nfischer:
As it is, the only time commands output to the console is for errors (and for `echo()` statements). `config.silent` actually only suppresses error messages, not `echo()` commands (although I think this should be changed, but that's just my opinion). So that's why I was thinking it should be like `config.fatal`'s current behavior, since `config.silent` functions to suppress error messages.

That's just a thought though, and I don't mind much one way or another, since I don't think backward compatibility in this sense is terribly important. The try-catch approach is also good for suppressing errors, and having one approach probably makes things a little cleaner.

test/set.js
```diff
@@ -25,9 +25,10 @@ assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n'
 
 // set -e
 var result = shell.exec('node -e \"require(\'../global\'); set(\'-e\'); ls(\'file_doesnt_exist\'); echo(1234);\"');
-assert.equal(result.code, 1);
+var uncaughtErrorExitCode = process.version.indexOf('v0.10') === 0 ? 8 : 1;
+assert.equal(result.code, uncaughtErrorExitCode);
 assert.equal(result.stdout, '');
-assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n');
+assert(result.stderr.indexOf('Error: ls: no such file or directory: file_doesnt_exist') >= 0);
```
nfischer: Why does this have "Error: " prepended to the front?

test/set.js
```diff
@@ -25,9 +25,10 @@ assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n'
 
 // set -e
 var result = shell.exec('node -e \"require(\'../global\'); set(\'-e\'); ls(\'file_doesnt_exist\'); echo(1234);\"');
-assert.equal(result.code, 1);
+var uncaughtErrorExitCode = process.version.indexOf('v0.10') === 0 ? 8 : 1;
```
nfischer: Is this only for version v0.10, or is this for all versions earlier than v0.11? If it's the latter, a `< 'v0.11'` might be better

test/set.js
jrmclaurin: Instead of directly printing the message to stderr, we're now throwing an Error, so we end up at node's uncaught error handling. That's why you end up with "Error:" in front of the message.

nfischer: Ah, makes sense. Sounds good

test/set.js
jrmclaurin: You're right. Not sure exactly when that behavior changed. I'll try to figure that out and make the test work everywhere.

jrmclaurin:
OK, I made that node.js version check in test/set.js actually work right :-)

test/exec.js
```diff
@@ -19,20 +19,16 @@ assert.ok(result.code > 0);
...
-};
 
 shell.config.fatal = true;
 
-var result = shell.exec('asdfasdf'); // could not find command
-assert.equal(exitcode, 1);
+assert.throws(function() {
+  shell.exec('asdfasdf'); // could not find command
+}, function(err) {
+  return err.message === 'exec: internal error';
```
ariporad: Two things:
1. I'm pretty sure you can just pass in the expected error message here, and it will compare.
2. Is there any way to make the error message a little more descriptive?

ariporad:
Looks good,but I have one small nitpick.

jrmclaurin:
@ariporad You were right that assert.throws can just look at the message, so I made that change. Do you want a more detailed message when I'm creating the Error? I can't think of anything else I could put in there that would be helpful.

ariporad:
@jrmclaurin: I was thinking something like "Command not found"

jrmclaurin:
@ariporad Good point--I hadn't realized that the error message from the command isn't making it into the thrown error. To fix that problem, it looks like src/exec.js would need to scrape some stderr and / or stdout, and stuff it in the message that's being sent to common.error. Here's the scene of the crime, src/exec.js line 151:

if (code !== 0)  {
    common.error('', true);
  }

ariporad:
@jrmclaurin: Ok. I'm working on re-doing exec, so don't worry.

LGTM!

ariporad:
@nfischer: feel free to merge.

nfischer:
I'll test this out today and merge if I have no issues

nfischer:
LGTM! I renamed the PR to better describe the change that's being merged, since it's deviated a bit from the original intent. Thanks, @jrmclaurin!

PR #358, opened 2016-02-16
feat(set): add -f option to disable globbing

### LINKED ISSUE #344: Add -f option to set()
The docs for unix `set` can be found [here](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)

`set -f` prevents things from globbing. This would be a useful addition if #343 is resolved (and doesn't really make sense to resolve before then), since it could be argued this makes code much safer (at the cost of convenience).

```Javascript
set('-f');
touch('*.txt'); // this creates a literal file named *.txt
```

nfischer:
See the docs for `set('-f')`. This is analogous to Bash's functionality.

It's a nice option to have safer code for the bulk of the utilities (this doesn't affect `exec()` *yet* but this will disable globbing for that too once that's an option)

Fixes #344

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -7,6 +7,7 @@ var config = {
   silent: false,
   fatal: false,
   verbose: false,
+  noglob: false
 };
 exports.config = config;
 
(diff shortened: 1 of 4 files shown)


nfischer:
@ariporad this should be ready for review now

src/common.js
```diff
@@ -7,6 +7,7 @@ var config = {
   silent: false,
   fatal: false,
   verbose: false,
+  noglob: false
```
ariporad: Suggestion (Totally optional): Maybe make this `glob`, and have it on by default?

ariporad:
LGTM!

PR #359, opened 2016-02-16
feat(glob): use glob module for globbing

### LINKED ISSUE #225: ls globbing does not behave like shell, consider using glob.sync

nfischer:
Switch to the glob module to do shell globbing. Fixes a bug in `cp()` where
hidden files were not copied recursively.

Replaces #275, fixes #225. This is a less-aggressive version of #275 (doesn't modify `ls()` nearly as much). We should probably add some more glob-related tests though, like `file?.txt` and things of that sort.

I wrote this up because this will fix one of the bugs I found when writing the `ShellString` refactor for free.

For what it's worth, this does hurt performance slightly (it takes 16 seconds on average to run `npm test` instead of 15 seconds), but not by very much. We might be able to get that performance back some other way.

Closes #275, #225.

### DIFF
--- a/src/cp.js
+++ b/src/cp.js
@@ -159,7 +159,7 @@ function _cp(options, sources, dest) {
     }
   }
 
-  sources = common.expand(sources);
+  sources = common.expand(sources, {dot: true});
 
   sources.forEach(function(src) {
     if (!fs.existsSync(src)) {
--- a/package.json
+++ b/package.json
@@ -28,7 +28,9 @@
   "bin": {
     "shjs": "./bin/shjs"
   },
-  "dependencies": {},
+  "dependencies": {
+    "glob": "^7.0.0"
+  },
   "devDependencies": {
     "coffee-script": "^1.10.0",
     "jshint": "~2.1.11"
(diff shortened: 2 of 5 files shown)


src/cp.js
```diff
@@ -159,7 +159,7 @@ function _cp(options, sources, dest) {
     }
   }
 
-  sources = common.expand(sources);
+  sources = common.expand(sources, {dot: true});
```
ariporad: Isn't this done in `common.wrap`?

ariporad:
LGTM.

I'm going to try to simplify `ls` a bit, in a separate PR.

ariporad:
Also, @nfischer, thank so much for doing this.

src/cp.js
nfischer: This is still necessary because this is how we turn a directory into a list of its contents. We can probably do an alternative approach, though

nfischer:
I agree, we probably can and should trim down on ls's code a bit

Issue #159, opened 2014-09-08
Don't modify string prototype


patrickarlt:
Ran into an issue debugging an issue with a coworker.

If the `string` module is included anywhere downstream of `shelljs` you get a pair of warnings.

```bash
to: wrong arguments

toEnd: wrong arguments
```

Here is the script to duplicate the issue.

```js
var shelljs = require('shelljs');
var S = require('string');

// will produce output...
```

This has already been caught by several other issues...

* https://github.com/jprichardson/string.js/issues/73
* https://github.com/paazmaya/grunt-trimtrailingspaces/issues/1
* https://github.com/admc/wd/issues/220

Since shelljs is so heavily relied upon it might be a better to not modify the global `String` prototype.

cookch10:
Also experiencing this issue.  Please fix.

steveoh:
[to](https://github.com/arturadib/shelljs/blob/20db182fcbc96cb8bb4ece2c748b15b755b022ba/shell.js#L58)
and [toEnd](
https://github.com/arturadib/shelljs/blob/20db182fcbc96cb8bb4ece2c748b15b755b022ba/shell.js#L62)

prototype violations. This should be fairly simple to fix?

jprichardson:
I've published a new version `3.1.1` of `string.js` that should work around this problem.

myndzi:
Extending the prototype at all is dubious, especially implicitly instead of explicitly, but doing so with functions that dirty the filesystem is just nasty. I just spent two hours of my day running down where a mysterious file was coming from that's a result of what appears to be nasty behavior on the part of both shelljs AND string.js; either fix will do, but they both need it.

nfriedly:
+1 This looks to be the root cause of https://github.com/watson-developer-cloud/nodejs-wrapper/issues/56 / https://github.com/jprichardson/string.js/issues/160

az7arul:
version `3.3.1` of `string.js` now ignores the `to` and `toEnd` methods to get around this problem

garaboncias:
+1 pleasse do not modify or at least make it not enumerable.

nfischer:
Bumping this thread, since this will finally be resolved in #360. Thanks for reporting this!

nfischer:
> Even an option to disable it would work great. Thanks!

[...]

Issue #154, opened 2014-08-22
`exec(...).to(file)` should work


aurium:
We know `exec(...).output.to(file)` works, but... why to be so verbose?

`node -e "require('shelljs/global'); exec('git status').to('/tmp/file')"` should work.

ariporad:
Hmm... I'm not sure. On one hand it makes sense, but the other it's confusing.

@nfischer: Do you think we should implement this?

nfischer:
@ariporad I was thinking of something similar actually. I was thinking of refactoring everything to return a `ShellString` which acts like a string, but also has a `.code`, `.stdout` (equivalent to default behavior), and `.stderr`, among some other features. If `exec()` gets refactored too, then either style would work.

ariporad:
@nfischer: Hmm... I think it would make more sense if everything returned a ShellResult, because the stdout duplication would be weird.

ariporad:
Alternatively, the ShellString could contain all output (stdout + stderr), then we could have ShellString.stdout/stderr.

nfischer:
This is related to #148

nfischer:
This will be resolved in #360.

Once that is merged, the **only** way to achieve this is with `exec(...).to(file)`, as `exec(...).output.to(file)` and `exec(...).stdout.to(file)` will actually no longer work

nfischer:
Closing this. Please see my comment above if you see a break in behavior

PR #360, opened 2016-02-16
Refactor shellstring

### LINKED ISSUE #356: Standardize command output
We need to standardize the output of the commands. Currently, they return a whole bunch of things. (See the [`shx` output section](http://ariporad.link/1ojYTMg)).

I think that @nfischer is working on a standardized `ShellString` for all commands to return. I'm not sure about the details.

At bear minimum, I think that all commands need to return something with `stdout` and `stderr` properties. This will allow #269 to simply have `error()` return the exit code of the previous command (equivalent to `$?`). It will also greatly simplify `shx`.

nfischer:
This is a PR to refactor commands to return `ShellString` objects instead of regular strings. This still has some broken tests, so it's far from ready, but I think in principle this approach will work.

Feel free to fix any tests and/or post the diffs.

Fixes #356, #154.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -1,6 +1,12 @@
+// jshint -W053
+// Ignore warning about 'new String()'
+'use strict';
+
 var os = require('os');
 var fs = require('fs');
 var glob = require('glob');
+var _to = require('./to');
+var _toEnd = require('./toEnd');
 
 // Module globals
 var config = {
@@ -46,11 +52,26 @@ function error(msg, _continue) {
 }
 exports.error = error;
 
-// In the future, when Proxies are default, we can add methods like `.to()` to primitive strings.
-// For now, this is a dummy function to bookmark places we need such strings
-function ShellString(str) {
-  return str;
-}
+//@
+//@ ### ShellString(str)
+//@
+//@ Examples:
+//@
+//@ ```
+//@ var foo = ShellString('hello world');
+//@ ```
+//@
+//@ Turns a regular string into a string-like object similar to what each
+//@ command returns. This has special methods, like `.to()` and `.toEnd()`
+var ShellString = function (str, stderr) {
+  var that = new String(str);
+  that.to = wrap('to', _to, {idx: 1});
+  that.toEnd = wrap('toEnd', _toEnd, {idx: 1});
+  that.stdout = str;
+  that.stderr = stderr;
+  return that;
+};
+
 exports.ShellString = ShellString;
 
 // Return the home directory in a platform-agnostic way, with consideration for
@@ -225,6 +246,13 @@ function wrap(cmd, fn, options) {
             return accum;
           }
         }, []);
+        // Convert ShellStrings to regular strings
+        args = args.map(function(arg) {
+          if (arg.constructor.name === 'String') {
+            return arg.toString();
+          } else
+            return arg;
+        });
         // Expand the '~' if appropriate
         var homeDir = getUserHome();
         args = args.map(function(arg) {
--- a/README.md
+++ b/README.md
@@ -331,8 +331,9 @@ Examples:
 cat('input.txt').to('output.txt');
 ```
 
-Analogous to the redirection operator `>` in Unix, but works with JavaScript strings (such as
-those returned by `cat`, `grep`, etc). _Like Unix redirections, `to()` will overwrite any existing file!_
+Analogous to the redirection operator `>` in Unix, but works with
+ShellStrings (such as those returned by `cat`, `grep`, etc). _Like Unix
+redirections, `to()` will overwrite any existing file!_
 
 
 ### 'string'.toEnd(file)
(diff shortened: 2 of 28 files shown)


ariporad:
@nfischer: It looks like this was implemented on top of #344, is that correct? If so, it looks like we'll have to do some git-fu to fix it.

src/common.js
```diff
@@ -250,6 +258,13 @@ function wrap(cmd, fn, options) {
             return accum;
           }
         }, []);
+        // Convert ShellStrings to regular strings
+        args = args.map(function(arg) {
+          if (arg.constructor.name === 'String') {
```
ariporad: You can just do `arg + ''` or `arg.toString()`.

ariporad:
@nfischer: Maybe I'm missing something, but how is returning a `ShellString` any different from returning a normal string?

Also, what are your thoughts on converting them to `ShellString`s inside `common.wrap`, vs. the command itself.

nfischer:
@ariporad Convering in common.wrap is fine with me, as long as it works.

JavaScript has `string`s and `String`s. The capital String is an object. The two are very similar (the same methods). They differ in type though, and that's significant. Because Strings are objects, we can assign properties to them without modifying the prototype (and that's the important part). ShellStrings are intentionally very similar to Strings. We can easily add features though, like .to(), pipes, stderr, code, etc

src/common.js
nfischer: That's what I'm doing in the next line, isn't it? This prevents us from calling toString() on things like vanilla objects and integers, which wouldn't be ideal (chmod takes an integer, touch sometimes takes an object containing options like -r)

nfischer:
This still isn't ready to merge. I'll ping this thread when it's ready for review.

nfischer:
@ariporad Please take a look at my implementation. This is quite a big refactor, but I think it has some nice benefits.

 - This fixes #159 (unfortunately, this breaks backwards compatibility, but I think it's worth it)
 - This will make it easy to add support for pipes, once everything is `common.wrap`ed (blocked by #362)
 - This makes sure `exec()` (synchronous) returns a `ShellString`, which allows users to write things like `exec('git status').to('file.txt')` instead of `exec('git status').stdout.to('file.txt')`. This makes the second syntax actually invalid, and it also removes the deprecated `output` attribute (we can add that back pretty easily though if it's important).

We can also consider extending this type to have a `.forEach()` method that iterates over each line in the standard output. Ex:

```Javascript
cat('file.txt').forEach(function (line) {
  echo("  " + line);
}); // indent every line of the file in stdout
```

This might make it easy to refactor `ls()` to return a ShellString, and still allow users to iterate over its output. This hasn't been implemented yet, however, and probably should be a separate PR.

src/common.js
ariporad: Oh, yeah.

src/common.js
```diff
@@ -46,11 +52,26 @@ function error(msg, _continue) {
...
+//@ ```
+//@
+//@ Turns a regular string into a string-like object similar to what each
+//@ command returns. This has special methods, like `.to()` and `.toEnd()`
+var ShellString = function (str, stderr) {
+  var that = new String(str);
+  that.to = wrap('to', _to, {idx: 1});
+  that.toEnd = wrap('toEnd', _toEnd, {idx: 1});
+  that.stdout = str;
+  that.stderr = typeof stderr !== 'undefined' ? stderr : state.error;
```
ariporad: I don't think we should default to `state.error` here, because:

```javascript
var str = ShellString('foo');
str.stderr; // This should be null, having it be anything else wouldn't make any sense.
str.to('foo.txt');
```

ariporad:
@nfischer: Is it possible to convert the result to a `ShellString` in `common.wrap`? Otherwise, LGTM!

src/common.js
nfischer: The rationale was for commands like `sed()`, which can sometimes have `stderr` output, and still return a value (imagine one file exists, and one file doesn't exist).

One alternative would be to not default it, but to have each call to `ShellString()` explicitly have a `stderr` argument if it needs it.

ariporad: I think that would be better.

nfischer:
@ariporad I thought about putting it in `common.wrap()`. The issue is that some commands return arrays (i.e. `ls()`), some return nothing (i.e. `cp()`), and some probably should return plain strings (like `tempdir()`). With some cleverness, we might be able to refactor for that, but it might just compliment things more (like an argument `{idx: 1, shellstring: true}`).

ariporad:
Ok.

LGTM!


Ari

src/common.js
nfischer: Done. Will push in a second

nfischer:
@ariporad Ok, the issues should be resolved now. LGTM, feel free to merge

nfischer:
@ariporad I went ahead and added some of these methods to the array returned by `ls()`, since I think this is a good direction to go. I left it as a separate commit for now, so it should be easier for you to find those changes. Let me know what you think.

If everything looks good, feel free to squash the changes together (or not, doesn't matter to me), and merge.

After this is merged, I can redo #269 so that each ShellString also has a `code` attribute (so it's similar to the object `exec()` returns).

ariporad:
@nfischer: One small nit, but it's not that big a deal, so I'll say LGTM.

As for the `code` prop, I say that we should just make `error()` return the code, because everything returns `stderr`.

ariporad:
Merge at will.

Issue #20, opened 2012-07-18
Cannot echo a string that starts with -


scottgonzalez:
I assume this is because `wrap()` checks for the first character of the first argument being `"-"`

https://github.com/arturadib/shelljs/blob/b5aa580668a80f7dfa59b91fcabf2048f05a6d0b/shell.js#L1035

PR #362, opened 2016-02-17
fix(verbose): verbose-style logging is consistent

nfischer:
This fixes a couple issues with `config.verbose` logging, and some related things.

 - `set('-v')` previously didn't log `echo()` commands because those weren't `common.wrap`'ed
 - `echo()` is now wrapped, but is designed to still allow echoing strings that start with "-" (ex. "-1*3+5") (see issue #20)
 - `echo()` still does not glob. This is because 1. globbing isn't that useful for `echo()` in my opinion, and 2. echo is frequently used enough that it should have good performance by default
 - internal errors are now outputted to stderr instead of stdout
 - `set('-v')`-style logging goes to stderr, not stdout. This resembles Bash's behavior with `set -v`

The way this is implemented, `echo()` still doesn't support any options, but we may want to consider allowing for that in the future (`-n` comes to mind). It'd be pretty easy to add support for that option once this is merged.

Also, it simplifies things if `echo()` is wrapped, since it ensures that commands like `echo(pwd())` still work the same, even if #360 lands.

### DIFF
--- a/test/set.js
+++ b/test/set.js
@@ -37,14 +37,16 @@ assert(result.stderr.indexOf('Error: ls: no such file or directory: file_doesnt_
 // set -v
 var result = shell.exec('node -e \"require(\'../global\'); set(\'-v\'); ls(\'file_doesnt_exist\'); echo(1234);\"');
 assert.equal(result.code, 0);
-assert.equal(result.stdout, 'ls file_doesnt_exist\n1234\n');
-assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n');
+assert.equal(result.stdout, '1234\n');
+assert.equal(result.stderr, 'ls file_doesnt_exist\nls: no such file or directory: file_doesnt_exist\necho 1234\n');
 
 // set -ev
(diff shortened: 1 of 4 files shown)


ariporad:
Can we add a test that verbose mode works with echo?

Otherwise than that, LGTM!

nfischer:
@ariporad I fixed one of the tests to check that echo is logged (it previously allowed it not to be logged, since it wasn't wrapped)

test/set.js
```diff
@@ -37,14 +37,16 @@ assert(result.stderr.indexOf('Error: ls: no such file or directory: file_doesnt_
 // set -v
 var result = shell.exec('node -e \"require(\'../global\'); set(\'-v\'); ls(\'file_doesnt_exist\'); echo(1234);\"');
 assert.equal(result.code, 0);
-assert.equal(result.stdout, 'ls file_doesnt_exist\n1234\n');
-assert.equal(result.stderr, 'ls: no such file or directory: file_doesnt_exist\n');
+assert.equal(result.stdout, '1234\n');
+assert.equal(result.stderr, 'ls file_doesnt_exist\nls: no such file or directory: file_doesnt_exist\necho 1234\n');
```
nfischer: @ariporad here's an example where it tests `echo()` being logged

ariporad:
LGTM!

PR #364, opened 2016-02-18
test(ln): add tests for linking to cwd

### LINKED ISSUE #363: ln('-sf', './', '<destination>') is not linking the right folder
https://travis-ci.org/TWExchangeSolutions/eslint-config-exchange-solutions/jobs/107132691

From what I can see there's a few issues.

1. For some reason the [`resolvedSourcePath`](https://github.com/shelljs/shelljs/blob/0da963b51f1895440ac4c96fb14248de1f028f58/src/ln.js#L46) is built up with the `dest` path?
2. On Mac OSX you don't even use the `resolvedSourcePath` when [creating the symlink](https://github.com/shelljs/shelljs/blob/0da963b51f1895440ac4c96fb14248de1f028f58/src/ln.js#L54), but it does not appear to be respecting the current working directory. This worked in 0.5.3 which was using the correct `resolvedSourcePath`.

Is there a reason not to use the `resolvedSourcePath` for all operating systems?

I can put together a PR for this, just trying to gather any backstory I'm missing.

nfischer:
Add tests for symlinking to `./` (this closes #363 if the behavior here is correct). If this behavior is not posix behavior, we should fix this instead.

### DIFF
--- a/test/ln.js
+++ b/test/ln.js
@@ -105,6 +105,24 @@ shell.ln('-s', 'ln', 'tmp/dir1');
 assert(fs.existsSync('tmp/ln/hello'));
 assert(fs.existsSync('tmp/dir1/hello'));
 
+// To current directory
+shell.cd('tmp');
+shell.ln('-s', './', 'dest');
+shell.touch('testfile.txt');
+assert(fs.existsSync('testfile.txt'));
+assert(fs.existsSync('dest/testfile.txt'));
+shell.rm('-f', 'dest');
+shell.mkdir('dir1');
+shell.cd('dir1');
+shell.ln('-s', './', '../dest');
+shell.touch('insideDir.txt');
+shell.cd('..');
+assert(fs.existsSync('testfile.txt'));
+assert(fs.existsSync('dest/testfile.txt'));
+assert(fs.existsSync('dir1/insideDir.txt'));
+assert(!fs.existsSync('dest/insideDir.txt'));
+shell.cd('..');
+
 shell.ln('-f', 'tmp/file1.js', 'tmp/file2.js');
 assert(fs.existsSync('tmp/file2.js'));
 assert.equal(


test/ln.js
```diff
@@ -105,6 +105,24 @@ shell.ln('-s', 'ln', 'tmp/dir1');
...
+shell.rm('-f', 'dest');
+shell.mkdir('dir1');
+shell.cd('dir1');
+shell.ln('-s', './', '../dest');
+shell.touch('insideDir.txt');
+shell.cd('..');
+assert(fs.existsSync('testfile.txt'));
+assert(fs.existsSync('dest/testfile.txt'));
+assert(fs.existsSync('dir1/insideDir.txt'));
+assert(!fs.existsSync('dest/insideDir.txt'));
```
mtscout6: This file should exist right? If it doesn't then the dest folder wasn't symlinked properly from the dir1 folder.

nfischer: I think it shouldn't. Symbolic relative links are relative to the destination, hard links are relative to the cwd.

`insideDir.txt` is inside `dir1`, not its parent directory, and `dest` points at `dir1`'s parent

Try out this behavior with UNIX Ln and I think it should behave the same

mtscout6: Ah, you're right. Not what I would have expected. Thanks for following up on this.

mtscout6: For posterity, this is the test I ran manually:

```
➜  ~  docker run -it --rm ubuntu
root@784ff6f8e5e4:/# cd tmp
root@784ff6f8e5e4:/tmp# mkdir dir1
root@784ff6f8e5e4:/tmp# cd dir1/
root@784ff6f8e5e4:/tmp/dir1# ln -s ./ ../dest
root@784ff6f8e5e4:/tmp/dir1# touch insideDir.txt
root@784ff6f8e5e4:/tmp/dir1# cd ../
root@784ff6f8e5e4:/tmp# ls
dest  dir1
root@784ff6f8e5e4:/tmp# ls dest
dest  dir1
root@784ff6f8e5e4:/tmp# cd dest
root@784ff6f8e5e4:/tmp/dest# ls
dest  dir1
root@784ff6f8e5e4:/tmp/dest# exit
exit
```

nfischer: It's a very confusing semantics. Hopefully the `path.resolve()` trick is good enough to fix things. I personally usually use full paths with GNU `ln` To get around the weirdness

nfischer:
@ariporad this is ready for review

ariporad:
@nfischer: Umm... I don't think I really understand this well enough to properly review it. It seems very confusing. I'll do some more research, and get back to you.

nfischer:
@ariporad this is related to how `ln` handles relative symbolic links (see #282). I think POSIX has somewhat confusing behavior, but at least it's consistent. This was a change we introduced in v0.6, I'm just adding a couple extra tests to make sure we don't get a regression.

ariporad:
@nfischer: Ok, LGTM!

ariporad:
LGTM!

PR #366, opened 2016-02-20
docs(exec): exec options only work for v0.11+

### LINKED ISSUE #350: Document that exec() options don't work on early versions of node
As far as I can tell, options to `exec()` and `execSync()` only work for node v0.11+. This should probably get documented in our README. If enough people are interested, we could try some hacks to get it to work (we could do this for `cwd`, for example), but I don't think it's worth it, since it really only affects v0.10 users (since we don't support v0.8 anymore).

nfischer:
Resolves #350 

As far as I can tell, node `v0.10` doesn't support many options to `child_process.exec()`. This doesn't really fix it, but it does document that these options don't work for that platform.

Whether or not we officially support `v0.10` (currently we test against it for unit tests), there are still users out there using `v0.10`, and I'm sure some of them will still try to mix ShellJS with it. I think, since this is a known issue, it's worth documenting (since there's no good workaround). Most users won't be affected by it, and the few that are affected will be aware (and maybe switch to a newer version of node because of it).

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,6 +507,7 @@ Available options (all `false` by default):
 + `silent`: Do not echo program output to console.
 + and any option available to NodeJS's
   [child_process.exec()](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
+  (only available on Node v0.11+)
 
 Examples:
 
--- a/src/exec.js
+++ b/src/exec.js
@@ -206,6 +206,7 @@ function execAsync(cmd, opts, pipe, callback) {
 //@ + `silent`: Do not echo program output to console.
 //@ + and any option available to NodeJS's
 //@   [child_process.exec()](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)
+//@   (only available on Node v0.11+)
 //@
 //@ Examples:
 //@


ariporad:
I say just don't support `v0.10`. It's approaching EOL (In October), and it's a headache.

Alternatively, we could figure out a way to polyfill the options, and then say that we support all actively supported versions of node (As of right now, `v0.10`, `v0.12`, and `v4`).

nfischer:
I view these options as kind of an edge-use-case. Why drop support for all v0.10 users over this? If I were a v0.10 user, I'd rather see the one non-working feature documented, rather than just assume the whole thing doesn't work because package.json says so, **or worse**: assume everything works because I don't think to check which versions are in package.json.

An alternative would be to "officially" support v0.12+, declare this in our README, and to just also test against v0.10 to see if there are easy breaks that we can fix. And if we hit a huge number of unsupported features, we can consider dropping v0.10 early.

But I agree, once it hits EOL, let's drop it regardless. At that point, the new Ubuntu LTS will use node 4, so I think very few users will actually use v0.10.

ariporad:
@nfischer: `npm` does warn if the versions don't match, but I think we should officially support `v0.12`+, but test against `v0.10`.

ariporad:
I'll submit a PR to do that.

nfischer:
I'm fine with that. Either way, I think we should update the docs here and also at the top of the README, declaring what versions we officially support

nfischer:
@ariporad What do you think of this? Should we merge it in, or close this? Now that we only support `v0.11+`, this is probably less necessary. The pro is that `v0.10` users won't have as many surprises. The con is that it clutters the docs a bit.

ariporad:
I'm going to vote :-1:, I think it's going to be confusing.

PR #368, opened 2016-02-20
chore: drop node v0.10 support

ariporad:
Drop Node `v0.10` support and document.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,17 @@ language: c++
 sudo: false
 env:
   - NODE_VERSION="0.10"
+  - NODE_VERSION="0.11"
   - NODE_VERSION="0.12"
+  - NODE_VERSION="iojs-v1"
+  - NODE_VERSION="iojs-v2"
+  - NODE_VERSION="iojs-v3"
   - NODE_VERSION="4"
   - NODE_VERSION="5"
+
+matrix:
+  allow_failures:
+    - NODE_VERSION="0.10"
 
 # keep this blank to make sure there are no before_install steps
 before_install:
@@ -21,11 +29,13 @@ os:
   - osx
 script:
   - npm test
+  # make sure when the docs are generated nothing changes (a.k.a. the docs have already been generated)
   - node scripts/generate-docs.js
   - git diff --quiet # make sure no files have changed
 
 # Gitter
 notifications:
+  email: false
   webhooks:
     urls:
       - https://webhooks.gitter.im/e/743dc6b83b6dd05111ee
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,10 +2,18 @@ language: c++
 sudo: false
 env:
   - NODE_VERSION="0.10"
+  - NODE_VERSION="0.11"
   - NODE_VERSION="0.12"
+  - NODE_VERSION="iojs-v1"
```
ariporad: I realized we weren't testing these, so I added them.

nfischer: We intentionally dropped v0.11 support because it's supposed to be compatible with v0.12. See #307.

Artur took out iojs support a month or so ago. Didn't iojs and node merge, so there's no reason to test against these versions? I could be mistaken, so please correct me if I'm wrong.

nfischer: ^ the commit I'm referring to is 6acfbe858f3ec453730bb5967837d7618c73ea9

ariporad: @nfischer: I know `v0.11` is supposed to be the same as `v0.12`, but I feel like it's probably worth testing.

As for iojs, yes, node >`4` are actually iojs releases. So the release line looks like this(-ish):

```
v0.9 --- v0.10 ---- v0.11 --- v0.12 ---- v0.13 --- v0.14 --- v4.0 --- v5.0 --- [Future]
               \                                           /
                 ------ v1.0 ------ v2.0 ------ v3.0 -----
```

nfischer: If iojs is merged back into the project, shouldn't it be ok to drop support for it? Is iojs actually still supported, or was it dropped in favor of node 4?

From #307:

> 0.11 is an unstable version that is not supported or updated anymore. 0.12 is fully compatible with 0.11 and as it is in maintenance mode, no breaking changes will be committed there.

Since it just slows CI down, and I don't see a strong benefit, I'd vote against including these versions.

ariporad: I'm not super strong on it, but I feel like we should test any version that we claim to support, just in case theres an incompatibilities.

nfischer: In that case, I'd rather stick with supporting `v0.12`, `v4`, and `v5` in that case. It's essentially the same as what we're testing on now (minus `v0.10`), and we don't have to worry about all the differences iojs might have going forward.

ariporad: Hmm...

I think that we should support (and therefore test) any active, LTS, or maintenance version of node (less `v0.10`). That would mean `v0.11`, `v0.12`, all releases of iojs, and node 4/5.

I doubt it would slow down travis that much, but if we really wanted to make it run fast, we could enable travis [fast finishing](https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing).

nfischer: I suppose this is fine then

nfischer:
@ariporad Please update the `engines` attribute inside package.json to be `>=0.11`. Then this is good to merge.

ariporad:
@nfischer: Good Catch! Fixed.

nfischer:
LGTM!

PR #369, opened 2016-02-21
refactor(ls): greatly simplify ls implimentation

ariporad:
This greatly simplifies the implementation of `ls`, and probably makes it a good deal more performant.

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -23,7 +23,7 @@ exports.pwd = common.wrap('pwd', _pwd);
 
 //@include ./src/ls
 var _ls = require('./src/ls');
-exports.ls = common.wrap('ls', _ls, {idx: 1});
+exports.ls = common.wrap('ls', _ls, {noGlob: true});
 
 //@include ./src/find
 var _find = require('./src/find');
--- a/shell.js
+++ b/shell.js
@@ -23,7 +23,7 @@ exports.pwd = common.wrap('pwd', _pwd);
 
 //@include ./src/ls
 var _ls = require('./src/ls');
-exports.ls = common.wrap('ls', _ls, {noGlob: true});
+exports.ls = common.wrap('ls', _ls, {idx: 1});
 
 //@include ./src/find
 var _find = require('./src/find');
--- a/src/ls.js
+++ b/src/ls.js
@@ -1,8 +1,10 @@
 var path = require('path');
 var fs = require('fs');
 var common = require('./common');
-var _cd = require('./cd');
-var _pwd = require('./pwd');
+var glob = require('glob');
+
+var globPatternAll = path.sep + '*';
+var globPatternRecrusive = path.sep + '**' + globPatternAll;
 
 //@
 //@ ### ls([options,] [path, ...])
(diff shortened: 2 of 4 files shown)


nfischer:
@ariporad does this work for `ls('-d')`? I noticed you cut out the mention of `options.directory`, which makes me suspicious this could work.

Also, this breaks some additional tests, so those should be resolved first.

ariporad:
@nfischer: So, we appear to have some inconsistent ls behavior. If your pass in a file/pattern with no matches, should it have an error, or just return an array with nothing in it?

nfischer:
@ariporad I accounted for this in `common.expand()`. In bash:

```Bash
$ ls this*file*doesnt*exist.txt
ls: cannot access this*file*doesnt*exist.txt: No such file or directory
# bash passes a literal 'this*file*doesnt*exist.txt' to the ls executable
```

ariporad:
@nfischer: So if it can't find the file it should error?

nfischer:
That's what `ls()` should do, yes. That's why I use `common.expand()` to inside `common.wrap()`. It will pass non-existing file names to commands (like `ls()`). Then those commands are expected to handle the non-existing file themselves (in `ls()`'s case, it should output an error message).

nfischer:
`glob.sync()` returns an empty array, I think, which probably isn't what you want

nfischer:
@ariporad pinging this to see if there's any work being done to re-add support for `-l` and `-d`.

ariporad:
@nfischer: Sorry, I've been sick. I'll work on that ASAP.

ariporad:
@nfischer: Both `-l` and `-d` should be back now.

shell.js
```diff
@@ -23,7 +23,7 @@ exports.pwd = common.wrap('pwd', _pwd);
 
 //@include ./src/ls
 var _ls = require('./src/ls');
-exports.ls = common.wrap('ls', _ls, {idx: 1});
+exports.ls = common.wrap('ls', _ls, {noGlob: true});
```
nfischer: Why was this taken out? Also, does `noGlob` even do anything here?

ariporad: We don't want to glob in common.wrap, we want to do it in late itself, because we want to optionally pass different opts to common.expand depending on the flags passed in.

src/ls.js
```diff
@@ -1,8 +1,10 @@
 var path = require('path');
 var fs = require('fs');
 var common = require('./common');
-var _cd = require('./cd');
-var _pwd = require('./pwd');
+var glob = require('glob');
+
+var globPatternAll = path.sep + '*';
+var globPatternRecrusive = path.sep + '**' + globPatternAll;
```
nfischer: s/Recrusive/Recursive (and the places below)

shell.js
nfischer: I don't think `noGlob` is a real option (no `idx` value will have the same effect)

Also, would it be possible to modify `common.expand()` to take `all` as an argument? It might be cleaner to keep `glob` inside `common.js`

test/ls.js
```diff
@@ -144,7 +144,7 @@ assert.equal(result.indexOf('resources/ls/file1.js') > -1, true);
 
 // wildcard, should not do partial matches
 var result = shell.ls('resources/ls/*.j'); // shouldn't get .js
-assert.equal(shell.error(), null);
+assert.ok(shell.error());
 assert.equal(result.length, 0);
```
nfischer: We should remove this line, since this is an invalid case

test/ls.js
```diff
@@ -144,7 +144,7 @@ assert.equal(result.indexOf('resources/ls/file1.js') > -1, true);
 
 // wildcard, should not do partial matches
 var result = shell.ls('resources/ls/*.j'); // shouldn't get .js
-assert.equal(shell.error(), null);
+assert.ok(shell.error());
```
nfischer: Not sure why this was the opposite before, but this seems like it's correct now. Good fix

test/ls.js
```diff
@@ -249,7 +249,6 @@ assert.equal(result.length, 6);
 // long option, single file
 var result = shell.ls('-l', 'resources/ls/file1')[0];
 assert.equal(shell.error(), null);
-assert.equal(result.name, 'resources/ls/file1');
```
nfischer: Why is this line removed? This looks correct to me

src/ls.js
```diff
@@ -46,119 +48,51 @@ function _ls(options, paths) {
 
   if (!paths)
     paths = ['.'];
-  else if (typeof paths === 'object')
-    paths = paths; // assume array
   else if (typeof paths === 'string')
     paths = [].slice.call(arguments, 1);
 
-  var list = [];
+  paths = common.expand(paths, { dot: options.all });
```
nfischer: Why not perform this in common.wrap? This doesn't give correct behavior if you do it here like this. Doing it in `common.wrap` is the easiest quick fix.

```Bash
$ ls -A
.secret.txt foo.txt bar.txt
$ ls *.txt # this only shows non-hidden files
foo.txt bar.txt
$ cat test.js
require('shelljs/global');
echo(ls('*.txt')); // this shouldn't output secret files either
$ node test.js # but it actually does
.secret.txt foo.txt bar.txt
```

shell.js
ariporad: It's a real option.

Moving the all option into wrap would be really complex, and probobly not that useful. ls needs to do some weird stuff with common.expand, so I think we should just keep it in ls.

src/ls.js
ariporad: I cannot believe I did that. :faceplam:

test/ls.js
ariporad: That was a mistake. I'll put it back.

shell.js
nfischer: `config.noGlob` is real, `options.noGlob` isn't. You're specifying `options.noGlob` here (just like how `idx` is really `options.index` [here](https://github.com/shelljs/shelljs/blob/master/src/common.js#L275))

nfischer:
@ariporad I'm going to make some adjustments to this branch and see if I can fix some of the errors we're seeing.

ariporad:
Ok, thanks! If not, I'll try to look into it tomorrow or when I get back.

nfischer:
@ariporad Take a look now. I fixed the bug you had for Windows (you forgot `/g` on a regex, that's all). Nice PR!

I tested performance and saw that this was very very slow... So I did my best to squeeze out some more performance. This is about 2x faster than your initial revision (and about 2x uglier...). This is still slower (also about 2x...) than the master branch, but it should be a lot cleaner.

So it's a tradeoff. More maintainable and a bit slower, or we can stick with the tried and true method and get faster.

As for the tests, only one of them *needed* to change (the one you changed for `ls('...*.j')`). I just changed some others because I realized I could format them a bit nicer, make them more robust (the tests using `idx` were actually written poorly, my fault), and let them catch more.

I also fixed a but where `ls -l resources/ls` would return values with names including `resources/ls` (they should be relative to `resources/ls`, not the current directory), so that's what's going on there.

nfischer:
@ariporad This should be rebased off master. If you can, see if you find ways to improve performance (or refactor things into a function), since we're still taking a performance hit for the normal use case, which isn't ideal.

ariporad:
@nfischer: I'm going to try something, give me a second.

ariporad:
@nfischer: Actually, let me talk this through with you first.

Personally I vote for slower and more maintainable, it's not like `ls` is a big bottleneck.

It also might be better to try to hijack `fs.stat`, so common.wrap looks something like this:

```javascript
function wrap(cmd, fn, options) {
  return function() {
    var oldStat = fs.stat;
    var statResults = {};
    fs.stat = function (file) {
        if (!statResults[file]) statResults[file] = oldStat.apply(this, arguments);
        return statResults[file];
    }

    // Whatever common.wrap used to do

    fs.stat = oldStat;
  };
}
```

I bet this would speed things up a bit.

ariporad:
Nope, no it wouldn't.

But I did learn how to benchmark node things.

nfischer:
@ariporad I just pushed another unit test to `ls()` to make sure everything is working nicely. Assuming everything still passes (it should), does this look good to merge?

ariporad:
LGTM!

ariporad:
Oh, wait, I have to rebase this off master.

ariporad:
LGTM

PR #370, opened 2016-02-21
feat(pipe): add support for pipes between commands

### LINKED ISSUE #148: pipe to proc
Does shelljs have something to simulate "`command1 | command2`"?

A possible usage:
```javascript
echo(mylist.join('\n')).pipe('egrep "[0-9]"')
```

nfischer:
Fixes #148

Commands can now be piped, of the form:

```javascript
grep('foo', 'file1.txt', 'file2.txt').sed(/o/g, 'a').to('output.txt');
echo('files with o\'s in the name:\n' + ls().grep('o'));
cat('test.js').exec('node'); // pipe to exec() call
```

Commands can send their output to another command in a pipe-like fashion. `sed`, `grep`, `cat`, `exec`, `to`, and `toEnd` can appear on the right-hand side of a pipe. Pipes can be chained. You can even pipe into async `exec()`.

This also improves performance slightly. I got rid of the slow, home-grown `escape()` function inside of `src/exec.js` and replaced it with `JSON.stringify()`, which I measured to be slightly faster, and much more reliable (it properly escapes newlines, tabs, etc. as well).

I couldn't think of any tests to add for piping to `exec()` on Windows (can't think of any native commands that can be on the right-hand side of a pipe). I'm fairly certain this code does work on Windows, I just couldn't think of a use-case. Worst case, we can go back and have someone more Windows-savvy make a PR adding the tests.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -5,6 +5,7 @@
 var os = require('os');
 var fs = require('fs');
 var glob = require('glob');
+var shell = require('..');
 var _to = require('./to');
 var _toEnd = require('./toEnd');
 
@@ -57,18 +58,27 @@ exports.error = error;
 //@
 //@ Examples:
 //@
-//@ ```
+//@ ```javascript
 //@ var foo = ShellString('hello world');
 //@ ```
 //@
 //@ Turns a regular string into a string-like object similar to what each
 //@ command returns. This has special methods, like `.to()` and `.toEnd()`
-var ShellString = function (str, stderr) {
-  var that = new String(str);
-  that.to = wrap('to', _to, {idx: 1});
-  that.toEnd = wrap('toEnd', _toEnd, {idx: 1});
-  that.stdout = str;
+var ShellString = function (stdout, stderr) {
+  var that;
+  if (stdout instanceof Array) {
+    that = stdout;
+    that.stdout = stdout.join('\n')+'\n';
+  } else {
+    that = new String(stdout);
+    that.stdout = stdout;
+  }
   that.stderr = stderr;
+  that.to    = function() {wrap('to', _to, {idx: 1}).apply(that.stdout, arguments); return that;};
+  that.toEnd = function() {wrap('toEnd', _toEnd, {idx: 1}).apply(that.stdout, arguments); return that;};
+  ['cat', 'sed', 'grep', 'exec'].forEach(function (cmd) {
+    that[cmd] = function() {return shell[cmd].apply(that.stdout, arguments);};
+  });
   return that;
 };
 
@@ -281,3 +291,8 @@ function wrap(cmd, fn, options) {
   };
 } // wrap
 exports.wrap = wrap;
+
+function _readFromPipe(that) {
+  return that instanceof String ? that.toString() : '';
+}
+exports.readFromPipe = _readFromPipe;
--- a/src/ls.js
+++ b/src/ls.js
@@ -3,8 +3,6 @@ var fs = require('fs');
 var common = require('./common');
 var _cd = require('./cd');
 var _pwd = require('./pwd');
-var _to = require('./to');
-var _toEnd = require('./toEnd');
 
 //@
 //@ ### ls([options,] [path, ...])
@@ -154,15 +152,7 @@ function _ls(options, paths) {
   });
 
   // Add methods, to make this more compatible with ShellStrings
-  list.stdout = list.join('\n') + '\n';
-  list.stderr = common.state.error;
-  list.to = function (file) {
-    common.wrap('to', _to, {idx: 1}).call(this.stdout, file);
-  };
-  list.toEnd = function(file) {
-    common.wrap('toEnd', _toEnd, {idx: 1}).call(this.stdout, file);
-  };
-  return list;
+  return new common.ShellString(list, common.state.error);
 }
 module.exports = _ls;
 
--- a/src/sed.js
+++ b/src/sed.js
@@ -22,6 +22,9 @@ function _sed(options, regex, replacement, files) {
     'i': 'inplace'
   });
 
+  // Check if this is coming from a pipe
+  var pipe = common.readFromPipe(this);
+
   if (typeof replacement === 'string' || typeof replacement === 'function')
     replacement = replacement; // no-op
   else if (typeof replacement === 'number')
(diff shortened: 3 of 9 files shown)


src/common.js
```diff
@@ -67,6 +71,10 @@ var ShellString = function (str, stderr) {
   var that = new String(str);
   that.to = wrap('to', _to, {idx: 1});
   that.toEnd = wrap('toEnd', _toEnd, {idx: 1});
+  that.cat = wrap('cat', _cat, {idx: 1});
```
ariporad: Can this just be `shell.to`? It seems like repeatedly re-wrapping command isn't performant, and it's not DRY.

src/ls.js
```diff
@@ -162,6 +166,18 @@ function _ls(options, paths) {
   list.toEnd = function(file) {
     common.wrap('toEnd', _toEnd, {idx: 1}).call(this.stdout, file);
   };
+  list.cat = function() {
+    return common.wrap('cat', _cat, {idx: 1}).apply(this.stdout, arguments);
```
ariporad: Two things:
1. See my comment above about re-wrapping commands.
2. Idea: What if we turn `ShellString` into something like this:

```javascript
function ShellString(that, err) {
    that = new String(that);
    return ShellStringify(that, that, err);
}

function ShellStringify(that, stdout, stderr) {
    that.stdout = stdout || that.toString();
    that.stderr = stderr || '';
    ['toEnd', 'end', 'cat', 'sed', 'grep', 'exec'].forEach(function(method) {
        that[method] = shell[method];
    });
    return that;
}
```

Then we can use it like this:

```javascript
function _ls(...) {
    // ...
    var list = [...];
    return common.ShellStringify(list, list.join('\n'));
}

function _doSomethingElse(...) {
    // ...
    var list = [...];
    return new common.ShellString('foo');
}

function _somethingBadHappened(...) {
    // ...
    var list = [...];
    return new common.ShellString(null, null, 'everything\'s broken!');
}
```

nfischer: I actually just started refactoring `ShellString()` to have a similar affect. It still takes 2 arguments, but the first is the stdout and the second is the stderr. stdout can be either an array or a string, and it'll return either a "ShellArray" (an array with all the extra methods), or a `ShellString`, respectively. This simplifies the code in `src/ls.js`, which I suspect is what you were going for.

I've also fixed the re-wrapping issue, and removed all the extra `require()`s from `src/ls.js`.

src/common.js
nfischer: In terms of performance, it actually doesn't hurt much. `wrap()` is actually very quick, since it just returns another function. But I agree with the DRY part, so this should be fixed now.

nfischer:
@ariporad this should be a much cleaner implementation now. `ShellString()` is a little different than you discussed, but I think it solves the same problem. Let me know what you think.

src/common.js
```diff
@@ -57,18 +58,28 @@ exports.error = error;
...
+  if (stdout instanceof Array) {
+    that = stdout;
+    that.stdout = stdout.join('\n')+'\n';
+  } else {
+    that = new String(stdout);
+    that.stdout = stdout;
+  }
   that.stderr = stderr;
+  that.to    = function() {wrap('to', _to, {idx: 1}).apply(that.stdout, arguments); return that;};
+  that.toEnd = function() {wrap('toEnd', _toEnd, {idx: 1}).apply(that.stdout, arguments); return that;};
```
ariporad: Why are we still re-wraping these?

src/common.js
```diff
@@ -57,18 +58,28 @@ exports.error = error;
...
+  that.cat   = function() {return shell.cat.apply(that.stdout, arguments);};
+  that.sed   = function() {return shell.sed.apply(that.stdout, arguments);};
+  that.grep  = function() {return shell.grep.apply(that.stdout, arguments);};
+  that.exec  = function() {return shell.exec.apply(that.stdout, arguments);};
```
ariporad: Small, optional nit: maybe iterate over an array of method names?

src/common.js
```diff
@@ -57,18 +58,28 @@ exports.error = error;
...
+    that = stdout;
+    that.stdout = stdout.join('\n')+'\n';
+  } else {
+    that = new String(stdout);
+    that.stdout = stdout;
+  }
   that.stderr = stderr;
+  that.to    = function() {wrap('to', _to, {idx: 1}).apply(that.stdout, arguments); return that;};
+  that.toEnd = function() {wrap('toEnd', _toEnd, {idx: 1}).apply(that.stdout, arguments); return that;};
+  that.cat   = function() {return shell.cat.apply(that.stdout, arguments);};
```
ariporad: Small nitpick: maybe bind these instead of using a closure.

src/common.js
nfischer: I believe these are only wrapped in this spot (they're methods, so they're wrapped here now).

src/common.js
nfischer: Whoops! Forgot to add that in. I'll do that, thanks!

test/pipe.js
```diff
@@ -0,0 +1,83 @@
...
+shell.rm('-rf', 'tmp');
+shell.mkdir('tmp');
+
+//
+// Invalids
+//
+
+// no piped methods for commands that don't return anything
+assert.throws(function() {
+  shell.cp('rm filedoesntexist').cat();
```
ariporad: Maybe make this a valid call to cp? Just to remove variables?

src/sed.js
```diff
@@ -22,6 +22,9 @@ function _sed(options, regex, replacement, files) {
     'i': 'inplace'
   });
 
+  // Check if this is coming from a pipe
+  var pipe = common.readFromPipe(this);
```
ariporad: Idea: have a 'pipe' option for common.wrap, when enabled pass pipe as the last arg. (Optional)

test/pipe.js
nfischer: SGTM

src/common.js
nfischer: I tried binding instead, but it wound up being even more complicated, and still wasn't working completely correctly. I think it's simplest to just leave as closures.

src/sed.js
nfischer: Hmmm... not sure if that would be any cleaner. I thought about prepending `'-'` to the list of files inside `common.wrap()`, but individual utilities would still need to check for the piped input, so it wouldn't save us much.

test/pipe.js
nfischer: Done.

src/common.js
nfischer: Done. I did something similar to what you proposed above.

src/common.js
ariporad: Is is possible to cache the call to common.wrap?

src/sed.js
ariporad: Ok.

ariporad:
One small nit, otherwise looks good.

src/common.js
nfischer: Not sure of a good way to do this, because if I move it out of the function, then `that` isn't defined anymore. So then the code only works for strings, not arrays as well. So I think we might have to leave it as-is. The code should still be very efficient, since `wrap` is just a closure.

Or we can modify `.to()` to `join` `this` if it's an array, but that would incur a performance hit.

ariporad: OK.

ariporad:
LGTM!

PR #375, opened 2016-02-26
Add Cash cross-reference

dthree:
As per:

https://github.com/dthree/cash/pull/24#

:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ The project is [unit-tested](http://travis-ci.org/shelljs/shelljs) and battled-t
 
 If you have feedback, suggestions, or need help, feel free to post in our [issue tracker](https://github.com/shelljs/shelljs/issues).
 
+For a non-programmatic implementation of Unix shell commands on top of Node.js, check out [Cash](https://github.com/dthree/cash).
+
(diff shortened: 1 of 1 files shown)


nfischer:
LGTM!

ariporad:
@nfischer: What do you think? I feel like it would be better if we linked to the wiki in the README, or just move all the related projects to the README?

nfischer:
I think we should like to our wiki in the README. I'm not opposed to linking to cash directly as well (after all, they link back to us).

dthree:
It's totally up to you guys, whatever you want to do is fine with me. Just let me know what to change.

ariporad:
Hmm...

I feel that if they're putting us in their README, we should do the same, so maybe we should just put all related projects in the README. (Plus, then it's version controlled and we have to approve changes).

nfischer:
I say that we try it out as-is, and also add a link to our wiki (I can do that as a separate commit).

nfischer:
I'm going to go ahead and merge this in. I may adjust things a bit in the README when I go through and update docs.

dthree:
Thanks! :star2:

PR #377, opened 2016-02-29
Check if Windows paths end in \ for recursive cp()

insin:
Paths which needed to be manipulated using the `path` module prior to calling `cp()` will contain `\` when running on Windows.

Found this while investigating the cause of ryanflorence/react-project#14. react-project needs to manipulate paths first because the stuff it uses `cp()` to copy is relative to its own location in `node_modules` rather than the cwd.

### DIFF
--- a/src/cp.js
+++ b/src/cp.js
@@ -143,7 +143,7 @@ function _cp(options, sources, dest) {
     // Recursive allows the shortcut syntax "sourcedir/" for "sourcedir/*"
     // (see Github issue #15)
     sources.forEach(function(src, i) {
-      if (src[src.length - 1] === '/') {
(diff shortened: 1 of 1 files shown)


nfischer:
Thanks! I'll investigate. I may refactor this slightly, and merge manually.

nfischer:
@insin Thanks for the PR! After looking at this, I decided that I didn't really like the way `cp()` was handling directories to begin with (checking for a trailing `/` seems like a silly way to check if something's a directory).

I opened #380, which would get rid of this section. Would you be able to review and confirm that this works for your needs?

insin:
Great, I'll have a look later today when I get some time.

nfischer:
Closing this since #380 has been merged.

nfischer:
@insin Thanks again for the PR (and for alerting us to this issue)!

Issue #376, opened 2016-02-27
Copy not accepting wildcards + file ending anymore when using -r


kurtextrem:
So while `ls("src/*.html")` returns an array of html files, `copy("-r", "src/*.html", "dist")` results in an error. It works without -r.
I'm on Windows 7 x64, Node 5.7, using 0.6.

ariporad:
@kurtextrem: Hi! Thanks for the bug report!

This is actually expected behavior, `cp` with the `-r` options means to copy recursively, which only works on directories. If you want to copy all `.html` files in `src/` recursively (ie. even `src/foo/bar/baz.html`), then you're looking for `cp('src/**/*.html', 'dist')`.

The error message is incorrect though, and I'll fix that.

I'm going to close this, but if I've misunderstood, please feel free to reopen!

nfischer:
I'm reopening this, since `-r` *should* be allowed on files. `cp('file.txt', 'dest/')` should do the exactly the same thing as `cp('-r', 'file.txt', 'dest/')`. Just tested this on v0.6 and there seems to be some weird behavior going on.

I'll test again off the master branch and provide a fix if I see the bug there as well. Thanks for reporting this!

kurtextrem:
Thank you for the explanation :)
Glad that my report wasn't that useless.

nfischer:
@kurtextrem As a workaround for now, you can do what @ariporad suggests (no `-r` option).

kurtextrem:
Done :+1: thanks.

nfischer:
Looks like this *was* a bug in `v0.6`, but isn't a bug in the master branch. The best kind of bugs are the ones you unintentionally fix :+1:

I've added some more unit tests for this behavior. This fix will be available in `v0.7`.

@kurtextrem could you please look over the tests and confirm that I'm correctly handling the behavior you describe? Take a look at #378. Thanks!

nfischer:
Looks like I spoke too soon...

I think this is still an issue. Steps to reproduce:

```Javascript
> touch('file.txt');
> cp('file.txt', 'file2.txt'); // this works correctly. `file2.txt` is a regular file
> cp('-r', 'file.txt', 'file3.txt'); // this is wrong. `file3.txt` is actually a directory containing `file1.txt`
```

Looks like this bug is visible [on this line in cash](https://github.com/dthree/cash/blob/master/gulpfile.js#L92).

kurtextrem:
Great, thanks!

[...]

PR #378, opened 2016-02-29
Add unit tests to prevent regression (see #376)

nfischer:
This doesn't change anything, it just adds tests. Looks like this was a bug in `v0.6`, but has been since fixed. I'll update this if I don't get feedback from #376 about this.


ariporad:
LGTM!

PR #379, opened 2016-02-29
New commands: sort(), head(), and tail()

nfischer:
These are initial versions of the `head()` and `sort()` commands. **Not ready to merge**. Feel free to take a look and give feedback on what I have implemented so far.

I haven't written unit tests yet, but I will do so soon.

The reason these are together in one PR is that I wanted to be able to hack out a translation of an adaptation of the famous [Donald Knuth shell problem](http://www.leancrew.com/all-this/2011/12/more-shell-less-egg/):

```Bash
#!/bin/bash
cat input.txt |
tr -cs A-Za-z '\n' |
tr A-Z a-z |
sort |
uniq -c |
sort -rn |
head -n $1
```

Will now (correctly) translate to:

```Javascript
#!/usr/bin/env node
require('shelljs/global');
config.silent = true;
echo(cat('input.txt')
  .exec('tr -cs A-Za-z \'\\n\'')
  .exec('tr A-Z a-z')
  .sort() // sort works!
  .exec('uniq -c')
  .sort('-rn') // it even takes options!
  .head('-n', process.argv[2])); // and head() should take options of the style `head({'-n', 5}, 'file.txt')`
```

Also, if someone wants `tail()`, chime in and I can write that out as well.


nfischer:
I just added in the test cases, so this is ready for review. If anyone from the community wants to chime in for features, please do.

The semantics are slightly different than unix's `sort`: capitalized words are sorted before non-capitalized words in this function, but unix `sort` doesn't prefer them. So we sort things like:

```
Foo
bar
```

While unix sorts things like:

```
bar
Foo
```

This difference could be avoided, but I'm not sure if it really matters so much.

nfischer:
@ariporad If you get the chance, this should be ready for review. I see that it's failing on `v0.10`, but I think this will pass on the other versions. I can investigate `v0.10` if we think it's a priority.

Feel free to review but not merge, since there's no huge rush to get these features. We can always use more input from the community. On the other hand, if we merge now, we can always go back and add features before the next release.

README.md
```diff
@@ -330,6 +330,20 @@ containing the files if more than one file is given (a new line character is
 introduced between each file). Wildcard `*` accepted.
 
 
+### head([{'-n', <num>},] file [, file ...])
+### head([{'-n', <num>},] file_array)
+
+Examples:
+
+```javascript
+var str = head({'-n, 1}, 'file*.txt');
```
ariporad: You're missing a quote her.

README.md
```diff
@@ -330,6 +330,20 @@ containing the files if more than one file is given (a new line character is
 introduced between each file). Wildcard `*` accepted.
 
 
+### head([{'-n', <num>},] file [, file ...])
+### head([{'-n', <num>},] file_array)
```
ariporad: Sorry if this is stupid, but what does head do? And maybe we should include a description here.

README.md
nfischer: I'll fix

src/sort.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+  if (options.numerical) {
+    sorted = lines.sort(function(a, b) {
+      var objA = parseNumber(a);
+      var objB = parseNumber(b);
+      if (objA.hasOwnProperty('num') && objB.hasOwnProperty('num')) {
+        return (objA.num !== objB.num) ? (objA.num - objB.num)
+            : (objA.value.localeCompare(objB.value));
+      } else {
+        return (objA.value.localeCompare(objB.value));
```
ariporad: What's localeCompare?

README.md
nfischer: I'll update the description (I just realized I forgot to do that...)

nfischer:
Looks like there's iojs compatibility issues. I'll try to fix that today and see if I can get `v0.10` working as well. I may also try to get unix semantics (I think I know how to do that).

ariporad:
A few nits, otherwise looks good.

src/sort.js
nfischer: It's the default string comparison. It returns `-1` if `this` is "less than" (sorted before) the argument.

So:

```
'foo'.localeCompare('bar'); // 1
'foo'.localeCompare('zed'); // -1
'foo'.localeCompare('foo'); // 0
```

README.md
nfischer: Should be fixed now (good catch)

README.md
nfischer: Check the description now

nfischer:
Ok, I think I fixed the inconsistency. Also, this is now compatible with how unix sort works, which is a plus. Should be good to merge now.

nfischer:
Let's hold off to merge this. I think this should be refactored to read input in via buffers (vs. reading it all at once), when possible. This is to account for the case where a file is infinite in size (or maybe just super large) and we print out the first 10 lines, keeping the operation constant time instead of hanging indefinitely.

I should have time to refactor this today. I can also add in `tail()` to the mix in another PR, for the sake of completeness (`tail` prints the last 10 lines, as opposed to `head` which prints the first 10 lines).

ariporad:
SGTM!

nfischer:
@ariporad I refactored this into 3 commits. `tail()` is a new command. This also modifies `head()` so that it only reads as much of the file as necessary (so it works for files that are infinite in length, like if you run `yes > file.txt`).

Leave any comments/concerns and I can address when I get back.

nfischer:
@ariporad This should be ready for review

src/sort.js
```diff
@@ -0,0 +1,87 @@
...
+        (objA.num - objB.num) :
+        unixCmp(objA.value, objB.value));
+  } else {
+    return unixCmp(objA.value, objB.value);
+  }
+}
+
+//@
+//@ ### sort([options,] file [, file ...])
+//@ ### sed([options,] file_array)
```
ariporad: Why is this also called `sed`? I thought `sed` was find/replace?

ariporad:
@nfischer: One comment, also can you either squash your final commit or make it comply with the commit guidelines. Thanks!

src/sort.js
nfischer: Whoops! That was a mistake. I'll fix that

nfischer:
@ariporad Rebased this and fixed the issues

nfischer:
@ariporad Let me know once you've taken a look at this. I'll address any issues and rebase. If there are no issues, feel free to give the thumbs up and I'll rebase and merge.

ariporad:
LGTM!

nfischer:
Ok great! I'll rebase this and merge

nfischer:
LGTM (because of rebase)

facundoolano:
any chance of having this published to npm?

nfischer:
Hopefully soon. We had some last minute concerns raised, but it should be resolved soon

PR #380, opened 2016-03-01
refactor(cp): clean up code and fix #376

### LINKED ISSUE #376: Copy not accepting wildcards + file ending anymore when using -r

nfischer:
Fixes issue in #376. Simplifies the code, and slight perf win.

This fixes the semantics of `cp()`, which had deviated from unix `cp` a bit. If anyone is available to review, that would be good.

Goals of this PR:

 - improve performance
 - improve correctness

### DIFF
--- a/test/set.js
+++ b/test/set.js
@@ -6,7 +6,6 @@ var oldConfigSilent = shell.config.silent;
 shell.config.silent = true;
 
 shell.rm('-rf', 'tmp');
-shell.mkdir('tmp');
 
 //
 // Valids
--- a/src/cp.js
+++ b/src/cp.js
@@ -53,8 +53,8 @@ function cpdirSyncRecursive(sourceDir, destDir, opts) {
   if (!opts) opts = {};
 
   /* Create the directory where all our junk is moving to; read the mode of the source directory and mirror it */
-  var checkDir = fs.statSync(sourceDir);
   try {
+    var checkDir = fs.statSync(sourceDir);
     fs.mkdirSync(destDir, checkDir.mode);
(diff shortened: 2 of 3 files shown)


nfischer:
@ariporad can you take a look at this? Let me know what you think, or if there's any other perf improvements we can squeeze out.

test/set.js
```diff
@@ -6,7 +6,6 @@ var oldConfigSilent = shell.config.silent;
 shell.config.silent = true;
 
 shell.rm('-rf', 'tmp');
-shell.mkdir('tmp');
```
ariporad: Why is this here? It seems completely unrelated.

ariporad:
@nfischer: One tiny nit, and could you please rebase off `master`? Otherwise LGTM.

test/set.js
nfischer: One of the tests assumed incorrect semantics of `cp()`. When I fixed the semantics, I realized this test was failing.

The directory actually gets created with a `cp()` command on [this line](https://github.com/shelljs/shelljs/blob/fix-cp-r-option/test/set.js#L20)

ariporad: Ok.

ariporad:
@nfischer: Can you rebase of master? Otherwise LGTM!!!!

nfischer:
@ariporad I just re-pushed

PR #383, opened 2016-03-05
refactor(cd): use process.env.OLDPWD to store previous dir

nfischer:
This should be a simple refactor. This uses `process.env.OLDPWD` to store the previous directory (used by `cd -`), instead of a separate variable as we had it. It's a little closer to bash's semantics if we use `process.env`, since now people can run `echo(env.OLDPWD)` and see a useful result.


ariporad:
LGTM!

PR #384, opened 2016-03-05
feature: use rechoir

nfischer:
This is a redo of #258, since it looks like it may have been abandoned.

### DIFF
--- a/test/shjs.js
+++ b/test/shjs.js
@@ -4,8 +4,8 @@ var assert = require('assert');
 
 function runScript(name) {
   // prefix with 'node ' for Windows, don't prefix for OSX/Linux
-  var cmd = (process.platform === 'win32' ? 'node' : '') + path.resolve(__dirname, '../bin/shjs');
-  var script = path.join('resources', 'shjs', name);
+  var cmd = (process.platform === 'win32' ? 'node ' : '') + path.resolve(__dirname, '../bin/shjs');
+  var script = path.resolve(__dirname, 'resources', 'shjs', name);
   return shell.exec(cmd + ' ' + script, { silent: true });
 }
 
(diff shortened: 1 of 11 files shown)


nfischer:
@TimothyGu thanks for the command. Fixed it.

If someone could please review this, that would be much appreciated.

test/shjs.js
```diff
@@ -4,10 +4,9 @@ var assert = require('assert');
 
 function runScript(name) {
   // prefix with 'node ' for Windows, don't prefix for OSX/Linux
-  return shell.exec((process.platform === 'win32' ? 'node ' : '') +
-                    path.resolve(__dirname, '../bin/shjs') +
-                    ' ' +
-                    path.resolve(__dirname, 'resources', 'shjs', name), { silent: true });
+  var cmd = (process.platform === 'win32' ? 'node ' : '') + path.resolve(__dirname, '../bin/shjs');
```
TimothyGu: I'm not sure if I have already mentioned this, but the best way to spawn another `node` process is through `process.argv[0]`. On both Unix and Windows, `node` refers to the executable in PATH, while the user might have executed the script with the absolute path to another `node` executable. `process.argv[0]` guarantees that the current used Node.js is always used.

nfischer: Yeah, that works. So does `process.execPath` I believe. I'll adjust.

nfischer:
Updated the tests to use `process.execPath` instead of "`node`" to make things more robust.

ariporad:
It looks like the tests aren't passing, otherwise looks good.

nfischer:
Yeah, I see the issue. I just need to wrap process.execPath with JSON.stringify(). Or you can take out the last commit and merge that (it's only the execPath stuff that broke tests). Away from my computer at the moment

ariporad:
LGTM, will merge once tests pass.

nfischer:
Sorry, just found another issue. Now it should be resolved.

nfischer:
Merging this, since it LGTM

PR #385, opened 2016-03-06
Re-run benchmarks after each build

ariporad:
With this PR, after every successful travis build, the latest build of shelljs/benchmarks gets re-run (which in turn runs the latest shelljs/shelljs#master commit). That's then pushed to https://shelljs.github.io/benchmarks.

This currently doesn't really work for PRs, as shelljs/benchmarks doesn't run branches.

@shelljs/contributors, @shelljs/bench: Ready for review.


ariporad:
Oh, I should mention that this only affects Travis. AppVeyor hasn't been touched.

ariporad:
@nfischer: Ready for review.

nfischer:
@ariporad could you move the `.sh` script into a folder (not at the top level)? Also, please add that to the npmignore file.

What happens if a regular user runs the shell script? I'm not very familiar with this. If, for example, we have a PR from local branch `feat-foo-bar` to `master`, does this run benchmarks once `feat-foo-bar` is pushed (and passes), once when it's made into a PR, and also once when it's merged into master? And would all of those runs of the benchmarks be on the master branch?

nfischer:
The benchmarks are still very experimental, so I'm hesitant to wire them into ShellJS just yet. I'd prefer to get a better HTML display page and a larger body of benchmarks.

nfischer:
Bumping back the milestone/priority, since this doesn't necessarily need to be in v0.7.

PR #386, opened 2016-03-06
Add cli version ...err combine work?

### LINKED ISSUE #202: CLI Version

levithomason:
I had finished this PR to close #202 as I did not see a cli that could run _commands_ in the docs.  Then I found #244 and the corresponding [shx](https://github.com/shelljs/shx) repo, whoops!

Figured I'd open this here anyhow for discussion on any possible integration into the `shx` project.  There are tests and a doc snippet.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,22 @@ run ShellJS scripts much like any shell script from the command line, i.e. witho
 $ shjs my_script
 ```
 
+If you'd like to run ShellJS commands (instead of scripts) from the command line, use the `shx` CLI.
+
+```bash
+$ shx echo "awesome!"
+``` 
+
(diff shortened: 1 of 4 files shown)


levithomason:
With shx well underway, gonna close this PR and incorporate any ideas from here over there.

nfischer:
Sounds good!

PR #387, opened 2016-03-08
feat(async): all commands have async versions

### LINKED ISSUE #2: [Feature Request] Async versions of all commands

nfischer:
Fixes #2. This is an initial stab at async-ing everything. **not ready to merge, since docs haven't been updated**.

I think this is a nice interface, since you can specify commands to be async either with:

```Javascript
cp('file1.txt', 'out1.txt'); // synchronous, like normal
cp('file2.txt', 'out2.txt', {async: true}); // asynchronous, no callback
cp('file3.txt', 'out3.txt', {async: true}, function(code, stdout, stderr) { // asynchronous, with callback
  console.log(code);
  // ...
});
cp('file4.txt', 'out4.txt', function(code, stdout, stderr) { // callback implies asynchronous
  console.log(code);
  // ...
});
```

This is consistent with the precedent set by `exec()`.

An alternative is to pass in the actual return value (usually a ShellString). This is *better* for using with pipes. I think the following is a suitable workaround, however, with this implementation:

```javascript
grep('largefile.txt', function(code, stdout, stderr) {
  var ret = ShellString(stdout, stderr);
  ret.sed(/o/g, 'a');
});
```


src/common.js
```diff
@@ -274,7 +283,24 @@ function wrap(cmd, fn, options) {
         });
         if (!config.noglob && options && typeof options.idx === 'number')
           args = args.slice(0, options.idx).concat(expand(args.slice(options.idx)));
-        retValue = fn.apply(this, args);
+        if (config.async || callback) {
+          process.nextTick(function() {
+            var r = fn.apply(this, args);
```
ariporad: This won't really work, because if the function itself blocks (ie. it uses blocking APIs), then this will too.

nfischer: Why would this not work? The function won't be called until everything on the event loop is dealt with, right?

ariporad: Yes, but imagine this scenario:

```javascript
setTimeout(() => console.log('It\'s been 1 second!'), 1000);
exec('sleep 10; echo "It\'s been 10 seconds!"', function noop() {}, { async: true });
```

Outputs:
> It's been 10 seconds.
> It's been 1 second.

src/common.js
```diff
@@ -274,7 +283,24 @@ function wrap(cmd, fn, options) {
...
+            // Convert return value to an object that makes sense
+            if (typeof r === 'boolean')
+              r = {stdout: '', stderr: '', code: (r ? 0 : 1)};
+            else if (typeof r === 'string')
+              r = {stdout: r, stderr: (state.error || ''), code: (state.error ? 1 : 0)};
+            else if (typeof r === 'undefined')
+              r = {stdout: '', stderr: (state.error || ''), code: (state.error ? 1 : 0)};
+
+            if (callback)
+              callback.call(this, r.code, r.stdout, r.stderr);
```
ariporad: Also, we should probably give the callback a `ShellString`.

nfischer: The issue is that this is inconsistent with `exec()`, which already takes a callback of this form.

With these parameters, the user can construct the equivalent `ShellString`.

src/common.js
nfischer: @ariporad I just tried this example. I saw:

> It's been 1 second!
> It's been 10 seconds!

Also, `exec()`'s async version is totally different than the rest of the commands.

Another example:

```
> setTimeout(() => console.log('It\'s been 1 second!'), 5000);
> ls({async: true}, function() { echo('hi'); });
```

Output (as I would expect, since `ls()` is super fast) :

> hi
> It's been 1 second!

ariporad: @nfischer: Whoops, my bad.

Try this (I know it's unrealistic times, but for argument's sake):

    node -e "function sleep(time) { var start = Date.now(); while (start + time > Date.now()) ; }; setTimeout(() => console.log('1 sec'), 1000); sleep(2000); console.log('2 sec');"

rektide: @ariporad your last example uses no shelljs. the actual output is what a JS programmer ought expect: you have to yield execution.

PR #393, opened 2016-03-13
Add shelljs as a keyword in package.json

nfischer:
Since we're actually not the first result when you search on npm for "shelljs", and we probably should be...

Let me know if there are any other keywords we should change.


ariporad:
LGTM!

ariporad:
@levithomason: Any other keyword ideas?

PR #394, opened 2016-03-16
feat: adding error codes to ShellJS

nfischer:
This is a redo of #269.

This adds error codes as a `.code` attribute to ShellStrings, but doesn't implement the `errorCode()` function.

This also does not switch the implementation of the `error()` function (that should be in another PR, IMO).

Tests will be added soon.

### DIFF
--- a/test/pipe.js
+++ b/test/pipe.js
@@ -11,13 +11,6 @@ shell.mkdir('tmp');
 // Invalids
 //
 
-// no piped methods for commands that don't return anything
-assert.throws(function() {
-  shell.cp('resources/file1', 'tmp/');
-  assert.ok(!shell.error());
-  shell.cp('resources/file1', 'tmp/').cat();
-});
-
 // commands like `rm` can't be on the right side of pipes
 assert.equal(typeof shell.ls('.').rm, 'undefined');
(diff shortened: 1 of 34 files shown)


nfischer:
@ariporad @levithomason This should be ready for review.

This adds the `.code` attribute to everything.

If we like this implementation and merge this, we can open a new PR to change `error()` to return the code instead of the last command's error string. I didn't include it in this PR because error codes themselves shouldn't be a breaking change, and it's easy enough to swap that implementation in a separate PR.

I left this as 2 commits, since the first one adds error codes, and the second one refactors everything to return a ShellString (which gives every command the `.code` and `.stderr` attributes). Feel free to squash if you prefer.

test/pipe.js
```diff
@@ -11,13 +11,6 @@ shell.mkdir('tmp');
 // Invalids
 //
 
-// no piped methods for commands that don't return anything
-assert.throws(function() {
```
ariporad: Why did this get removed?

nfischer: Almost every command returns a ShellString now. So the number of "commands that don't return anything" has shrunk to almost zero.

ariporad: OK.

ariporad:
LGTM!

PR #400, opened 2016-03-18
feat(glob): expose config.globOptions.

nfischer:
Allow users to customize the settings of `glob.sync()` (if they so-desire). This doesn't change the default behavior.

I imagine this is a feature a few users may want (users wanted something similar for `exec()` options in #163, #284, #132 ).

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -17,6 +17,7 @@ var config = {
   fatal: false,
   verbose: false,
   noglob: false,
+  globOptions: {},
   maxdepth: 255
 };
 exports.config = config;
@@ -179,7 +180,7 @@ exports.parseOptions = parseOptions;
 // For example:
 //   expand(['file*.js']) = ['file1.js', 'file2.js', ...]
 //   (if the files 'file1.js', 'file2.js', etc, exist in the current dir)
-function expand(list, opts) {
+function expand(list) {
   if (!Array.isArray(list)) {
     throw new TypeError('must be an array');
   }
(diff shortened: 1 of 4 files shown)


src/common.js
```diff
@@ -182,6 +183,7 @@ function expand(list, opts) {
   if (!Array.isArray(list)) {
     throw new TypeError('must be an array');
   }
+  opts = opts || config.globOptions;
```
ariporad: Should we merge these here, instead of overriding them?

ariporad:
One nit, and can we get this rebased off master? Thanks!

src/common.js
nfischer: Actually, we're not using `opts` anymore (it was added for something, but we refactored the code to not need it). I think it's cleaner to just use `config.globOptions` instead.

I don't think our code should ever need to pass options to `glob.sync()`. We experimented with using `glob.sync('dirName/*', {dot: true})` for `ls -a`, but it's much easier to just us `fs.readDirSync()` instead (it's simple, it's faster, and it actually works correctly for a directory named `dir*With?Glob[chars]/`).

nfischer:
Rebased. I changed `function expand(list, opts)` to `function expand(list)`, since we never were actually passing the `opts` argument anywhere.

ariporad:
LGTM!

ariporad:
LGTM!

nfischer:
re- LGTM

nfischer:
re-re- LGTM (rebased off master)

Issue #398, opened 2016-03-17
Edit the docs to emphasize ShellStrings and Pipes


nfischer:
The introduction of `ShellStrings` is a pretty big change, since a lot of people rely on the old behavior of strings have the `.to()` method (here's an example from our friends over at [dthree/cash](https://github.com/dthree/cash/blob/master/test/cat.js#L10-L12)).

We might want to consider adding that back in to the `String.prototype` for now, and give a deprecation warning, since this might be a big impediment to users migrating to the new version. Or, if we decide to keep it as-is, we should document it a bit more clearly, and make it obvious for how people can create a file with the syntax:

```javascript
ShellString('some text').to('file.txt');
```

Also, I think pipes are one of the more useful features in ShellJS v0.7, so these should probably be advertised better in the docs.

On a related note, it might be worth starting guides in the wiki pages about how to migrate from version to version. We've been introducing quite a few breaking changes, so it might ease the transition for users to have these breaking changes highlighted (and not only buried in the changelog with all the features/bug fixes/etc.).

dthree:
Agreed. It's kind of awkward that ShellJS is not `v1.0`, when it has been such a stable part of the Node ecosystem for so long and is in production use in a ton of applications. It should be `1.0` already, which would address these semver breaks.

nfischer:
Yeah, it's been fairly stable for a long time. Unfortunately, I think the project has had some big issues that really should be dealt with before we go to `v1.0` (like how v0.6 and earlier modify the string prototype, with no way to opt-out).

In regards to this feature specifically, I'm leaning more toward adding it back into master as a default, and figuring out some way to opt-out (since the option to opt-out has been suggested as a suitable work-around). This would give people all the convenience they have now (and ShellJS is primarily about convenience, IMO), while also offering a way to make it comply more nicely with other modules. Any suggestions from the community would be greatly appreciated.

dthree:
Okay nice. Makes sense.

When I was first playing with SHJS, I was a bit surprised on the String extension, but I figured there were just two sides to the coin. If you're using a global module, you have to expect that. 

I consider global commands (`rm`, etc.) just as "bad" as extending String, honestly, so if I didn't want any of that in a project, I wouldn't import global.

ariporad:
@nfischer: what about if in global mode, normal strings have all the methods of ShellStrings? Then we can also add a 'noConflict' method.
[...]

PR #401, opened 2016-03-19
refactor(string): modify string protoype, but only for shelljs/global

nfischer:
`require('shelljs/global')` extends String.prototype to have the `.to()` and `.toEnd()` methods again. This also adds tests for the global require.

This is along the lines of the discussion in #398. The advantage of this is that ShellJS is still backwards compatible for `require('shelljs/global')`, but we still have an option to avoid extending the String prototype (just use `require('shelljs')` instead). This would make things more convenient (and `global` is a convenience script primarily), and would ease the transition to ShellJS v0.7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ containing the files if more than one file is given (a new line character is
 introduced between each file). Wildcard `*` accepted.
 
 
-### 'string'.to(file)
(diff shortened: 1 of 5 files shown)


ariporad:
LGTM!

PR #403, opened 2016-03-20
test(rm): add tests to prevent a future regression

### LINKED ISSUE #332: rm -rf incorrect behaviour
`rm('-rf', target)` should never throw an exception when the target doesn't exist. It does, however, when it is *inside* a non-existent directory:

Expected:

```js
mkdir('./tmp');
rm('-rf', './tmp');
rm('-rf', './tmp'); // no error
rm('-rf', './tmp/*'); // no error
rm('-rf', './tmp/non/existent'); // no error
```

Actual:

```js
mkdir('./tmp');
rm('-rf', './tmp');
rm('-rf', './tmp'); // ok
rm('-rf', './tmp/*'); // rm: no such file or directory: ./tmp
rm('-rf', './tmp/non/existent'); // same
```

nfischer:
Fixes #332 

This is pretty straightforward. The issue existed on v0.6, but not on master. This just adds the tests to make sure we don't run into this issue again.

In case you're wondering, the "**only -f**" case is indeed bash-compatible. `rm -f real_dir/` is an error, `rm -f fake_dir/` is not (assuming that `fake_dir/` does not exist, of course).

Feel free to adjust the tests as necessary. I checked them against bash on Linux, and this is compatible with that.

### DIFF
--- a/test/rm.js
+++ b/test/rm.js
@@ -43,6 +43,26 @@ result = shell.rm('-f', 'asdfasdf');
 assert.equal(shell.error(), null);
 assert.equal(result.code, 0);
 
+// directory does not exist, but -fr specified
+result = shell.rm('-fr', 'fake_dir/');
+assert.equal(shell.error(), null);
+assert.equal(result.code, 0);
+
+// directory does not exist, but *only -f* specified
+result = shell.rm('-f', 'fake_dir/');
+assert.equal(shell.error(), null);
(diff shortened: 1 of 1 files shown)


ariporad:
LGTM!

ariporad:
LGTM!

nfischer:
LGTM. Merging since this is fairly trivial

Issue #406, opened 2016-03-23
Pipe example from 'Readme' doesn't work


hjek:
```
$ cat pipe.js 
require('shelljs/global')
echo('files with o\'s in the name:\n' + ls().grep('o'));

$ shjs pipe.js 
/home/pelle/pipe.js:3
echo('files with o\'s in the name:\n' + ls().grep('o'));
                                             ^
TypeError: ls(...).grep is not a function
    at Object.<anonymous> (/home/pelle/pipe.js:3:46)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Function.Module.runMain (module.js:447:10)
    at startup (node.js:141:18)
    at node.js:933:3
```

ariporad:
Hi @hjek!

I think the problem is most likely that you're using shelljs 0.6 (which is currently on npm), where as the README on master is the development version, which is 0.7. Can you confirm that this is the case?

Thanks!

hjek:
That was the problem, and the [documentation on there](https://www.npmjs.com/package/shelljs) also doesn't mention pipes.

nfischer:
The README is still the best place to check for docs, but it documents all the new features that may or may not be released yet. I'll add a warning to the top of the README to clarify this.

Issue #405, opened 2016-03-22
new release required. npm is out of date


AshHeskes:
The documentation on http://documentup.com/shelljs/shelljs is out of sync with http://documentup.com/shelljs/shelljs. Specifically the grep command options '-l' was added in https://github.com/shelljs/shelljs/commit/90569dcdac96249516999c5e7afea05b6698d2cd and npm hasn't been updated since.

ariporad:
Hi @AshHeskes,

Yes, the documentation on master doesn't exactly match what's on NPM, because master is the development version (So `grep('-l')` won't work with the current npm version!). Sorry for any confusion!

I'm going to close this, but feel free to reopen if I've misunderstood the problem.

nfischer:
Yup, a new release is on the way. A few more loose ends to tie up, and then we'll have v0.7 released to npm. Should be this week or next :+1:

PR #410, opened 2016-03-25
docs: warn that README contains newest features

nfischer:
Might be a good idea, in light of #405 and #406

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,14 @@ $ shx touch foo/bar.txt
 $ shx rm -rf foo
 ```
 
+## A quick note about the docs
+
+For documentation on all the latest features, check out our
+[README](https://github.com/shelljs/shelljs). To read docs that are consistent
+with the latest release, check out [the npm
+page](https://www.npmjs.com/package/shelljs) or
+[shelljs.org](http://documentup.com/shelljs/shelljs).
+
 ## Installing
 
 Via npm:


README.md
```diff
@@ -41,6 +41,14 @@ $ shx touch foo/bar.txt
 $ shx rm -rf foo
 ```
 
+## A quick note about the docs
+
+For documentation on all the latest features, check out our
+[README](https://github.com/shelljs/shelljs). To read docs that are consistent
+with the latest release, check out [the npm
+page](https://www.npmjs.com/package/shelljs) or
+[shelljs.org](http://documentup.com/shelljs/shelljs).
```
ariporad: Actually, we don't own shelljs.org, and documentup is generated from master.

nfischer: Shelljs.org redirects to documentup, right? And documentup is only generated when we actually generate it, which is usually once per release. Npm is pretty reliable though, and there's no way that its docs would be wrong.

ariporad: Hmm... I didn't know it did that. TIL.

ariporad:
LGTM! But could you rebase?

nfischer:
@ariporad If we switch to a `dev` branch like you suggested, it would probably cleaner to just close this. That way docs would not get updated with new features until the next release.

If we find an error with the docs (maybe something is confusing or incomplete), we can merge that into `master` sooner and rebase `dev` off of `master`.

ariporad:
SGTM

ariporad:
@nfischer: One problem with this is that we would push the notice to npm, which wouldn't make much sense.

nfischer:
@ariporad What do you mean by the latest notice? Not quite following you.

ariporad:
If we add a notice that the README is the dev version, then when we push
the README to npm, the notice will be included. Which doesn't make much
sense.


Ari

nfischer:
I see your point. This notice isn't intended to say so much as "these docs are out of date, don't trust them," but more as "regardless of whether you're at Github or npm, the latest features are at Github and the most-recent-release docs are at npm." If there's a way to better phrase that, let's do it.

ariporad:
KK. LGTM!

PR #411, opened 2016-03-27
Add forEach() method to String-based ShellStrings

### LINKED ISSUE #396: Idea: add a .forEach method to each ShellString
This would be good to add, allowing users to iterate over each the stdout from a command.

This can be done in Bash by running:

```bash
for k in $(cat package.json); do # wrap each word in <> brackets and print on a new line
  echo "<$k>"
done

# Or you can change IFS (internal field separator) to break on newlines
IFS='
'
for k in $(cat package.json); do # indent each line of the output
  echo "  $k"
done
```

A possible API would look like:

```javascript
ls().forEach(function (file) { // this already works for ls()
  echo('This is a file: ' + file);
});

// something similar could work for cat(), grep(), etc.
cat('file1.txt').forEach(function (line) {
  echo('> ' + line); // put '> ' in front of each line
});
```

Alternatives would be to iterate over each word (delimited by whitespace) (this is similar to bash's default behavior, which is sometimes useful).

We could also consider adding the ability to pass an option to `.forEach` that specifies how the output is split:

```javascript
var wordCount = 0;
cat('file.txt').forEach(function (a) { wordCount++; }, {split: 'word'});
console.log(wordCount);
```

TODO:

- [ ] Decide if this feature is worth adding
- [ ] Decide if we should allow the `{split: ...}` option
- [ ] Figure out a sensible default behavior when the `{split: ...}` option is not specified (I think "iterate over lines" and "iterate over words" are the two most sensible choices, but I'm open to other suggestions).

freitagbr:
Fixes #396 

Adds a `forEach()` method to String-based ShellStrings. This method behaves like `Array.forEach` in regular JavaScript, but it takes an optional object argument:

``` javascript
cat('file.txt').forEach(function (line) {}); // by default, splits by line
cat('file.txt').forEach(function (line) {}, { split: 'line' });
```

There are several built-in splitting methods: `'line'`, `'character'`, `'whitespace'`, and `'word'`. If no split argument is provided, the default splitting method is `'line'`.

An additional method, `'regex'`, is provided, which allows you to split by a custom expression, which can be a regex or a string:

``` javascript
cat('file.txt').forEach(function (split) {}, { split: 'regex', regex: /\n\t/g });
cat('file.txt').forEach(function (split) {}, { split: 'regex', regex: '\n\t' });
```

In this case, `split: 'regex'` is optional. Simply passing `{ regex: /\n\t/g }` is enough.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -35,6 +35,14 @@ delete process.env.OLDPWD; // initially, there's no previous directory
 var platform = os.type().match(/^Win/) ? 'win' : 'unix';
 exports.platform = platform;
 
+// regex lookup for ShellString.forEach()
+var regexLookup = {
+  'character': '',
+  'whitespace': /\s+/,
+  'word': /\W+/,
+  'line': /\n/
+};
+
 function log() {
   if (!config.silent)
     console.error.apply(console, arguments);
@@ -99,40 +107,22 @@ var ShellString = function (stdout, stderr, code) {
 
     that.forEach = function (callback, opts) {
       var options = opts || {};
+      var regex;
 
       if (typeof options !== 'object')
         options = {};
 
-      if (typeof options.split === 'undefined') {
-        // if only regex provided, set split to 'regex'
-        if (options.regex)
-          options.split = 'regex';
+      if (options.regex)
+        regex = options.regex;
 
-        // default to 'line' if no options given
-        else
-          options.split = 'line';
-      }
+      // default to line if invaild options is passed
+      else if (!(options.split in regexLookup))
+        regex = /\n/;
 
-      switch (options.split) {
-        case 'character':
-          this.stdout.split('').forEach(callback);
-          break;
-        case 'whitespace':
-          this.stdout.split(/\s+/).forEach(callback);
-          break;
-        case 'word':
-          this.stdout.split(/\W+/).forEach(callback);
-          break;
-        case 'regex':
-          this.stdout.split(options.regex || '').forEach(callback);
-          break;
-        case 'line':
-          this.stdout.split(/\n/).forEach(callback);
-          break;
-        default:
-          this.stdout.split(/\n/).forEach(callback);
-          break;
-      }
+      else
+        regex = regexLookup[options.split];
+
+      return this.stdout.split(regex).forEach(callback);
     };
   }
   that.stderr = stderr;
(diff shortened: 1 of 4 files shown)


src/common.js
```diff
@@ -96,6 +96,44 @@ var ShellString = function (stdout, stderr, code) {
...
+        case 'word':
+          this.stdout.split(/\W+/).forEach(callback);
+          break;
+        case 'regex':
+          this.stdout.split(options.regex || '').forEach(callback);
+          break;
+        case 'line':
+          this.stdout.split(/\n/).forEach(callback);
+          break;
+        default:
```
nfischer: Can we rewrite this to be:

```javascript
case 'line':
case 'default':
  this.stdout.split( // etc. ...
break;
```

src/common.js
```diff
@@ -96,6 +96,44 @@ var ShellString = function (stdout, stderr, code) {
   } else {
     that = new String(stdout);
     that.stdout = stdout;
+
+    that.forEach = function (callback, opts) {
```
nfischer: Will this override the `.forEach()` that's part of the Array type by default (I'm thinking of `ls()` here)? If so, how would this compare in performance?

src/common.js
freitagbr: We could, though JSHint complains about it.

src/common.js
freitagbr: In this case, I am adding a .forEach method to that, where that is a String(), which doesn't have a .forEach method by default. In the case of ls(), that is an Array, so I don't override the forEach method.

src/common.js
nfischer: Actually, on second thought, an alternative refactor would be to use a dictionary instead of a switch:

```javascript
var regexLookup = { // perhaps move this out of this scope to avoid having to redeclare it
  'character': '',
  'whitespace': /\s+/,
  // ...
  'line': /\n/
};

this.stdout.split(regexLookup[options.split]).forEach(callback);
```

That's up to you though, whichever approach you think is better.

src/common.js
nfischer: Ah, I see. I skipped over the if/else. SGTM

src/common.js
nfischer: > We could, though JSHint complains about it.

Ah, didn't realize JSHint would complain. Feel free to add the necessary "ignore warning" comment to avoid the error message if you'd rather do that. Either way is fine.

freitagbr: You and your lookup tables! I agree, a dictionary is a simpler way of implementing this.

PR #417, opened 2016-04-01
fix: error message now printed for fatal failures

### LINKED ISSUE #372: Error message isn't always printed
This is a bug that I think was introduced by #357. Fatal errors don't print error messages unless config.fatal is set. Non-fatal errors still get logged, as before.

```JavaScript
> ls('filedoesntexist'); // this outputs to the console
> cd('dirdoesntexist'); // this doesn't output to the console
```

nfischer:
Fixes #372 and adds a test to check for this.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -64,16 +64,15 @@ function error(msg, _code, _continue) {
   if(config.fatal)
     throw new Error(log_entry);
 
+  if (msg.length > 0)
+    log(log_entry);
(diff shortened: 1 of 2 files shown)


nfischer:
@ariporad This should be ready for review. Fixes one of the bugs blocking v0.7.

ariporad:
LGTM!

Issue #397, opened 2016-03-17
Consider changing shell.error() to return the error code


nfischer:
This is something I intentionally left out of #394, but is something we've discussed doing. Since this would be a breaking change, I thought I'd leave it here as an open issue to see if anyone has any input.

The current behavior of `shell.error()` is to return the stderr output of the last command (only if it failed) or `null` if the command succeeded.

``` javascript
shell.echo('hello world');
shell.echo(shell.error()); // this is null
shell.cd('noexist');
shell.echo(shell.error()); // outputs: cd: no such file or directory: noexist

exec('git status');
if (shell.error()) {
  // git status failed
} else {
  // git status succeeded
}
```

If we change it to output the `code` returned instead, it will have the same truthiness (commands that succeed have code 0 and commands that fail have code > 0), so the last example with the `if()` expression still works.

Because nearly all commands return shell strings, the workaround for people relying on the error message of the commands would be to use the `.stderr` attribute of the return value:

``` javascript
var result = cp('somefile.txt', 'other.txt');
echo(result.stderr); // this will work now
```

nfischer:
To continue the discussion on this, I'm going to argue to keep the current implementation of `error()`. Most people I would imagine are using it as a boolean (i.e. `if (error()) /* do this */ else /* do that */`), however some people may be using it for the string value.

Changing it would be yet another breaking change, and I'd prefer to minimize breaking changes (since we have quite a few so far). You can make the argument that the `.stderr` attribute makes it easy enough to get the last message output, but the `.code` attribute makes it just as easy to get the last error code, so it's a pretty even convenience trade-off (and I don't think that justifies breaking scripts).

I think a better alternative would be to deprecate using the string value (suggest that people only use `error()` to get a boolean, and rely on `.stderr` for the code attribute). I'm bumping this issue back to v0.8 for now, but I'll write a PR for the docs to make it into v0.7.

ariporad:
SGTM

nfischer:
Bumping to v0.9 milestone

nfischer:
Probably won't do this. Most error conditions have an error code of `1`, making it impossible to distinguish between the conditions with just the code.

PR #418, opened 2016-04-05
docs(error): deprecate relying on string value

nfischer:
This is the PR I proposed in #397. This deprecates relying on `shell.error()` to return an error string. This will give us the freedom to either keep it the same, or change it in the future, but this keeps the actual implementation the same currently.


nfischer:
@ariporad This is rebased now

ariporad:
LGTM!

PR #421, opened 2016-04-06
feat(cp): -P option, plus better handling of symlinks

### LINKED ISSUE #413: Add support for cp -P option
This is the opposite of the `-L` option. `-P` is actually not the default on bash, since it's `-L` by default for copying files, and `-P` by default for copying directories recursively.

Also, adding the `-P` option will help in case any bugs arise with `cp -L`, since we can direct people to use that until we fix them.

Edit: I originally wrote `-p`, but this should be `-P`

nfischer:
Fixes #413 

Ok, so this wound up being a bigger PR than I anticipated. I originally wanted to implement the `-P` option for `cp()`, but I realized that ShellJS didn't have very good semantics for handling links. So this PR fixes a variety of bugs surrounding behavior of links.

 - `rm('badlink')` used to not work, now it does
 - `cp('link', 'dest')` will follow the link
 - `cp('-R', 'link', 'dest')` will **not** follow the link (regardless of whether it points to a directory or not)
 - `cp('-P', 'link', 'dest')` will of course not follow the link (regardless of whether or not `-R` is used)
 - `touch('badlink')` creates the file it points to (as `touch` should)
 - `mkdir('badlink')` outputs an error message instead of crashing
 - copying a directory that has links within it works properly (used to break)


src/cp.js
```diff
@@ -185,7 +209,7 @@ function _cp(options, sources, dest) {
       return; // skip file
     }
     var srcStat = fs.statSync(src);
-    if (srcStat.isDirectory()) {
+    if ((!options.noFollowsymlink) && srcStat.isDirectory()) {
```
ariporad: Really nitty nitpick: The parens are redundant.

ariporad:
One tiny nit, otherwise looks good. Also, could you rebase this?

nfischer:
> Also, could you rebase this?

I think the new squash-merge takes care of that, right?

ariporad:
@nfischer: Alas, it does not :cry:.

nfischer:
Oh, where does it put the one PR commit then if not on top of master?

ariporad:
@nfischer: It gives you two options:

1. Rebase your commit.
2. Merge (not rebase, merge), master into your branch, them rebase master onto your branch.

nfischer:
Hmm ok. Does this look good now?

ariporad:
LGTM!

PR #422, opened 2016-04-11
Contributing guidelines

nfischer:
Opening this as a pull request, since I think it's time we start seriously discussing this.

I modified the contributing guidelines @ariporad wrote up a while ago. This has been updated due to some changes in Github, such as the "Update branch" button and the squash-merge feature. There's no need to bother to tell people to squash commits if Github is going to do it anyway for us.

The idea behind these guidelines is to be *guidelines*, not strict rules. If someone sees the link and wants to follow the contributing guidelines, that's awesome. If not, that's ok too. I don't want to be overly strict and deter contributors in the process.

I also added a Github issue template (for bug reports). The majority of issues we get are bug reports, this just helps us get the info we need sooner rather than later (OS, node version, ShellJS version, etc.). I figure if someone wants to do a feature request, they can figure out that "OS" and "ShellJS version" don't really apply, and they'll just format their Github issue in a way that makes sense.

@shelljs/contributors please give input, since I want to make sure these guidelines aren't going to deter contributions.


nfischer:
This is based off the instructions [here](https://github.com/blog/2111-issue-and-pull-request-templates)

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+
+We love to receive bug reports (we're always trying to make ShellJS more
+stable). If you've found a bug, please follow these steps:
+
+ - Search for any issues that may have been created already. We often receive
+   duplicates, and cutting down on this is helpful. If someone else has already
+   reported it, please ping that issue thread.
+ - Let us know your version of NodeJS (`node -v`), your version of ShellJS (from
+   `package.json`), your OS, and any other useful information.
+ - Give an example ShellJS command to reproduce the error.
```
ariporad: I don't know how we would say this, but we get a lot of issues that say "This script hangs/has no output/exits X", but then they can't or wont tell us what the script actually is, and it would be nice to cut down on that.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
+# Contributing
```
ariporad: I think we should move this file out of `.github`, into the root of the project, because some people (at least I do), go looking for a `CONTRIBUTING.md`, but wouldn't think to look in `.github` (Github added that feature quite recently).

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+ - Let us know your version of NodeJS (`node -v`), your version of ShellJS (from
+   `package.json`), your OS, and any other useful information.
+ - Give an example ShellJS command to reproduce the error.
+
+## Pull requests
+
+PRs are welcome! However, we ask that you follow a few guidelines:
+
+ - Please add tests for all changes/new features.
+ - Make sure your code passes `npm test`. Please check the CI (both Appveyor and
```
ariporad: Maybe add something about using AppVeyor (and/or Travis) to test thing if you don't have access to a specific OS.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
+Node version:
+
+ShellJS version (the most recent version you see the bug on):
+
+Operating system:
+
+Description of the bug:
+
+Example ShellJS command to reproduce the error:
+
```
ariporad: Maybe change this to:

    Example ShellJS command to reproduce the error:
    
    ```javascript
    
    // everything.is(broken)
    
    ```

.github/CONTRIBUTING.md
nfischer: I said "Example ShellJS command to reproduce the error:" Pretty sure that'll do the trick. Let me know if you think otherwise.

ariporad: I phrased that poorly.

A lot of people will give us something like this:

```javascript
exec('command', function(...) {
    // Something is broken here
});
```

But they wont or can't tell us what `command` is, so we can't debug it. So maybe add something like `Please include the full commands of any ``exec``'s you use`

.github/CONTRIBUTING.md
nfischer: Most proposed changes are simple enough that they work on both Windows and Unix. If someone submits something that breaks on one OS, we can just mention that if they don't figure it out themselves (and are actually willing to debug that way).

.github/ISSUE_TEMPLATE.md
nfischer: Good idea

.github/CONTRIBUTING.md
nfischer: sure

nfischer:
@ariporad comments have been addressed

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+   `package.json`), your OS, and any other useful information.
+ - Give an example ShellJS command to reproduce the error.
+
+## Pull requests
+
+PRs are welcome! However, we ask that you follow a few guidelines:
+
+ - Please add tests for all changes/new features.
+ - Make sure your code passes `npm test`. Please check the CI (both Appveyor and
+   Travis) to make sure it passes). If you can't figure out why something
```
ariporad: There's an extra paren here.

nfischer: Done

ariporad:
LGTM!

Issue #420, opened 2016-04-06
ShellString's stdout property is inconsistent


scott113341:
Hello!  First of all, ShellJS is awesome, and I'm really excited about the project!

I was messing around with [shx](https://github.com/shelljs/shx) a little and noticed some inconsistent behavior with the `ShellString` class:

* When a `ShellString` object is created with a `String`, the resulting `ShellString`'s `stdout` property is set to that string. [source](https://github.com/shelljs/shelljs/blob/193efa7048d03db184ae3c27566b3a5f54699fdd/src/common.js#L97)
* When a `ShellString` object is created with an `Array`, the resulting `ShellString`'s `stdout` property is set to a newline-joined string of that `Array` **with an extra newline character appended**. [source](https://github.com/shelljs/shelljs/blob/193efa7048d03db184ae3c27566b3a5f54699fdd/src/common.js#L94)

Thank you, and please let me know if you'd like me to make a PR with additional `ShellString` tests.

```
shelljs|master ⇒ git status
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working directory clean
shelljs|master ⇒ git rev-parse HEAD
193efa7048d03db184ae3c27566b3a5f54699fdd
shelljs|master ⇒ node -v
v5.10.0
shelljs|master ⇒ npm -v
3.8.3
shelljs|master ⇒ node
> var shell = require('./shell.js');
undefined
> var ShellString = shell.ShellString;
undefined
> var output = new ShellString('meow');
undefined
> output.stdout;
'meow'
> output = new ShellString(['cat', 'says', 'meow']);
[ 'cat',
  'says',
  'meow',
  stdout: 'cat\nsays\nmeow\n',
  stderr: undefined,
  code: undefined,
  to: [Function],
  toEnd: [Function],
  cat: [Function],
  head: [Function],
  sed: [Function],
  sort: [Function],
  tail: [Function],
  grep: [Function],
  exec: [Function] ]
> output.stdout;
'cat\nsays\nmeow\n'
> 
```

ariporad:
Hi @scott113341,

That behavior is actually intentional. `ShellString` is a bit of a misnomer, `ShellResult` would probably be better. So the `ShellString` itself is whatever the result of the command is (currently this is a String or an Array, depending on the command). However, `shellstring.stdout` is _guaranteed_ to be a string, and it should be equivalent to the stdout of the equivalent unix command.

For example:

```javascript
[...]

PR #425, opened 2016-04-14
fix(ls): no trailing newline for empty directories

nfischer:
`ls('empty')` used to have a `.stdout` property of `"\n"`, which is wrong. This fixes that. Trailing newlines are only appended for arrays that have length \> 0.

Once this gets fixed, almost all newline output will be consistent for the various shelljs commands. This is a partial fix for #420 (`pwd()`, `echo()`, and `which()` still need to be fixed).


ariporad:
LGTM!

PR #428, opened 2016-04-20
fix: null is no longer confused for an object

nfischer:
This is an issue in the case of `echo(which('fakecmd'))`. This used to succeed
in v0.6 but was broken during the ShellString refactor on the master branch.


ariporad:
LGTM!

Issue #340, opened 2016-02-04
Asking for feedback on the `make` command


ariporad:
Hi Everyone!

Me and @nfischer (but mostly me), were wondering if you guys used `shelljs/make`, and/or if you'd be sad if it went away.

Thoughts?

nfischer:
This is a feature I don't personally use. I don't mind keeping it if there are people out there that use it, although I see very few bug reports/feature requests for the `make` command, which makes me suspect it's less popular than the rest of the API.

ariporad:
@nfischer: My thoughts exactly.

zephraph:
I don't really get why this feature is being deprecated. Was it a maintenance burden? You yourselves noted that there weren't many issues coming in on this feature, so why remove it?

Here are at least two examples of where it's used:
- https://github.com/box/t3js/blob/master/Makefile.js
- https://github.com/eslint/eslint/blob/master/Makefile.js

What's the rational behind this?

nfischer:
Given that we've already had a couple comments on this, and some prominent projects use this feature, I don't see much benefit to moving forward with removing this. We may want to consider reverting #431.

nfischer:
I'm going to close this, since there's no reason as of yet to remove this feature.

nfischer:
I think I forgot to publicize this here, but the documentation for `shelljs/make` moved to https://github.com/shelljs/shelljs/wiki/The-make-utility.

PR #431, opened 2016-04-26
chore(make): depreciate shelljs/make

ariporad:


make.js
```diff
@@ -1,5 +1,8 @@
 require('./global');
 
+console.log('WARNING: shelljs/make is deprecated, and will be going away soon. You should upgrade ' +
+            'your code to stop using it. You should also submit feedback at https://git.io/vwgId.');
```
nfischer: Could we use the full URL? The shortened URL may appear as less trustworthy.

nfischer:
@ariporad I updated the warning output to be a little more concise. I took out the URL, since I'm sure if someone has a concern, they can just post an issue and we can redirect the conversation to the original thread.

ariporad:
LGTM!

timburgess:
Why is make functionality being deprecated?

nfischer:
@timburgess Please take a look at #340 and we'll continue the conversation there

PR #433, opened 2016-04-26
chore: test against node v6

nfischer:
Looks like node v6 just got released. Just tried this out via nvm and saw that one of our tests broke for node v6, so I fixed that test.

Not sure if appveyor can test against node v6 yet, but I added it to the mix anyway.

### DIFF
--- a/test/pwd.js
+++ b/test/pwd.js
@@ -23,6 +23,6 @@ var _pwd = shell.pwd();
 assert.equal(_pwd.code, 0);
 assert.ok(!_pwd.stderr);
 assert.equal(shell.error(), null);
-assert.equal(path.basename(_pwd), 'tmp');
+assert.equal(path.basename(_pwd.toString()), 'tmp');
 
 shell.exit(123);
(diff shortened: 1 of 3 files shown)


nfischer:
Update: I removed node v6 from appveyor, since it doesn't seem to be supported yet.

ariporad:
Oops... I didn't see this one when opening #434. I'll close that one.

ariporad:
Oh, I could review this, couldn't I?

test/pwd.js
```diff
@@ -23,6 +23,6 @@ var _pwd = shell.pwd();
 assert.equal(_pwd.code, 0);
 assert.ok(!_pwd.stderr);
 assert.equal(shell.error(), null);
-assert.equal(path.basename(_pwd), 'tmp');
+assert.equal(path.basename(_pwd.toString()), 'tmp');
```
ariporad: ? Why is this here? Seems unrelated.

nfischer:
Yup, feel free to merge if this is good

test/pwd.js
nfischer: It fails without this. In node v6, I believe `basename()` requires a regular string. If you can get it to pass without this, feel free to change as necessary

ariporad: Kk

ariporad:
LGTM.

Being real smart today I am.

PR #437, opened 2016-05-01
docs: comment code better to help contributors

nfischer:
I realized that our code base can be somewhat confusing in key areas. This should help clarify some of the more confusing parts for other contributors.

### DIFF
--- a/scripts/generate-docs.js
+++ b/scripts/generate-docs.js
@@ -7,11 +7,12 @@ echo('Appending docs to README.md');
 cd(__dirname + '/..');
 
 // Extract docs from shell.js
-var docs = grep('//@', 'shell.js');
+var docs = grep('^//@', 'shell.js');
 
+// Now extract docs from the appropriate src/*.js files
 docs = docs.replace(/\/\/\@include (.+)/g, function(match, path) {
   var file = path.match('.js$') ? path : path+'.js';
-  return grep('//@', file);
+  return grep('^//@', file);
 });
 
 // Remove '//@'
--- a/scripts/run-tests.js
+++ b/scripts/run-tests.js
@@ -21,14 +21,14 @@ var jsfiles = common.expand([pwd() + '/*.js',
                              pwd() + '/src/*.js',
                              pwd() + '/test/*.js'
                             ]).join(' ');
+// Perform linting on all javascript files
 if (exec(JSON.stringify(process.execPath)+' '+pwd()+'/'+JSHINT_BIN+' '+jsfiles).code !== 0) {
   failed = true;
   echo('*** JSHINT FAILED! (return code != 0)');
-  echo();
 } else {
   echo('All JSHint tests passed');
-  echo();
 }
+echo();
 
 //
 // Unit tests
@@ -43,13 +43,13 @@ ls('*.js').forEach(function(file) {
   }
 });
 
+echo();
+
 if (failed) {
-  echo();
   echo('*******************************************************');
   echo('WARNING: Some tests did not pass!');
   echo('*******************************************************');
   exit(1);
 } else {
-  echo();
   echo('All tests passed.');
 }
--- a/shell.js
+++ b/shell.js
@@ -87,16 +87,16 @@ exports.tail = common.wrap('tail', _tail, {idx: 1});
 //@include ./src/toEnd
 
 //@include ./src/sed
-var _sed = require('./src/sed'); // don't glob-expand regexes
-exports.sed = common.wrap('sed', _sed, {idx: 3});
+var _sed = require('./src/sed');
+exports.sed = common.wrap('sed', _sed, {idx: 3}); // don't glob-expand regexes
 
 //@include ./src/sort
 var _sort = require('./src/sort');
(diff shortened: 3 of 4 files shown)


scripts/generate-docs.js
```diff
@@ -7,11 +7,12 @@ echo('Appending docs to README.md');
 cd(__dirname + '/..');
 
 // Extract docs from shell.js
-var docs = grep('//@', 'shell.js');
+var docs = grep('^//@', 'shell.js');
```
nfischer: This makes the regex a bit more specific than what it was before. This lets us have comments like what we see in shell.js line 30:

```javascript
// The //@include includes the docs for that command
```

Otherwise, this would attempt to include docs from a file named `includes the docs for that command.js` (which obviously doesn't exist).

scripts/run-tests.js
```diff
@@ -21,14 +21,14 @@ var jsfiles = common.expand([pwd() + '/*.js',
...
+// Perform linting on all javascript files
 if (exec(JSON.stringify(process.execPath)+' '+pwd()+'/'+JSHINT_BIN+' '+jsfiles).code !== 0) {
   failed = true;
   echo('*** JSHINT FAILED! (return code != 0)');
-  echo();
 } else {
   echo('All JSHint tests passed');
-  echo();
 }
+echo();
```
nfischer: Just moving this `echo()` call out of the if-else, since we do it for either condition.

shell.js
```diff
@@ -18,6 +18,21 @@ var common = require('./src/common');
...
+// var functionName = require('./src/fileName');
+// exports.nameOfCommand = common.wrap(nameOfCommand, functionName, {idx: firstIndexToExpand});
+// ```
+//
+// The //@include includes the docs for that command
+//
+// firstIndexToExpand should usually be 1 (so, put {idx: 1})
+// Increase this value if the command takes arguments that shouldn't be expanded
+// with wildcards, such as with the regexes for sed & grep
+
```
nfischer: This is just an explanation of the boilerplate code. In particular, this explains what `{idx: 1}` is all about (which is not very intuitive otherwise).

src/common.js
```diff
@@ -259,10 +259,9 @@ function wrap(cmd, fn, options) {
     try {
       var args = [].slice.call(arguments, 0);
 
+      // Log the command to stderr, if appropriate
       if (config.verbose) {
-        args.unshift(cmd);
-        console.error.apply(console, args);
-        args.shift();
+        console.error.apply(console, [cmd].concat(args));
```
nfischer: I realized this could be a one-liner, so I went ahead and made the change. This doesn't affect the behavior.

nfischer:
@ariporad Let me know if anything here should be changed/updated

shell.js
```diff
@@ -72,15 +87,15 @@ exports.tail = common.wrap('tail', _tail, {idx: 1});
 //@include ./src/toEnd
 
 //@include ./src/sed
-var _sed = require('./src/sed');
+var _sed = require('./src/sed'); // don't glob-expand regexes
```
ariporad: Shouldn't the `// don't glob-expand regexes` be on the line below?

src/common.js
```diff
@@ -259,21 +261,24 @@ function wrap(cmd, fn, options) {
...
       }
 
-      if (options && options.notUnix) {
+      if (options && options.notUnix) { // this branch is for exec()
         retValue = fn.apply(this, args);
-      } else {
+      } else { // and this branch is for everything else
         if (args[0] instanceof Object && args[0].constructor.name === 'Object') {
-          args = args; // object count as options
+          args = args; // a no-op, allowing the syntax `touch({'-r': file}, ...)`
```
ariporad: I think we should just remove this bit, it's useless.

shell.js
nfischer: Good point. I think I put it there just because the line was getting long, but I agree it makes more sense as you suggested.

src/common.js
nfischer: I agree. I think I wrote it like that because some old code had that, but it's kind of silly. I just trimmed out the `args = args` part, but left the `if()` as-is, since it makes the boolean logic simpler to read I think. If you have any suggestions on trimming down the logic, that would be even better, since it gets pretty hairy here.

ariporad: Ok, I'd say that we should do something like (instead of `args = args;`, `// pass` or `// noop`, to indicate that nothing's happening).

src/common.js
```diff
@@ -259,21 +261,24 @@ function wrap(cmd, fn, options) {
...
       }
 
-      if (options && options.notUnix) {
+      if (options && options.notUnix) { // this branch is for exec()
         retValue = fn.apply(this, args);
-      } else {
+      } else { // and this branch is for everything else
         if (args[0] instanceof Object && args[0].constructor.name === 'Object') {
-          args = args; // object count as options
+          // a no-op, allowing the syntax `touch({'-r': file}, ...)`
```
nfischer: @ariporad is this different from what you had in mind?

ariporad: Nope, looks great.

ariporad:
LGTM

PR #443, opened 2016-05-16
chore: setup changelog

levithomason:
This PR adds the same changelog flow added to `shx`.  Also generates the first changelog.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{md,markdown}]
+trim_trailing_whitespace = false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,494 @@
+# Change Log
+
+## [Unreleased](https://github.com/shelljs/shelljs/tree/HEAD)
+
+[Full Changelog](https://github.com/shelljs/shelljs/compare/v0.7.0...HEAD)
+
+**Closed issues:**
+
+- Stdout is empty on Git log command [\#439](https://github.com/shelljs/shelljs/issues/439)
+- Cannot read toString of null when using execSync [\#415](https://github.com/shelljs/shelljs/issues/415)
+- Adding callback to basic commands [\#102](https://github.com/shelljs/shelljs/issues/102)
+
+**Merged pull requests:**
+
+- docs: comment code better to help contributors [\#437](https://github.com/shelljs/shelljs/pull/437) ([nfischer](https://github.com/nfischer))
+- chore\(CI\): update appveyor [\#436](https://github.com/shelljs/shelljs/pull/436) ([nfischer](https://github.com/nfischer))
+- chore: test against node v6 [\#433](https://github.com/shelljs/shelljs/pull/433) ([nfischer](https://github.com/nfischer))
+- chore\(make\): depreciate shelljs/make [\#431](https://github.com/shelljs/shelljs/pull/431) ([ariporad](https://github.com/ariporad))
+- docs: warn that README contains newest features [\#410](https://github.com/shelljs/shelljs/pull/410) ([nfischer](https://github.com/nfischer))
+
+## [v0.7.0](https://github.com/shelljs/shelljs/tree/v0.7.0) (2016-04-25)
+[Full Changelog](https://github.com/shelljs/shelljs/compare/v0.6.0...v0.7.0)
+
+**Closed issues:**
+
+- exec\('nohup node some.js &'\) [\#426](https://github.com/shelljs/shelljs/issues/426)
+- cp copy to symlinked folder [\#414](https://github.com/shelljs/shelljs/issues/414)
+- Invalid version number \(0.0.1alpha1\) [\#399](https://github.com/shelljs/shelljs/issues/399)
+- shelljs Breaks SemVer for Alpha and Pre-Release Versions [\#390](https://github.com/shelljs/shelljs/issues/390)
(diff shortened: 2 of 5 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{md,markdown}]
+trim_trailing_whitespace = false
```
levithomason: Standardized editor config options.  Added as markdown files use two spaces at the end of a line for a line break, but in all other files, you want whitespace trimmed.

CHANGELOG.md
```diff
@@ -0,0 +1,494 @@
+# Change Log
```
levithomason: Initial changelog.

RELEASE.md
```diff
@@ -1,13 +1,20 @@
...
-* Bump version in package.json. Any breaking change or new feature should bump
-  minor (or even major). Non-breaking changes or fixes can just bump patch.
-* Update README manually if the changes are not documented in-code. Run
+1. Ensure master passes CI tests
+1. Bump version:
+  - `$ npm version <major|minor|patch>`
+
+    >`major` - breaking API changes  
+    >`minor` - backwards-compatible features  
+    >`patch` - backwards-compatible bug fixes
```
levithomason: Removed previous manual package bumping and committing instructions for `npm version` command.  It bumps package.json, creates tags, and bump commit.

Also, the version suggestions were ambiguous and not semver.org compliant.  Updated to match the [semver](semver.org) specs.

package.json
```diff
@@ -23,7 +23,8 @@
   "homepage": "http://github.com/shelljs/shelljs",
   "main": "./shell.js",
   "scripts": {
-    "test": "node scripts/run-tests"
+    "test": "node scripts/run-tests",
+    "changelog": "bash scripts/changelog.sh"
```
levithomason: I think it is helpful to standardize all scripts to npm scripts.

scripts/changelog.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
levithomason: Again, copied from shx here.  As we round this out, I'll likely release an npm module for this so we don't have to update many scripts.

levithomason:
This PR brings up a larger point, v1.0.0.  Technically, when used in production you are at 1.0.  I think we're there and should release a 1.0 as the next release.

The biggest advantage here is that breaking changes are properly noted by the major version number.  I'm updating some projects with shelljs right now and I'm not sure what 0.6 to 0.7 may have broke or if it did.

Lastly, we could take a note from facebook/reactjs.  They recently went from `v0.14.0` to `v15.0.0` on the next release.  This was because reactjs is in production use all over the place, including facebook.

scripts/changelog.sh
nfischer: @levithomason I completely agree with making this an npm module (this is what ShellJS is for, anyway). I actually ready wrote out a release.js script (translated from your release.sh). If that's fine with you, I can publish that under the ShellJS org (shouldn't be an issue, since it's a derivative work under the MIT license).

levithomason: Absolutely.

levithomason: Actually, let me clarify.  I was referring to an npm module for my changelog api, are you thinking of a release script for npm?

nfischer: I wrote one out for `release.sh` at shx to begin with. We could also do one for the changelog. (let me know if you would prefer to not have `release.sh` transpiled to JS).

I can work on translating `changelong.sh` as well or instead.

levithomason: Got it.  Yea, I think there are two scripts here.  Starting with bash > shelljs transpiled versions sounds like a good start.  I can see the shelljs release as a shelljs org npm package.  You're clear by me if you'd like to get that out the door.  If you're gonna do it right away, feel free to update this PR as well.

At some point, I'll likely pull out the generate changelog functionality of said release script and publish my own module.  Just so it is easy to generate a changelog for any project.

nfischer: @levithomason Here's an initial version, if you'd like to take a look. I believe you and @ariporad should automatically be collaborators, so feel free to make changes.

https://github.com/shelljs/release/blob/master/index.js

I also have a similar one for changelog.js if you'd like to base your module off of that. It still relies on `curl`, however.

levithomason: `echo("  Usage: bash $0 <major|minor|patch>");` should be `node`.  Other than that, the once over looks good.

nfischer: Whoops, missed that one. Will fix.

ariporad:
LGTM.

With regards to v1.0.0, semver defines v0 as a beta, so that unstable packages can be developed easily.

I personally think that we need to start progressing to 1.0.0, but I have a few things I want to work out before doing so, I'll create a milestone when I'm off mobile.

ariporad:
Big picture ShellJS discussion in #444.

levithomason:
@nfischer we good on this PR then?  Assuming you'll make script updates on another PR.

nfischer:
LGTM, but why has this not passed CI?

levithomason:
It did pass on my last commit.  But the auto changelog commit `[ci skip]` seems to put the test in permanent suspense?

image

nfischer:
If that's the case, we should probably wait to merge until we can figure that out (don't want to accidentally freeze our CI)

levithomason:
I just mean only that commit is not reporting for some reason.  Tests are still running, and passing as seen here: https://travis-ci.org/shelljs/shelljs/builds/130655225

I'll try to remove that commit and re-add another one to see if that works.

levithomason:
If this doesn't work, we can actually remove the [ ci skip ] for our setup.

levithomason:
OK, removed the ci skip so the statuses show through on the PR.  Tests were still running, but they never reported back.  Suppose the extra test for changelog-only commits doesn't hurt.

Just need another sign off now.

PR #449, opened 2016-05-27
docs(exec): explicitly mention the `shell` option

nfischer:
Since we had a couple issues about this option.

If you can think of a way to phrase this more clearly and concisely, please go ahead and do so.


ariporad:
LGTM

Issue #446, opened 2016-05-26
ls() not returning an Array


jcbohin:
### Node version:

v6.2.0

### ShellJS version (the most recent version/Github branch you see the bug on):

0.7.0

### Operating system:

OSX 15.5.0

### Description of the bug:

`shelljs.ls()` returns an object containing results (files and folders as elements of the array) and a bunch of methods and no longer an Array (as it used to do until 0.7 and as described in [the docs](https://github.com/shelljs/shelljs#lsoptions-path_array) : 

> Returns array of files in the given path, or in current directory if no path provided.

### Example ShellJS command to reproduce the error:

```javascript
require('shelljs').ls('/')
```

nfischer:
It returns an array with some extra attributes. Does this break your code?

jcbohin:
Well, I have some issues with the attributes that don't act as expected (especially for an Array), like `sort()` for example:

with shelljs@0.6.0:
```
> shelljs.ls('/tmp/').sort()
[ 'KSOutOfProcessFetcher.501.ppfIhqX0vjaTSb8AJYobDV7Cu68=',
  'VSCode Crashes',
  'com.apple.launchd.QWh3jiFEDS',
  'com.apple.launchd.TUty48UNRv',
  'com.apple.launchd.TX6hPuCu3V',
  'ct.shutdown',
  'test',
  'test-shelljs',
  'test2' ]
>
```

with shelljs@0.7.0:
```
> shelljs.ls('/tmp/').sort()
{ [String: 'com.apple.launchd.QWh3jiFEDS\ncom.apple.launchd.TUty48UNRv\ncom.apple.launchd.TX6hPuCu3V\nct.shutdown\nKSOutOfProcessFetcher.501.ppfIhqX0vjaTSb8AJYobDV7Cu68=\ntest\ntest-shelljs\ntest2\nVSCode Crashes\n']
  stdout: 'com.apple.launchd.QWh3jiFEDS\ncom.apple.launchd.TUty48UNRv\ncom.apple.launchd.TX6hPuCu3V\nct.shutdown\nKSOutOfProcessFetcher.501.ppfIhqX0vjaTSb8AJYobDV7Cu68=\ntest\ntest-shelljs\ntest2\nVSCode Crashes\n',
  stderr: null,
  code: 0,
  to: [Function],
  toEnd: [Function],
  cat: [Function],
  head: [Function],
  sed: [Function],
  sort: [Function],
  tail: [Function],
  grep: [Function],
[...]

PR #450, opened 2016-05-27
test(ls): add case for trailing slash on dir name

nfischer:
In response to [this comment](https://github.com/shelljs/shelljs/issues/446#issuecomment-222088271). It appears this behavior doesn't work correctly on OS X 15.5 (but works on ubuntu and windows).


ariporad:
LGTM!

Issue #391, opened 2016-03-13
Plugin System.


levithomason:
**[NOTE: This thread was originally about the `open` command. I'm hijacking it. -@ariporad]**

It would be nice to have a plugin system for shelljs which allows custom commands.

nfischer:
@levithomason Thanks for the suggestion!

Unfortunately, `open` seems to have very different semantics across different systems. `open` on OS X will do as you described, while `xdg-open` on a Linux system does the equivalent. [This stackoverflow post](http://stackoverflow.com/questions/264395/linux-equivalent-of-the-mac-os-x-open-command) discusses it.

I can't remember off the top of my head what `open` does on Linux, but I know there is such a command, so I'm hesitant to implement an `open()` function that differs from that in semantics. Maybe if we could find a different (non-ambiguous) name? Let me know if you have any suggestions.

levithomason:
It seems the common denominator (except Windows) to opening a file with the default program is `open()`.

```
$ gnome-open file.txt   // Linux
$ xdg-open file.txt     // Linux
$ gvs-open file.txt     // Linux
$ open file.txt         // Mac
$ file.txt              // Windows
```
Ref [Windows](http://superuser.com/questions/246825/open-file-from-the-command-line-on-windows) and [Linux](http://askubuntu.com/questions/101965/what-is-the-replacement-for-gnome-open-in-gnome).

Though, [`$ open` on Linux](http://man7.org/linux/man-pages/man2/open.2.html) seems to open a file descriptor but not the default program?

nfischer:
Yeah, `open` has pretty different behavior between mac and Linux.

I think this would perhaps be better done as a separate module that plugs in. We could do a `shelljs-open` module which would call `xdg-open` for Linux (that's a shell script available on pretty much any distro), `open` on Mac, and just the filename on Windows. Wouldn't be hard to implement, and we could probably find a clever way to make it feel like it's part of the shelljs package itself when both are included.

@levithomason What do you think?

ariporad:
So, I'd actually like to hijack this thread as a way of discussing custom shelljs commands.
Currently, this is what we suggest:
```javascript var git = require('git'); var open = require('@shelljs/open'); var shell = require('shelljs');
shell.cd('foo/bar/baz'); git.checkout('branch'); open('foo.txt'); ```
[...]

PR #451, opened 2016-05-29
Plugins

ariporad:
This is a very, very rough first draft of an implementation of #391. 

Currently, I've written the code that will handle all the plugins, and started porting stuff from `common.js` to it.

Some things about this PR:
- It is written in ES6. Since the change to a plugin design will be a huge refactor, I figure this is a good chance to switch to babel. (@levithomason, what are your thoughts on setting up a build system, similar to shx, for this repo?)
- It has no tests. It's very much a rough idea, some things won't work as some parts haven't been integrated.
- I have `// XXX:` comments next to everything I'm unsure about, bikeshedding is requested.

So, @levithomason, @nfischer: What do you think?


.editorconfig
```diff
@@ -0,0 +1,10 @@
+[*]
```
ariporad: My editor was complaining about this not being here, so I went ahead and added it. I'm happy to move it to another PR  if needed.

plugin.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+
+const PLUGIN_ID_KEY = '___shelljs_plugin_id___';
```
ariporad: Let me explain about the whole plugin id thing:

There are two versions of every command: The one that's actually exposed as a property of `$`, and the internal representation, which has some other metadata and state.

This is a cheap and easy way to link them. The thing that's actually exposed gets a non-enumerable, non-writable, non-configurable property called `___shelljs_plugin_id___`, which is set to a number. Then, we have an array of internal plugin representations [here](#diff-18e559ebd3cc204a52bc1dc3db2adf8bR17), where `plugins[i].id === i`.

plugin.js
```diff
@@ -0,0 +1,112 @@
...
+    value: id
+  });
+  return id;
+}
+
+const getPluginID = plugin => plugin[PLUGIN_ID_KEY];
+
+module.exports = function setupShellPlugins($) {
+  const used_plugins = [];
+  const plugins = [{ // $
```
ariporad: Plugin ID #0 is always the initial shelljs object, and as such has to be created and setup here, with some values hardcoded.

plugin.js
```diff
@@ -0,0 +1,112 @@
...
+    silent: false,
+    fatal: false,
+    verbose: false,
+    noglob: false,
+    globOptions: false,
+    maxdepth: false,
+  };
+
+
+  const utils = {
```
ariporad: Most of what was in common.expand will now go here, which is an object of utilities passed into every plugin. It's referred to in documentation and comments as `plugin.utils`.

plugin.js
```diff
@@ -0,0 +1,112 @@
...
+    platform: os.type().match(/^Win/) ? 'win' : 'unix',
+    plugin(target, fullname, func) {
```
ariporad: This function is actually what creates a new shelljs command. It would be invoked like this:

```javascript
$.use(($, utils) => {
  var test = utils.plugin($, 'test', msg => console.log('$.test: ' + msg));
  utils.plugin(test, 'testing', msg => console.log('$.test.testing: ' + msg));
  utils.plugin($.test, 'testing2', msg => console.log('$.test.testing2:' + msg));
  $.test === test; // true
});

// You can also extend existing commands.

$.use(($, utils) => {
    utils.plugin($.test, 'test.testing3', msg => console.log('$.test.testing3'));
});

$.test('foo!'); // $.test: foo!
$.test.testing('bar!'); // $.test.testing: bar!
$.test.testing2('baz!'); // $.test.testing2: baz!
$.test.testing3('quz!'); // $.test.testing3: quz!
```

There's currently no way to create commands that can't be invoked (ex. you can't just call `$.git()`). I'm thinking the syntax might be `utils.plugin($, 'whatever');`

src/common.js
```diff
@@ -30,8 +31,10 @@ var state = {
 };
 exports.state = state;
 
+// XXX: This needs a new home
 delete process.env.OLDPWD; // initially, there's no previous directory
```
ariporad: Unsure if we need this, but I feel like it should probably go in `shell.js`.

src/common.js
```diff
@@ -86,7 +89,8 @@ exports.error = error;
 //@ ```
 //@
 //@ Turns a regular string into a string-like object similar to what each
-//@ command returns. This has special methods, like `.to()` and `.toEnd()`
+//@ command returns. This has special methods, like `.to()` and `.toEnd()
+// XXX: Is this used by anything other than `wrap()`? Does it need to go on plugin.utils?
```
ariporad: So, I'm undecided where to put this.

First of all, does `wrap` auto convert the return value to a shell string?

My gut feeling is that this should be in it's own file (but not it's own plugin), and available as `$.ShellString`

src/common.js
```diff
@@ -111,6 +115,7 @@ exports.ShellString = ShellString;
 
 // Return the home directory in a platform-agnostic way, with consideration for
 // older versions of node
+// XXX: I don't think this should be a part of plugin.utils. Not sure what should happen to it. Maybe just inline it.
```
ariporad: I'm not sure where this is used, or why it's needed. I'm thinking we should maybe just use an external ponyfill. (I like [this one](https://github.com/sindresorhus/home-or-tmp), because it works if the home dir doesn't exist. If we don't want that, [this one](https://www.npmjs.com/package/os-homedir) looks good, as does [this one](https://www.npmjs.com/package/real-homedir).)

src/common.js
```diff
@@ -125,6 +130,7 @@ exports.getUserHome = getUserHome;
 //   parseOptions('-a', {'a':'alice', 'b':'bob'});
 // Returns {'reference': 'string-value', 'bob': false} when passed two dictionaries of the form:
 //   parseOptions({'-r': 'string-value'}, {'r':'reference', 'b':'bob'});
+// XXX: Is it sufficient to have this as part of wrap(), or should it be part of plugin.utils? (ie. do plugins need access to it directly?)
 function parseOptions(opt, map) {
```
ariporad: Unsure if plugins need access to this directly, or having it as part of `wrap` is sufficient.

src/common.js
```diff
@@ -201,6 +207,7 @@ exports.expand = expand;
 
 // Normalizes _unlinkSync() across platforms to match Unix behavior, i.e.
 // file can be unlinked even if it's read-only, see https://github.com/joyent/node/issues/3006
+// XXX: I'm not really sure why this is here or if it's still needed. I think this probobly isn't going to be a part of plugin.utils.
 function unlinkSync(file) {
```
ariporad: No clue what's going on here.

src/common.js
```diff
@@ -217,6 +224,7 @@ function unlinkSync(file) {
 exports.unlinkSync = unlinkSync;
 
 // e.g. 'shelljs_a5f185d0443ca...'
+// XXX: This isn't going to be a part of plugin.utils, it should be replaced by mktmp
```
ariporad: I'll submit a PR replacing this with a proper `mktemp` command.

src/common.js
```diff
@@ -236,6 +244,7 @@ exports.randomFileName = randomFileName;
 // extend(target_obj, source_obj1 [, source_obj2 ...])
 // Shallow extend, e.g.:
 //    extend({A:1}, {b:2}, {c:3}) returns {A:1, b:2, c:3}
+// XXX: This isn't going to be a part of plugin.utils. We should upgrade to Babel, Then we'll use Object.assign (or a ponyfill)
```
ariporad: I think we should upgrade to babel and use Object.assign, or we could just use [a polyfill](https://www.npmjs.com/package/object-assign)

src/common.js
```diff
@@ -248,6 +257,7 @@ function extend(target) {
 exports.extend = extend;
 
 // Common wrapper for all Unix-like commands
+// XXX: I'm unsure how this is going to integrate with plugins.
```
ariporad: So, we need to do some bikeshedding on how this is going to work.

Currently, `common.wrap` does mostly five things (all of which can be disabled with `notUnix: true`:

1. Parses `-options` if present.
2. Expands globs.
3. Expands `~`.
4. Converts ShellString to normal strings.
5. Handles the case where `common.error` (now `plugin.utils.error`) has to throw an error to stop execution, but we don't actually want to throw an error (because `config.fatal === false`).

Here's how I think this should be done:

1. Add a fourth parameter to `plugin.utils.plugin`, which is `options`. Then the first parameter is automatically `parseOptions`'d.
2. Add a fifth param, which is a list of options for `wrap`. (There's some slight tweaking I still want to do, primarily having `idx` default to 1, but that's not super relevant.)
3. All commands are wrapped by default.
4. `ShellString` conversions and `common.error` handling should be on by default, and should be a seperate option to disable.
5. Unless you say otherwise, your return value is `ShellString`'d

src/common.js
nfischer: I believe we discussed this already when this was initially added

src/common.js
nfischer: This is for unlinking files. It's used for rm and anything else that may need to do the same operation

ariporad: Would you be OK if we moved this into `src/rm.js`, and had everything else use `rm`?

src/common.js
ariporad: Hmm... Would you be ok creating a `polyfills.js` file which contains all of our polyfills?

.editorconfig
nfischer: Don't know much about this. If you're sure it doesn't break it, it's fine with me

ariporad: I'm sure. It's a standardized way to specify code style: editorconfig.org

src/common.js
nfischer: It needs to be run only once (clears out state). Doesn't matter where that happens. I think it's not a big issue, unless this file can be included multiple times for one shelljs instance

src/common.js
nfischer: Not all commands return shellstrings, so this should not happen every time.

src/common.js
nfischer: Isn't that what `common.js` is for--common utilities that most files depend on in some way? Since wrap uses this, why bother moving it out? Let me know what your thoughts are.

src/common.js
nfischer: It depends on how it's used elsewhere. If rm fails, it pollutes common.error, which might mess things up for other commands, whereas unlink itself wouldn't modify that value

src/common.js
nfischer: Is there a problem with this function? If so, let's replace it with something better.

src/common.js
ariporad: Ok.

src/common.js
ariporad: Actually, I think I know where this should go. Never mind.

src/common.js
ariporad: I'll look into this. On the whole, I think it's good to encourage commands using other commands.

src/common.js
ariporad: No, I just feel like it's a bit out of place here: Everything else is shell-related in some way, except for this, which is just a generic utility function. I don't know...

nfischer:
Closing this in favor of the in-progress implementation we have now. We should revisit this PR when we implement plugins with subcommands.

PR #453, opened 2016-05-31
feat(command): new command: uniq()

joshi-sh:
The [uniq](http://manpages.ubuntu.com/manpages/precise/en/man1/uniq.1.html) utility removes adjacent identical lines from it's input. I have implemented uniq with support for the -i (ignore case) option.


ariporad:
@nfischer: What are your thoughts on this? I'm not really sure this needs to be a part of shelljs, this could maybe be a plugin, or just a js lib... It's not really shell specific.

nfischer:
@ariporad This is part of GNU core-utils ([here's the source for that](https://www.gnu.org/software/coreutils/manual/coreutils.html#uniq-invocation)), so it seems pretty shell-specific to me. What do you think?

I think this could be a pretty useful utility for some people.

@DaedalusUsedPerl Could you please update this to fix the lint errors?

test/uniq.js
```diff
@@ -0,0 +1,59 @@
...
+    fs = require('fs');
+
+shell.config.silent = true;
+
+shell.rm('-rf', 'tmp');
+shell.mkdir('tmp');
+
+var result;
+
+result = shell.uniq();
```
nfischer: Please mark this section with:

```
//
// Invalids
//
```

See [here](https://github.com/shelljs/shelljs/blob/master/test/cat.js#L11-13) as an example.

test/uniq.js
```diff
@@ -0,0 +1,59 @@
...
+assert.ok(shell.error());
+assert.ok(result.code);
+
+assert.equal(fs.existsSync('/asdfasdf'), false); // sanity check
+result = shell.sort('/adsfasdf'); // file does not exist
+assert.ok(shell.error());
+assert.ok(result.code);
+
+//uniq file1
```
nfischer: Please mark this section with:

```
//
// Valids
//
```

See [here](https://github.com/shelljs/shelljs/blob/master/test/cat.js#L26-28) as an example.

test/uniq.js
```diff
@@ -0,0 +1,59 @@
...
+assert.equal(shell.error(), null);
+assert.equal(result.code, 0);
+assert.equal(shell.cat('resources/uniq/file1u').trimRight(), 
+             shell.cat('resources/uniq/file1t').trimRight());
+             
+//cat file1 |uniq
+result = shell.cat('resources/uniq/file1').uniq();
+assert.equal(shell.error(), null);
+assert.equal(result.code, 0);
+assert.equal(result + '', shell.cat('resources/uniq/file1u').trimRight() + '\n');
```
nfischer: Why do we need `.trimRight()` here? If this is an issue with ShellStrings being slightly different than regular strings, than you could probably get away with:

`shell.cat('resources/uniq/file1u').toString()`

src/uniq.js
```diff
@@ -0,0 +1,74 @@
...
+  uniqed = uniqed.
+             //Do we want only duplicated objects?
+             filter(function(obj){return options.duplicates ? obj.count > 1 : true;}).
+             //Are we tracking the counts of each line?
+             map(function(obj){return (options.count ? (lpad(7,obj.count) + " ") : "") + obj.ln;}).
+             join('\n') + '\n';
+
+  var res = new common.ShellString(uniqed, common.state.error, common.state.errorCode);
+  if(output){
+      res.to(output);
```
nfischer: Does anything get returned in this branch?

src/uniq.js
```diff
@@ -0,0 +1,74 @@
...
+    common.error('no input given');
+
+  var lines = (input ? fs.readFileSync(input, 'utf8') : pipe).
+              trimRight().
+              split(/\r*\n/);
+
+  //Perform a run-length encoding of the lines
+  var uniqed = [{count: 1, ln: lines[0]}];
+  lines.slice(1).forEach(function(line){
+      var cmp = options.ignoreCase ?
```
nfischer: Could you please update this to 2 spaces for indentation? Looks like it's 4 here. Thanks!

README.md
```diff
@@ -405,6 +405,24 @@ Return the contents of the files, sorted line-by-line. Sorting multiple
 files mixes their content, just like unix sort does.
 
 
+### uniq([options,] [input, [output]])
+Available options:
+
++ `-i`: Ignore differences in case when comparing
++ `-c`: Prefix lines by the number of occurrences
```
nfischer: This sounds a bit ambiguous. Could you reword this?

README.md
```diff
@@ -405,6 +405,24 @@ Return the contents of the files, sorted line-by-line. Sorting multiple
 files mixes their content, just like unix sort does.
 
 
+### uniq([options,] [input, [output]])
+Available options:
+
++ `-i`: Ignore differences in case when comparing
```
nfischer: Maybe just, "ignore case while comparing"?

README.md
```diff
@@ -405,6 +405,24 @@ Return the contents of the files, sorted line-by-line. Sorting multiple
...
++ `-c`: Prefix lines by the number of occurrences
++ `-d`: Only print duplicate lines, one for each group
```
nfischer: It's unclear what a "group" is here

shell.js
```diff
@@ -94,6 +94,10 @@ exports.sed = common.wrap('sed', _sed, {idx: 3, canReceivePipe: true}); // don't
 var _sort = require('./src/sort');
 exports.sort = common.wrap('sort', _sort, {idx: 1, canReceivePipe: true});
 
+//@include ./src/uniq
+var _uniq = require('./src/uniq');
+exports.uniq = common.wrap('uniq', _uniq, {idx: 1, canReceivePipe: true});
```
nfischer: Thanks for the refactor!

src/uniq.js
```diff
@@ -0,0 +1,76 @@
+var common = require('./common');
+var fs = require('fs');
+
+//add c spaces to the left of str
+function lpad(c, str){
+    var res = "" + str;
```
nfischer: Could you make this an indent of 2 spaces please?

src/uniq.js
```diff
@@ -0,0 +1,76 @@
...
+    if(res.length < c){
+        res = Array((c-res.length)+1).join(" ") + res;
```
nfischer: nit: we prefer single-quoted strings over double-quoted strings when convenient

src/uniq.js
```diff
@@ -0,0 +1,76 @@
...
+  var pipe = common.readFromPipe(this);
+
+  if (!input && !pipe)
+    common.error('no input given');
+
+  var lines = (input ? fs.readFileSync(input, 'utf8') : pipe).
+              trimRight().
+              split(/\r*\n/);
+
+  //Perform a run-length encoding of the lines
```
nfischer: What does this comment mean? It'd be better if it was a bit more descriptive.

src/uniq.js
```diff
@@ -0,0 +1,76 @@
...
+  var uniqed = [{count: 1, ln: lines[0]}];
+  lines.slice(1).forEach(function(line){
+    var cmp = options.ignoreCase ? 
+                line.toLocaleLowerCase().localeCompare(uniqed[uniqed.length-1].ln.toLocaleLowerCase()) :
```
nfischer: Is there a reason this uses `toLocaleLowerCase` vs plain `toLowerCase`? If so, that's fine. I just don't want any unexpected results down the road.

nfischer: Can we move `uniqed[uniqed.length-1].ln` out of the loop?

src/uniq.js
```diff
@@ -0,0 +1,76 @@
...
+                line.localeCompare(uniqed[uniqed.length-1].ln);
+    if(cmp !== 0){
+      uniqed.push({count: 1, ln: line});
+    }else{
+      uniqed[uniqed.length - 1].count++;
+    }
+  });
+  uniqed = uniqed.
+             //Do we want only duplicated objects?
+             filter(function(obj){return options.duplicates ? obj.count > 1 : true;}).
```
nfischer: Could you write out callbacks in the form:

```javascript
filter(function (obj) {
  // Do we want only duplicated objects?
  return options.duplicates ? obj.count > 1 : true;
}).map( // etc ...
```

nfischer:
@joshi-sh I left some comments. Overall, looks pretty thorough! This is something I'd really like to see in ShellJS. Most of the comments are very nitpicky, so I can fix them if you'd like.

src/uniq.js
joshi-sh: Just to keep it consistent with `sort`, which uses `localeCompare`.

nfischer: @joshi-sh Yeah, I know it uses `localeCompare`, but why `toLocaleLowerCase`?

joshi-sh:
Your comments are pretty reasonable, actually. I just copied the GNU man pages in lieu of writing docs.

src/uniq.js
joshi-sh: Some locales (like Turkish) don't follow the usual case mappings in Unicode. In most locales, it behaves exactly the same as `toLowerCase`.

nfischer:
Gonna give this an LGTM. We might want to wait to merge until after shelljs v0.8.1, and then this can be in 0.8.2.

nfischer:
LGTM :+1:

nfischer:
I'll handle merging this once we get a patch release out with some of the fixes we've made since v0.7.0.

PR #454, opened 2016-06-01
chore: update jshint and move it to an npm script

nfischer:
I think this is worth it because:

 1. `jshint` is udpated, so we'll get whatever bug fixes there may have been as well as performance improvements
 2. This is now an npm package script instead of part of `run-tests.js`. Although cutting down on code is nice, this really shines in that it exposes linting to the commandline (so you can lint without re-running tests if you choose), and it lets jshint figure out which files to lint, instead of using `common.expand()` to expand the `*` character for Windows
 3. I removed most re-declarations of variables just because I know that's an issue for eslint, and I know we may migrate to use that in the future. This wasn't required by jshint, however
 4. I made linting a bit stricter, and actually caught some confusing parts of the code
 5. Linting now runs after tests. I think this is valuable if you want to first get your code correct, and then afterward fix it for stylistic consistency.

### DIFF
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,0 +1,2 @@
+test/resources/
+node_modules/
--- a/.jshintrc
+++ b/.jshintrc
@@ -3,5 +3,10 @@
   "sub": true,
   "undef": true,
   "unused": true,
+  "futurehostile": true,
+  "latedef": "nofunc",
+  "nocomma": true,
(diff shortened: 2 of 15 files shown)


nfischer:
@ariporad rebased this off master

ariporad:
LGTM!

PR #457, opened 2016-06-09
chore: clarify message for when docs are not generated

nfischer:
I noticed an issue we had was that if someone fails to generate docs, it isn't always clear to them that travis is failing for that reason. I wrote out my [own module](https://github.com/nfischer/travis-check-changes) to handle this sort of situation, but with hopefully nicer error reporting.

This PR updates ShellJS to use the module as a dev dependency, with an error message that is hopefully a bit more explanatory than what we had before.


ariporad:
LGTM!

PR #458, opened 2016-06-09
chore: revert depreciate shelljs/make (#431)

zephraph:
This reverts commit 5a31c7c4369b5e6dbf71b005d040c525ec07b68e from #431.

This reverts the deprecation of the make command as noted in #340.

### DIFF
--- a/make.js
+++ b/make.js
@@ -1,8 +1,5 @@
 require('./global');
 
-console.error('WARNING: shelljs/make is deprecated as of ShellJS v0.8.');
-console.error('Please migrate your code to use an alternate task runner.');
-
 global.config.fatal = true;
 global.target = {};
 


nfischer:
I'll leave this to @ariporad to review. This commit seems to properly revert the feature, if that's something we want to do.

Alternative approaches would be to display a different warning message instead.

zephraph:
What different message would be displayed? Unless the deprecation is followed through with I wouldn't think there's a need to have a warning message.

ariporad:
@zephraph: we've decided that this till will be unsupported going forward, so we're going to say something along those lines. Do you want to go ahead and add that, or should I?

zephraph:
I don't think a message is necessary. Deprioritize it as a feature for support and shift the burden of maintenance onto the community. We could add a message in the documentation to explain as much.

nfischer:
@zephraph's approach sounds good to me. LGTM

ariporad:
SGTM

PR #459, opened 2016-06-14
chore: update release process

nfischer:
Similar to shelljs/shx#63, this switches to the new release script. I also updated the release steps accordingly. Feel free to double check the release steps I wrote out, since we want to be sure these are in the right order.


nfischer:
Merge conflicts have now been resolved

ariporad:
LGTM!

PR #460, opened 2016-06-14
chore: switch to shelljs-changelog

nfischer:
Same script, but now it's refactored out as a dependency, and it runs in Node instead of bash.

Related: shelljs/shx#64


nfischer:
This is bumped to the most recent version of shelljs-changelog

ariporad:
LGTM!

PR #461, opened 2016-06-14
fix(mv): works across partitions

### LINKED ISSUE #1: #mv Won't Work Across Disks

nfischer:
Fixes #1 

Error reporting seems to work correctly: if an error is encountered at any point, the reported message still indicates it's the `mv` command that had the error.

I couldn't think of a way to mock tests for this, but let me know if you can think of something.

### DIFF
--- a/src/mv.js
+++ b/src/mv.js
@@ -1,6 +1,8 @@
 var fs = require('fs');
 var path = require('path');
 var common = require('./common');
+var cp = require('./cp');
+var rm = require('./rm');
 
 //@
 //@ ### mv([options ,] source [, source ...], dest')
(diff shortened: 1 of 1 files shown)


nfischer:
I tested this by moving files from my harddrive to a flash drive

ariporad:
LGTM

I wonder if we could test this with a virtual disk image. (I'm thinking something similar to how mac apps a distributed, like a .dmg file)

nfischer:
Huh? Are we talking about the same thing? This was supposed to be about a bug involving the `mv` command across disk partitions, not about how we distribute the software.

ariporad:
@nfischer: Yeah. I was thinking that a virtual dist and/or a VMish (like Docker) might work to simulate that for testing.

nfischer:
Is there a simple easy to set that up? This is a pretty obscure edge case. I'd rather not mess with docker as a dev dependency for something like this.

ariporad:
Probably not... Let's forget that for now.

nfischer:
Ok, I'll go ahead and merge this unless there's any objections

PR #464, opened 2016-06-17
refactor: expose pipe-ability to command configuration

nfischer:
I think this is one of the first steps we'll need to take in order to create a better setup for creating new commands/plugins. Now, contributors need only specify that their command can receive input from a pipe with the `receivesPipe` option set to `true`, instead of modifying the array that used to explicitly contain each command that could receive input from a pipe.

This cuts out one more spot where contributors need to make a modification.

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -71,15 +71,15 @@ exports.test = common.wrap('test', _test);
 
 //@include ./src/cat
 var _cat = require('./src/cat');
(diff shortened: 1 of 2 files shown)


ariporad:
LGTM!

nfischer:
Updated this to have slightly clearer identifier names. Unless the new names are an issue, I'll go ahead and merge this after CI completes.

ariporad:
LGTM

PR #466, opened 2016-06-22
fix(cp): -n option no longer raises error

### LINKED ISSUE #465: cp -n doesn't work correctly
### Node version:
(probably any)

v0.7

### Operating system:
(probably any)

### Description of the bug:
`cp('-n', 'foo', 'bar')` fails if `bar` exists. This should succeed but do nothing.

See above

This is related to shelljs/shx#67, as pointed out by @kirly-af

nfischer:
Fixes #465


ariporad:
LGTM once travis passes

PR #467, opened 2016-06-23
fix(ls): properly handle symlinks without a trailing slash on macOS

### LINKED ISSUE #446: ls() not returning an Array

ariporad:
This is a really cheap hack, but it works.

Fixes #446.


ariporad:
Could someone with Windows clone this branch and tell me if `test/resources/ls2` exists, and if it's properly a symlink that points to `test/resources/ls`? I think that's why AppVeyor is failing.

nfischer:
@ariporad This breaks compatibility with unix `ls` for regular files. Ex:

```javascript
> exec('ls file.txt');
file.txt
> ls('file.txt');
ls: no such file or directory: file.txt
```

This bug is not visible on the master branch of ShellJS.

We should probably add tests for `ls()` being called on regular files though (can't believe this wasn't already tested).

ariporad:
Hmm...

Maybe we should just file a bug in node for this... It doesn't really seem like our problem

nfischer:
Whether or not this is an issue we should address, it seems clear that this approach isn't the right way to address this.

@ariporad If you think this is a bug with node's core, feel free to file a report.

PR #470, opened 2016-07-01
chore: switch to shields.io, and add npm badge

nfischer:
Similar to shelljs/shx#70, but also keeps badges consistent with shx


nfischer:
@ariporad I'm gonna go ahead and give this an LGTM and merge once shields.io comes back up (looks like their servers are down right now)

nfischer:
LGTM

PR #472, opened 2016-07-05
Add multiline cmd support in exec

pavel-g:
child_process.execSync allow to use multiline commands, but shelljs.exec doesn't support such commands.

Please, add support such commands.

Sorry for my english :(

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -64,11 +64,15 @@ function execSync(cmd, opts, pipe) {
   opts.cwd = path.resolve(opts.cwd);
   var optString = JSON.stringify(opts);
 
+  var multilineCmdToScript = function(cmd) {
+    var lines = cmd.split("\n");
+    return JSON.stringify(lines) + '.join("\\n")';
+  };
   if (typeof child.execSync === 'function') {
     script = [
         "var child = require('child_process')",
         "  , fs = require('fs');",
-        "var childProcess = child.exec("+JSON.stringify(cmd)+", "+optString+", function(err) {",
+        "var childProcess = child.exec("+multilineCmdToScript(cmd)+", "+optString+", function(err) {",
         "  fs.writeFileSync("+JSON.stringify(codeFile)+", err ? err.code.toString() : '0');",
         "});",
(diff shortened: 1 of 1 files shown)


pavel-g:
I use multiline commands from shelljs for backup scripts. My local script execute commands on remote server over ssh.

For example:

require('shelljs/global');
// compose multiline command:
var cmd = [
  'ssh root@example.org bash <<EOF',
  'rm dump.sql'
  'mysqldump ... > dump.sql',
  // and more another commands
  'EOF'
].join("\n");
exec(cmd);

In my opinion, construct "bash <<EOF" is more useful then complex command with separators "&&" or ";".

pavel-g:
> I'm concerned that splitting on newlines could break behavior for shell commands that contain a newline in them.

You are right. Maybe command must split with regexp.

ariporad:
Hi @pavel-g!

My $0.02 is that you should just join commands with semicolons, not new lines. We already support that, and there are valid commands which are multiline. (Correct me if I'm misunderstanding).

For example, you could do:

```javascript
require('shelljs/global');
// compose multiline command:
var cmd = [
  'ssh root@example.org bash <<EOF',
  'rm dump.sql'
  'mysqldump ... > dump.sql',
  // and more another commands
  'EOF'
].join(";"); // <--- We join them with ;'s, not \n's

exec(cmd);
```

pavel-g:
```javascript
var cmd = [
  'ssh root@example.org bash <<EOF',
  'rm dump.sql',
  'mysqldump ... > dump.sql',
  // and more another commands
  'EOF'
].join(";"); // <--- We join them with ;'s, not \n's
```

return string: `ssh root@example.org bash <<EOF;rm dump.sql;mysqldump -h localhost > dump.sql;EOF`

bash not executed that command, because it waiting EOF in new line

pavel-g:
I found workaround:

1. Command joined with ";"
2. Redirect ">" replaced with "\\>"

nfischer:
Glad to see you found a workaround!

PR #474, opened 2016-07-20
fix: mkdir for invalid perms does not kill process

### LINKED ISSUE #473: shelljs should not kill process if node call throws exception
### Node version:
4.4.7

0.7.0

### Operating system:
Windows 10

### Description of the bug:
I'm working on code that calls shelljs mkdir to create a directory. On Windows, the call to fs.mkdir() sometimes fails because of permission issues or whatever. This is a failure that I should be able to catch with a try/catch. But shelljs catches the error, outputs it to the console and exits the process ([here](https://github.com/shelljs/shelljs/blob/master/src/common.js#L333)). A utility like shelljs should never take such drastic action - it should be up to me the consumer to decide what to do with such an error.

On windows:

```javascript
shelljs.mkdir('c:\\Program Files\\foo');
```

It should just throw in this situation so I can catch the error and decide what to do.

nfischer:
Fixes #473 

@ariporad, ready for review

### DIFF
--- a/src/mkdir.js
+++ b/src/mkdir.js
@@ -62,10 +62,15 @@ function _mkdir(options, dirs) {
       return; // skip dir
     }
 
-    if (options.fullpath)
-      mkdirSyncRecursive(dir);
-    else
-      fs.mkdirSync(dir, parseInt('0777', 8));
+    try {
+      if (options.fullpath)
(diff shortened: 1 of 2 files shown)


ariporad:
LGTM

PR #475, opened 2016-07-21
refactor: commands now register themselves

nfischer:
This is based off the plugin API spec. All ShellJS commands now register themselves from within their source file. The only modification that is required to `shell.js` is to put:

```javascript
//@include ./src/foo
require('./src/foo');
```

for a hypothetical new command named `foo()` implemented inside of `src/foo.js`.

In a possible future plugin, the `//@include` would be unnecessary (that's for our internal docs) and the `require()` statement would be all that is needed by the plugin's end-user. The plugin hypothetically doesn't even need to export anything, because the plugin author can require ShellJS as a peer dependency and modify the existing ShellJS instance using a call to `common.register()`.

Let's wait to merge this until after #452 is merged, since I can quickly resolve the merge conflict. Send feedback my way in the meantime, however.

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -8,7 +8,6 @@
 
 var common = require('./src/common');
 
-
 //@
 //@ All commands run synchronously, unless otherwise stated.
 //@ All commands accept standard bash globbing characters (`*`, `?`, etc.),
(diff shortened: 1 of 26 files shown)


ariporad:
LGTM!

PR #477, opened 2016-07-22
Fix mkdir malformed path

nfischer:
closes #468 

This is the same PR, but with style fixes.


nfischer:
Going to go ahead and merge, since this is essentially the same as the other PR. LGTM

PR #478, opened 2016-07-22
refactor: hook new uniq() command using new format

nfischer:
Fixes the new uniq() command to use the new format. This also adds a test to make sure it permits globbing (this was broken after merging #452 because of a recent refactor).


nfischer:
I'm going to go ahead and merge this, since it just updates to the new refactor.

LGTM

PR #479, opened 2016-07-23
refactor: move option parsing into common.wrap()

nfischer:
This is roughly in-line with what was proposed for the plugin API. This moves option parsing as part of `common.wrap()`, so contributors don't need to worry about making the call themselves. By default, `common.wrap()` acts as it did before, and will only parse options if options are specified when the plugin is registered.

For trickier commands, like `chmod()` and `set()`, which sometimes need special parsing, the contributor can choose to parse these commands with an explicit call to `common.parseOptions()`.

This deviates from the plugin API draft in that command options are passed in via the `cmdOptions` attribute, which I believe is more descriptive than an anonymous parameter to `common.register()`. This also makes it easier to leave option parsing optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,9 @@ introduced between each file).
 
 ### head([{'-n', \<num\>},] file [, file ...])
 ### head([{'-n', \<num\>},] file_array)
+Available options:
+
++ `-n <num>`: Show the first `<num>` lines of the files
 
 Examples:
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -318,6 +318,9 @@ introduced between each file).
 
 ### head([{'-n', \<num\>},] file [, file ...])
 ### head([{'-n', \<num\>},] file_array)
+Available options:
```
ariporad: Why these changes? They seem unrelated.

nfischer: @ariporad As I was digging through, I saw that the documentation had an error (it was missing the available option). You're correct, these changes are unrelated to the refactor.

ariporad: kk. LGTM

ariporad:
LGTM

PR #481, opened 2016-07-24
refactor: add wrapOutput option to auto-ShellString-ify command output

nfischer:
This lets you return a plain JS string (like before), but pass the `wrapOutput` option to `common.register()` in order to wrap this string in a ShellString. This wraps it with the following default parameters (which I think will work properly most of the time):

```js
return myReturnValue;
// This gets wrapped to imitate the statement:
return new common.ShellString(myReturnValue, common.state.error, common.state.errorCode);
// If any errors have occurred, put these in the .stderr attribute
// Whatever the resulting exit code is (0 unless there's an error), set that for the .code attribute
```

By default, `common.register()` works as before (so collaborators must opt-in for this feature). This lets functions return non-string output, or return a custom-built ShellString (if these default values aren't suitable).

**Do not merge until after #479 is either closed or merged**

### DIFF
--- a/src/cat.js
+++ b/src/cat.js
@@ -1,7 +1,11 @@
 var common = require('./common');
 var fs = require('fs');
 
-common.register('cat', _cat, {globStart: 1, canReceivePipe: true});
+common.register('cat', _cat, {
+  globStart: 1,
+  canReceivePipe: true,
+  wrapOutput: true,
+});
 
 //@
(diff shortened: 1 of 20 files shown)


ariporad:
LGTM!

nfischer:
This is just updated to resolve the conflicts introduced by #479. I'm going to re- LGTM this, since it's pretty trivial.

PR #482, opened 2016-07-24
fix: resolve a cylcic-dependency problem

nfischer:
If a plugin was imported before the ShellJS instance was created, the user's script would crash (`common.register` wouldn't be instantiated yet). Now plugins can be imported either before or after the ShellJS instance is created.

One really cool consequence of this fix is that now plugins are compatible with the `shelljs/global` script:

```js
require('shelljs-plugin-open');
require('shelljs/global');

open('file.txt'); // it works!
```

Another cool consequence is that this would allow me to rewrite [shelljs-exec-proxy](https://github.com/nfischer/shelljs-exec-proxy) as a ShellJS plugin, which would let it play nicely with other plugins (such as shelljs-plugin-open).

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -6,7 +6,13 @@
 // http://github.com/arturadib/shelljs
 //
 
-var common = require('./src/common');
+var commonPath = './src/common';
+var common = require(commonPath);
+if (!common.register) {
+  // If this isn't defined yet, we haven't fully finished loading ./src/common
+  delete require.cache[require.resolve(commonPath)];
+  common = require(commonPath);
+}
 
 //@
 //@ All commands run synchronously, unless otherwise stated.


nfischer:
Also, this should have no influence whatsoever on ShellJS when used normally, but would only help make plugins a bit easier :smile:

nfischer:
@ariporad A plugin would `require('src/common')` (or whatever file we eventually move the plugin utils into) in order to register its own commands (using `common.register()`).

If the plugin is included before the ShellJS instance is created (`require('shelljs-plugin-open')` precedes `require('shelljs')`), then `src/common.js` is `require()`ed before `shell.js`. This is where the problems arise. `src/common.js` [includes `shell.js`](https://github.com/shelljs/shelljs/blob/master/src/common.js#L8), so if `shell.js` hasn't yet been `require()`ed, Node goes and loads in `shell.js`. [This tries to load `src/common.js`](https://github.com/shelljs/shelljs/blob/master/shell.js#L9), but since we were in the middle of loading `src/common.js`, this returns `common = {}` (the empty object). Then, when `shell.js` loads in the built-in commands, it tries to `common.register()` them, but fails. This is because `common` is still `{}`, and doesn't have the `.register` attribute yet (and it won't until the initial `require()` completes). In essence, **`common.register()` is used before it gets defined.**

The problem didn't arise for normal ShellJS use because `shell.js` is `require()`ed first, so it is able to load in `src/common.js` completely before continuing to load each command (**so `common.register()` is defined before it is used**).

The solution here is to check if `common.register()` is defined as soon as `shell.js` is loaded for the first time. If `common.register` is not yet defined, invalidating the require-cache will force Node to load all of `src/common.js` (making sure that **`common.register()` gets defined before it's used**).

Hopefully that's a good explanation. I had to draw out dependency graphs to discover what the problem was and how to solve it. From what I've seen, however, it seems to work. I'll update the plugin after this is merged as a proof of concept.

ariporad:
Ahh... That makes sense. Thanks!

nfischer:
Yeah, it's a bit non-intuitive, but hopefully that was a clear enough explanation for all looking into the matter in the future.

PR #484, opened 2016-07-26
refactor: expose plugin utils & add initial tests

nfischer:
After more experimentation, I realized that #482 wasn't a perfect approach, but that this would be a better one. Even after #482, `shell.error()` was actually broken. This fixes that issue.

This also starts differentiating internal implementations (`src/common.js`) from exposed parts of the plugin API (`plugin.js`). This is of course still far from a complete plugin API, but this should be enough to make simple plugins (and still get benefits like globbing, shelljs-style error signaling, etc.).

All implementations are still within `src/common.js`, and internal commands still rely on that directly, but external plugins are intended to use `var plugin = require('shelljs/plugin')` to get functions like `plugin.register()`. This also makes it easy to modify the plugin API without breaking ShellJS commands.

I also started unit tests for the plugin API. This creates one sample plugin (with all the bells and whistles like automatic option parsing) and then makes sure it works properly. We may also want to create a simpler plugin (i.e. no automatic option-parsing, no wrapping the return value, etc.) and test that.

If you have any suggestions on naming conventions, or think there's a better approach to this issue, let me know. Like I said, this is still far from complete, but suggestions are welcome.

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -6,13 +6,7 @@
 // http://github.com/arturadib/shelljs
 //
 
-var commonPath = './src/common';
-var common = require(commonPath);
-if (!common.register) {
-  // If this isn't defined yet, we haven't fully finished loading ./src/common
-  delete require.cache[require.resolve(commonPath)];
-  common = require(commonPath);
-}
+var common = require('./src/common');
 
 //@
 //@ All commands run synchronously, unless otherwise stated.
--- a/plugin.js
+++ b/plugin.js
@@ -1,0 +1,16 @@
+// Various utilties exposed to plugins
+
+require('.'); // Create the ShellJS instance (mandatory)
+
+var common = require('./src/common');
+
+var exportedAttributes = [
+  'error',        // For signaling errors from within commands
+  'parseOptions', // For custom option parsing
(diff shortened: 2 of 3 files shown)


shell.js
```diff
@@ -6,13 +6,7 @@
 // http://github.com/arturadib/shelljs
 //
 
-var commonPath = './src/common';
```
ariporad: Why are we deleting this?

nfischer: This reverts #482, which I realized was not the best approach. We could leave #482 in, but it's just extra-complicated code once this gets merged, so I think it's best to revert.

nfischer:
@ariporad I just fixed the issue with node v0.10-12. This should be good now.

nfischer:
For reference, the bug I found with the approach in #482 is that there are two different instances of `common.state`: one that the plugin has access to and modifies, and one that the ShellJS instance has access to.

A plugin might signal an error using `plugin.error('some message')`, however any calls to `shell.error()` after that would not see that error message, since `shell.error()` accesses one `common.state` and `plugin.error()` modified a different one.

ariporad:
LGTM!

PR #488, opened 2016-07-27
feat(mktemp): add `mktemp` command.

ariporad:
Sorry, I thought I already had a PR for this.


shell.js
```diff
@@ -97,6 +97,11 @@ require('./src/which');
 //@include ./src/echo
 require('./src/echo');
 
+
+//@include ./src/mktemp
+var _mktemp = require('./src/mktemp');
+exports.mktemp = common.wrap('mktemp', _mktemp, {idx: 1});
```
nfischer: Could you refactor this to use `common.register()` in the source file itself?

src/mktemp.js
```diff
@@ -0,0 +1,88 @@
...
+//@ Creates a temporary file or directory in a sutible place, with a random, available name. You can optionally pass
+//@ one or more templates, which will override the default one. Each template will have all trailing `X`'s replaced with
+//@ a random letter, and will then be used as a path to create a temporary file/directory.
+//@
+//@ *WARNING: You MUST delete the file/directory when you're done with it, otherwise it will remain there forever.*
+function mktemp(options, templates) {
+  options = common.parseOptions(options, {
+    d: 'directory',
+    u: 'unsafe',
+  });
```
nfischer: Can you switch the `parseOptions` to the `cmdOptions` option in `common.register()`?

src/exec.js
```diff
@@ -15,12 +15,11 @@ common.register('exec', _exec, {notUnix:true, canReceivePipe: true});
...
-  var stdoutFile = path.resolve(tempDir+'/'+common.randomFileName()),
-      stderrFile = path.resolve(tempDir+'/'+common.randomFileName()),
-      codeFile = path.resolve(tempDir+'/'+common.randomFileName()),
-      scriptFile = path.resolve(tempDir+'/'+common.randomFileName()),
-      sleepFile = path.resolve(tempDir+'/'+common.randomFileName());
+  var stdoutFile = mktemp('-u')[0],
+      stderrFile = mktemp('-u')[0],
+      codeFile = mktemp('-u')[0],
+      scriptFile = mktemp('-u')[0],
+      sleepFile = mktemp('-u')[0];
```
nfischer: If you switch to the automated option parsing, you'll need to switch these calls to `mktemp({unsafe: true})`.

test/mktemp.js
```diff
@@ -0,0 +1,112 @@
...
+var isNotWindows = common.platform !== 'win';
+var result;
+
+
+//
+// Invalids
+//
+
+shell.touch(path.resolve('.', 'tmp/tmp.A'));
+result = shell.mktemp(path.resolve('.', 'tmp/tmp.A')); // Guarenteed to have a confilct, no randomness.
```
nfischer: Is it possible to actually specify a name like this? I just tried `$ mktemp foo` and it says "too few X's in template 'foo'" (and exits with status 1).

README.md
```diff
@@ -472,6 +472,29 @@ Prints string to stdout, and returns string with additional utility methods
 like `.to()`.
 
 
+### mktemp([options,] [templates...])
+
+Available Options:
+
++ `-d`: Create a directory instead of a file.
++ `-u`: Unsafe mode. (Delete the file before retuning). Only use if you know what you're doing.
```
nfischer: Is this what `-u` does? It says in the Linux man page that it performs a dry-run (doesn't create anything).

README.md
```diff
@@ -472,6 +472,29 @@ Prints string to stdout, and returns string with additional utility methods
...
+
+Examples:
+
+```javascript
+mktemp() // ['/tmp/tmp.shelljs.QOL1QLNZMSR0HR4S5FTS']
+mktemp('-d') // ['/tmp/tmp.shelljs.EGBZ4GVYFO4SO534F3WK']
+mktemp('/tmp/tmp.shelljs.foo.XXXXX') // ['/tmp/tmp.shelljs.foo.AQZJK']
+mktemp('-d', '/tmp/shelljs.foo.XXXXX', '/tmp/shelljs.bar.XXXXX') // ['/tmp/tmp.shelljs.foo.ZFUPD', '/tmp/tmp.shelljs/bar.I9XVF']
```
nfischer: Does Mac `mktemp` allow more than one template? For me it only allows one on Linux.

test/mktemp.js
```diff
@@ -0,0 +1,112 @@
...
+result = shell.mktemp('-d');
+assert.equal(shell.error(), null);
+assert.equal(result.code, 0);
+assert.equal(result.length, 1);
+assert.ok(fs.existsSync(result[0]));
+assert.ok(fs.statSync(result[0]).isDirectory());
+if (isNotWindows) assert.equal(fs.statSync(result[0]).mode & parseInt('777', 8), parseInt('0600', 8));
+
+// Custom Template
+result = shell.mktemp(path.resolve('.', 'tmp/tmp.XXX'));
```
nfischer: Why use `path.resolve()` here? Relative paths should be supported as well I think.

test/mktemp.js
```diff
@@ -0,0 +1,112 @@
...
+assert.ok(fs.statSync(result[0]).isDirectory());
+assert.ok(fs.existsSync(result[1]));
+assert.ok(fs.statSync(result[1]).isDirectory());
+if (isNotWindows) assert.equal(fs.statSync(result[0]).mode & parseInt('777', 8), parseInt('0600', 8));
+
+// Unsafe Mode
+result = shell.mktemp('-u');
+assert.equal(shell.error(), null);
+assert.equal(result.code, 0);
+assert.equal(result.length, 1);
```
nfischer: Could you also verify that this defaults to an appropriate temp directory, for both Unix and Windows?

src/mktemp.js
```diff
@@ -0,0 +1,88 @@
...
+var fs = require('fs');
+var common = require('./common');
+var mkdir = require('./mkdir');
+var touch = require('./touch');
+var rm = require('./rm');
+var chmod = require('./chmod');
+
+var tempdir;
+var DEFAULT_TEMPLATE = 'tmp.shelljs.XXXXXXXXXXXXXXXXXXXX';
+var LETTERS = 'abcdefghijklmnopqrstuvwkyz1234567890'.toUpperCase().split('');
```
nfischer: This should support both uppercase and lowercase. Perhaps:

```js
var LETTERS = 'abcdefghijklmnopqrstuvwkyz';
var NUMBERS = '0123456789';
var CHARACTERS = (LETTERS + LETTERS.toUpperCase() + NUMBERS).split('');
```

README.md
```diff
@@ -472,6 +472,29 @@ Prints string to stdout, and returns string with additional utility methods
...
+```
+
+Creates a temporary file or directory in a sutible place, with a random, available name. You can optionally pass
```
nfischer: s/sutible/suitable

README.md
```diff
@@ -472,6 +472,29 @@ Prints string to stdout, and returns string with additional utility methods
...
+one or more templates, which will override the default one. Each template will have all trailing `X`'s replaced with
+a random letter, and will then be used as a path to create a temporary file/directory.
```
nfischer: nit: I would change this to "random character", since it includes digits as well. I suppose "random alphanumeric character" is most correct.

ariporad:
@nfischer: Ready for Review

src/mktemp.js
```diff
@@ -0,0 +1,88 @@
...
+//@ a random letter, and will then be used as a path to create a temporary file/directory.
+//@
+//@ *WARNING: You MUST delete the file/directory when you're done with it, otherwise it will remain there forever.*
+function mktemp(options, templates) {
+  options = common.parseOptions(options, {
+    d: 'directory',
+    u: 'unsafe',
+  });
+  templates = Array.prototype.slice.call(arguments);
+  templates.shift(); // Get rid of options.
```
nfischer: If you change the line above this, you can delete this line: `s/call(arguments)/call(arguments, 1)` :smile:

src/mktemp.js
```diff
@@ -0,0 +1,88 @@
+var path = require('path');
+var fs = require('fs');
+var common = require('./common');
+var mkdir = require('./mkdir');
+var touch = require('./touch');
+var rm = require('./rm');
+var chmod = require('./chmod');
+
+var tempdir;
+var DEFAULT_TEMPLATE = 'tmp.shelljs.XXXXXXXXXXXXXXXXXXXX';
```
nfischer: Shouldn't the default be to put it in `/tmp` (i.e. `/tmp/tmp.PTgV6UfUKO`)?

nfischer:
@ariporad I left comments in-line. There's a lot, but they should be little fixes.

README.md
ariporad: Basically, but it makes sure that you have the permissions to create the file.

This would be useful if you had a command which took a path, and put something there, but that path wasn't allowed to exist beforehand. However, it's unsafe because theoretically someone else could put something there first.

src/mktemp.js
ariporad: macOS uses a case-insensitive file system, so that would break on it.

README.md
ariporad: It does for me at least. And I figure that it can't hurt.

README.md
nfischer: I think it's best to just produce a random name. That's how the Linux command works, even if you can't put anything there. Otherwise, we run the risk of creating a file but not being able to delete it, or throwing an error when the file can't be created, when all it should do is make the name.

I would just mark it as the name "dry-run", since "unsafe" sounds like it means "insecure" (which really, it should be more secure since it shouldn't actually touch the filesystem). This is of course all based on the Linux behavior, and if Mac behavior actually checks if it's writable, let me know.

src/mktemp.js
nfischer: Ah, thanks for reminding me. Your approach is better in that case.

shell.js
nfischer: Also, it doesn't seem to me like globbing would be advantageous here (and it would almost always fail, since few files contain `XXX` in their name).

src/mktemp.js
ariporad: Good point.

nfischer: We should probably just rely on the value of `tempdir`, since that's cross-platform

test/mktemp.js
ariporad: It works for me on macOS.

test/mktemp.js
ariporad: I'm not really sure what you mean here.

nfischer: On Linux, the default directory is `/tmp`:

```bash
$ mktemp
/tmp/tmp.LP43BPDpbn
```

I imagine the default directory will be the same on Mac OS X, however this obviously wouldn't fly on Windows. Whatever `tempdir()` returns should be sufficient though.

src/tempdir.js
```diff
@@ -12,10 +20,8 @@ function writeableDir(dir) {
   if (!fs.statSync(dir).isDirectory())
     return false;
 
-  var testFile = dir+'/'+common.randomFileName();
   try {
-    fs.writeFileSync(testFile, ' ');
-    common.unlinkSync(testFile);
+    getMktemp()('-u', path.resolve(dir, 'tmp.shelljs.XXXXXXXXXX'));
```
nfischer: Also here, this should be updated to `({ unsafe: true }, ...)` instead of `('-u', ...)`

README.md
ariporad: I'm pretty sure the mac version does indeed create and delete it, at least according to the man pages. But I'm happy to change it if you want.

ariporad:
@nfischer: Should be fixed.

README.md
nfischer: Yeah, so it looks like the Linux and Mac behavior do indeed deviate a bit. But this seems to be in-line with what the Mac man page says.

nfischer:
@ariporad looks like there's some merge conflicts, and one of the test cases appears to be failing.

I'm  concerned that the Mac and Linux behavior seem to be a bit different. I dug into things a bit, and I believe that POSIX doesn't actually specify `mktemp` (which explains why there's so many differences). Either platform we choose to follow, there's the chance that someone is expecting the other behavior. This is a less-common command, so can maybe get away with that.

Another approach would be to develop this as an external plugin. The upside is that I think it's more acceptable for a plugin to deviate from POSIX standards (ex. `open` on Mac is like `xdg-open` on Linux, since `open` on Linux is a very different thing). Also, it would give you a chance to try out the as-is API first-hand and figure out its pain-points. The downside, however, is that we wouldn't be able to use this to refactor ShellJS like in this PR.

nfischer:
Merge conflicts have been resolved with latest master. This still needs some work:

 - [ ] verify test cases
 - [ ] verify test coverage is 100%
 - [ ] add `tempdir()` tests back (it's being deprecated, not being removed)
 - [ ] squash commits (optional)
 - [ ] make sure this is somewhat unix compliant

nfischer:
Closing due to merge conflicts and general lack of need/interest.

PR #489, opened 2016-07-27
fix: conflicting options now properly override each other

nfischer:
Found this bug by looking at @ariporad's branch [here](https://github.com/shelljs/shelljs/commit/50e741e2e08e5e521e012d53d1fb3b19bfc0191e#diff-2c36c683dff5798fefed561dbec17803R158). Credit for finding this goes to him.

I added a unit test for it, and sure enough, the unit test won't pass unless this patch is applied.


ariporad:
LGTM!

PR #490, opened 2016-07-27
refactor: switch common.extend() to Object.assign ponyfill

nfischer:
Not a big deal, but I figure we may as well rely on the native implementation when it's available.

We could consider using [this package](https://www.npmjs.com/package/object-assign) instead, but it's probably not a big concern either way.


ariporad:
LGTM

PR #491, opened 2016-07-27
refactor: switch from notUnix to unix in wrap()

nfischer:
This is another refactor idea I got from one of @ariporad's refactors (I forget where exactly).

Instead of specifying `notUnix: true`, this specifies `unix: false` (probably a bit clearer down the road).


ariporad:
LGTM!

ariporad:
Force-merging this because LGTM isn't working.

PR #492, opened 2016-07-27
refactor: glob by default for commands

nfischer:
This PR accomplishes:

- `globStart` defaults to 1
- commands have argument globbing by default
- to opt out of globbing, use `allowGlobbing: false`

Commands have been refactored to take advantage of this new default value.

Also, this PR begins to store default values in a JS object for each `wrapOption`, which will probably make it easier to keep track of the defaults in the long run. Because this uses `common.expand()`, this will have a slight conflict with #490 (this is just because I renamed `expand` to `objectAssign` in that PR). I'll resolve the conflict after either PR is merged and update accordingly.

This might be a good time to discuss if we want `wrapOutput` to be true by default, since it seems like it'll be the norm for a command to use that option.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -101,7 +101,6 @@ function ShellString(stdout, stderr, code) {
   }
   that.stderr = stderr;
   that.code = code;
-
   // A list of all commands that can appear on the right-hand side of a pipe
   // (populated by calls to common.wrap())
   pipeMethods.forEach(function (cmd) {
@@ -313,7 +312,7 @@ function wrap(cmd, fn, options) {
 
         // Perform glob-expansion on all arguments after globStart, but preserve
         // the arguments before it (like regexes for sed and grep)
-        if (!config.noglob && typeof options.globStart === 'number') {
+        if (!config.noglob && options.allowGlobbing === true) {
           args = args.slice(0, options.globStart).concat(expand(args.slice(options.globStart)));
         }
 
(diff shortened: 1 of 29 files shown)


src/common.js
```diff
@@ -361,8 +361,18 @@ function _readFromPipe(that) {
 }
 exports.readFromPipe = _readFromPipe;
 
+var DEFAULT_WRAP_OPTIONS = {
+  allowGlobbing: true,
+  canReceivePipe: false,
+  cmdOptions: false,
+  globStart: 1,
+  wrapOutput: false,
```
ariporad: I vote for this to default to true.

nfischer:
This now also makes `wrapOutput` default to true, and refactors accordingly.

nfischer:
Gonna give this an LGTM since it just covers the refactors discussed. @ariporad chime in if there are any other issues.

nfischer:
The only difference, I suppose, is that this removes the exception thrown if `pipeOnly` and `canReceivePipe` conflict (since it didn't work as well when specifying default options). We can provide documentation instead saying that `pipeOnly` overrides `canReceivePipe` (which is probably also a reasonable behavior).

gyandeeps:
Upgrade from `0.7.2` to `0.7.3` broke `.test()` function since it has `allowGlobbing` to `true` now. 
* Was this intentional?
    * If yes, Why was this breaking change introduced with a minor/patch version?

nfischer:
@gyandeeps thanks for pointing this out. It looks like the change in behavior was a mistake. I'll review this more carefully and revert any changes to behavior.

PR #493, opened 2016-07-27
refactor: allow pipeOnly commands (methods on ShellStrings)

### LINKED ISSUE #487: Add a way to create commands that can receive from a pipe without being standalone commands
This is kind of a weird use-case, but it might be nice to allow plugin writers to create commands of the following form (`tr()` is the example):

```js
var ret = shell.cat('file.txt').tr('-d', 'A'); // ShellStrings have a .tr() method
assert.equal(shell.hasOwnProperty('tr'), false); // The shell.tr() command doesn't exist
```

This example would stay consistent with how the unix `tr` command works, since it doesn't actually take a filename as an argument and can only be used via pipes.

This is related to #486, but I think there's an important distinction. This still takes advantage of option parsing and perhaps other features of the plugin API. #486 is more about adding pure methods without any sort of wrapping, or completely changing the implementation of ShellStrings.

This functionality could be offered as an option to `plugin.register`:

```js
plugin.register('tr', _tr, {
  wrapOutput: true,
  cmdOptions: {
    d: 'delete',
  },
  canReceivePipe: true,
  pipedOnly: true, // perhaps there's a way to tie this in to the option above
});
```

nfischer:
Fixes #487 

This implements the `pipeOnly` option to `wrapOptions`. This is false by default.

This checks for conflicts between `pipeOnly` and `canReceivePipe` and throws an exception.

This also refactors some code to take advantage of these new changes. Two test cases were modified in order to give better error messages (in case globbing ever breaks for `.to()` or `.toEnd()`).

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -68,10 +68,11 @@ require('./src/head');
 //@include ./src/tail
 require('./src/tail');
 
-// The below commands have been moved to common.ShellString(), and are only here
-// for generating the docs
 //@include ./src/to
+require('./src/to');
+
 //@include ./src/toEnd
+require('./src/toEnd');
 
 //@include ./src/sed
 require('./src/sed');
--- a/src/common.js
+++ b/src/common.js
@@ -6,10 +6,9 @@ var os = require('os');
 var fs = require('fs');
 var glob = require('glob');
 var shell = require('..');
-var _to = require('./to');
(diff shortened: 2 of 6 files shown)


ariporad:
LGTM!

ariporad:
Force-merging this because LGTM isn't working.

Issue #480, opened 2016-07-24
exec doesnt seem to be working


jeremyrajan:
### Node version:

6.2.0
### ShellJS version

0.7.0
### Operating system:

Mac
### Description of the bug:

I am trying to run the following command (that I copied from docs) in my electron console.

![image](https://cloud.githubusercontent.com/assets/2890683/17082963/b3d787e2-51ad-11e6-8399-490ae148b18d.png)

Any idea, why this would be having. As I want to return the version number for checks. Other commands such as `which` are working but just `exec`

Thank you! :)

nfischer:
@ariporad can you reproduce this? This seems to be related to #471, and seems to be Mac-specific

ariporad:
@jeremyrajan: Can you expand the error messages to see the full stacks? Thanks!

jeremyrajan:
Hi @ariporad , apologies for delay. Following is the full stack trace:

```javascript
exec('node --version', {silent:true})

/Users/DextorG/Documents/Projects/Github/installer_app/node_modules/shelljs/src/com…:40 

exec: internal errorlog @ /Users/DextorG/Documents/Projects/Github/installer_app/node_modules/shelljs/src/com…:40

error @ /Users/DextorG/Documents/Projects/Github/installer_app/node_modules/shelljs/src/com…:68

_exec @ /Users/DextorG/Documents/Projects/Github/installer_app/node_modules/shelljs/src/exe…:256

(anonymous function) @ /Users/DextorG/Documents/Projects/Github/installer_app/node_modules/shelljs/src/com…:269

(anonymous function) @ VM345:1
```

Thank you :)

jeremyrajan:
Any idea guys? :). Neither of the exec commands are working for me

nfischer:
The only workaround I know might be `process.version` for the version. Doesn't fix this issue though.

ariporad:
Hmm...

We do seem to have a lot of problems with electron, don't we?

@jeremyrajan: Will you try this with the `fix/480` branch, and let me know what the output is? Thanks.

giacgbj:
@ariporad Do you need any further information?

Mac OS X 10.11.6

e.message
```
[...]

PR #502, opened 2016-08-05
chore: update issue template

nfischer:
Might help with cases like #480, where ShellJS is being used in a framework on top of NodeJS.


ariporad:
LGTM!

PR #503, opened 2016-08-06
feat: add overWrite option for commands

nfischer:
Commands are no longer overwritten by default (probably safer and smarter), but there is an option to allow that behavior if necessary (like if someone hypothetically writes a better `find()` command to replace the builtin one).


ariporad:
This looks great! One thing though, should this maybe throw an error? (Or at least log something?)

nfischer:
Done.

ariporad:
LGTM!

PR #504, opened 2016-08-06
chore: switch to eslint

nfischer:
Closes #317 

This switches us to using eslint instead of jshint, based off the airbnb style guide for es5.

**Note**: Most of the changes were created by `eslint --fix`.

I disabled linting rules whenever I encountered too many places where they would need fixing. Here are the rules I disabled (which we can go back and re-enable at our convenience):

```
    "quote-props": 0,
    "no-underscore-dangle": 0,
    "max-len": 0,
    "no-use-before-define": 0,
    "no-empty": 0,
    "no-else-return": 0,
    "no-throw-literal": 0,
    "newline-per-chained-call": 0,
    "consistent-return": 0,
    "no-mixed-operators": 0,
    "no-prototype-builtins": 0,
```

Any other refactors that I did manually should be mostly semantically equivalent to the way the code was before. Feel free to take your time reviewing, since there's the potential for things to change.

 - Any `for (key in obj)` loop was changed to `Object.keys(obj).forEach(function (key) {` (which I believe is slightly different semantically, but shouldn't have an effect on the intended usage).
 - I fixed some of the [`no-else-return`](http://eslint.org/docs/rules/no-else-return) errors, but ultimately disabled the rule.
 - I changed the nested ternary expression in `src/uniq.js` to be an else-if ladder for clarity.
 - I fixed some of the `consistent-return` errors, but ultimately disabled the rule (since `common.error()` actually does implicitly return a value).
 - I removed unnecessary character escaping (like `\@` in `generate-docs.js`).
 - I added a `default` case to a switch statement in `chmod()` because it didn't have one.

Make sure there's no semantic changes in the test files (I don't remember doing any intentionally).


src/chmod.js
```diff
@@ -4,31 +4,31 @@ var path = require('path');
...
 
-    OWNER_EXEC  : base.EXEC << 6,
-    OWNER_WRITE : base.WRITE << 6,
-    OWNER_READ  : base.READ << 6,
+    OWNER_EXEC: base.EXEC << 6,
+    OWNER_WRITE: base.WRITE << 6,
+    OWNER_READ: base.READ << 6,
 
     // Literal octal numbers are apparently not allowed in "strict" javascript.  Using parseInt is
     // the preferred way, else a jshint warning is thrown.
```
ariporad: Two things:

1. It might be good to note that they've been added back in ES6, for future reference.
2. Maybe change the reference to jshint to eslint?

src/chmod.js
```diff
@@ -124,7 +122,7 @@ function _chmod(options, mode, filePattern) {
     if (isNaN(parseInt(mode, 8))) {
       // parse options
       mode.split(',').forEach(function (symbolicMode) {
-        /*jshint regexdash:true */
+        /* jshint regexdash:true */
```
ariporad: We should get rid of this.

src/common.js
```diff
@@ -1,5 +1,6 @@
 // jshint -W053
```
ariporad: We should get rid of this.

src/common.js
```diff
@@ -222,16 +223,14 @@ exports.unlinkSync = unlinkSync;
...
   function randomHash(count) {
     if (count === 1)
-      return parseInt(16*Math.random(), 10).toString(16);
-    else {
-      var hash = '';
-      for (var i=0; i<count; i++)
-        hash += randomHash(1);
-      return hash;
-    }
+      return parseInt(16 * Math.random(), 10).toString(16);
```
ariporad: Can we add `{}`'s here?

src/common.js
```diff
@@ -284,30 +284,26 @@ function wrap(cmd, fn, options) {
         //    `cp([file1, file2, file3], dest);`
         // equivalent to:
         //    `cp(file1, file2, file3, dest);`
-        args = args.reduce(function(accum, cur) {
-          if (Array.isArray(cur)) {
+        args = args.reduce(function (accum, cur) {
+          if (Array.isArray(cur))
             return accum.concat(cur);
```
ariporad: Nit: Can we either add curly braces here, or have the `return` on the same line as the `if`?

src/common.js
```diff
@@ -284,30 +284,26 @@ function wrap(cmd, fn, options) {
...
-          } else
-            return arg;
+          return arg;
         });
 
         // Expand the '~' if appropriate
         var homeDir = getUserHome();
-        args = args.map(function(arg) {
+        args = args.map(function (arg) {
           if (typeof arg === 'string' && arg.slice(0, 2) === '~/' || arg === '~')
```
ariporad: Nit: Can we either add curly braces here, or have the `return` on the same line as the `if`?

src/dirs.js
```diff
@@ -21,11 +21,9 @@ function _isStackIndex(index) {
 
 function _parseStackIndex(index) {
   if (_isStackIndex(index)) {
-    if (Math.abs(index) < _dirStack.length + 1) { // +1 for pwd
+    if (Math.abs(index) < _dirStack.length + 1) // +1 for pwd
```
ariporad: Nit: Can we either add curly braces here, or have the `return` on the same line as the `if`?

src/sed.js
```diff
@@ -29,12 +29,12 @@ function _sed(options, regex, replacement, files) {
...
   var pipe = common.readFromPipe(this);
 
-  if (typeof replacement === 'string' || typeof replacement === 'function')
-    replacement = replacement; // no-op
-  else if (typeof replacement === 'number')
-    replacement = replacement.toString(); // fallback
-  else
-    common.error('invalid replacement string');
+  if (typeof replacement !== 'string' && typeof replacement !== 'function') {
+    if (typeof replacement === 'number')
```
ariporad: Nit: Can we either add curly braces here, or have the `return` on the same line as the `if`?

src/sed.js
```diff
@@ -29,12 +29,12 @@ function _sed(options, regex, replacement, files) {
...
+      replacement = replacement.toString(); // fallback
+    else
```
ariporad: Nit: Can we either add curly braces here, or have the body on the same line as the `if`?

src/uniq.js
```diff
@@ -47,30 +47,32 @@ function _uniq(options, input, output) {
...
-                          [{count: 1, ln: e}].concat(res));
-               }, []).filter(function(obj){
-                 //Do we want only duplicated objects?
-                 return options.duplicates ? obj.count > 1 : true;
-               }).map(function(obj){
-                 //Are we tracking the counts of each line?
-                 return (options.count ? (lpad(7,obj.count) + ' ') : '') + obj.ln;
-               }).join('\n') + '\n';
+  var uniqed = lines.reduceRight(function (res, e) {
+                 // Perform uniq -c on the input
```
ariporad: This indent seems weird.

src/uniq.js
```diff
@@ -47,30 +47,32 @@ function _uniq(options, input, output) {
               trimRight().
               split(/\r*\n/);
 
-  var compare = function(a, b){
-    return options.ignoreCase ? 
+  var compare = function (a, b) {
+    return options.ignoreCase ?
              a.toLocaleLowerCase().localeCompare(b.toLocaleLowerCase()) :
```
ariporad: Is this indent right?

test/head.js
```diff
@@ -76,23 +76,23 @@ assert.equal(shell.error(), null);
...
-                     ).join('\n')+'\n');
+                     ).join('\n') + '\n');
 
 // With `'-n' <num>` option
 result = shell.head('-n', 4, 'resources/head/file2.txt', 'resources/head/file1.txt');
 assert.equal(shell.error(), null);
 assert.equal(result.code, 0);
 assert.equal(result, topOfFile2.slice(0, 4).concat(
                          topOfFile1.slice(0, 4)
-                     ).join('\n')+'\n');
```
ariporad: This indent seems weird.

ariporad:
I went through and added some comments, one thing in perticular that I think we should change is this:

```javascript
// This is OK without curly braces, because it's just a guard statement:
if (whatever) return doFoo();

// This is NOT OK, because it might be confusing in the future:
if (whatever)
    return doFoo(); // Note the newline.

// This is OK, because there is braces:
if (whatever) {
    return doFoo();
}
```

nfischer:
@ariporad do you know the eslint rule for that?

src/chmod.js
nfischer: I'll fix the reference to jshint. The part about octals was around previously, so I think we can just fix that in a new PR

nfischer:
@ariporad just updated. Your comments should be addressed. I left the section about octal numbers, since it isn't really relevant to updating linting rules.

ariporad:
LGTM!

PR #506, opened 2016-08-13
refactor: readFromPipe() requires no arguments

### LINKED ISSUE #485: readFromPipe should be a function with no arguments
[`common.readFromPipe()`](https://github.com/shelljs/shelljs/blob/master/src/common.js#L359) is a useful function for any command that can be on the right-hand-side of a pipe (i.e. `sed` in the example `grep('foo', 'file.txt').sed(/o/g, 'a')`). This will be an important part of a future plugin API.

Currently, the function must be invoked as `common.readFromPipe(this)` (always passing `this` as the only parameter). This is awkward at best, and we should be able to write in a workaround to allow it to be invoked instead as `common.readFromPipe()` (no parameters).

nfischer:
`readFromPipe()` no longer requires `this` to be passed in as the first parameter.

Fixes #485

### DIFF
--- a/src/cat.js
+++ b/src/cat.js
@@ -21,7 +21,7 @@ common.register('cat', _cat, {
 //@ containing the files if more than one file is given (a new line character is
 //@ introduced between each file).
 function _cat(options, files) {
-  var cat = common.readFromPipe(this);
+  var cat = common.readFromPipe();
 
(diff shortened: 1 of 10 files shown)


ariporad:
LGTM!

PR #509, opened 2016-08-20
refactor: replace fs.existsSync

nfischer:
Because fs.existsSync is deprecated. This internally uses `fs.statSync()` instead, wrapped in a try-catch.

### DIFF
--- a/src/cat.js
+++ b/src/cat.js
@@ -28,7 +28,7 @@ function _cat(options, files) {
   files = [].slice.call(arguments, 1);
 
   files.forEach(function (file) {
-    if (!fs.existsSync(file)) {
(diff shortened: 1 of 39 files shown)


ariporad:
LGTM!

gyandeeps:
`fs.existsSync` is undeprecated now. 😸

nfischer:
Huzzah! [This comment](https://github.com/nodejs/node/issues/1592#issuecomment-251770788) confirms this. Thanks for pointing this out. In that case, we should eventually revert this PR and use the core API again.

It may be worth it to look over where this is used in our code and see if we're falling into any anti-patterns (like checking for a file before reading from it, instead of just reading and catching the exception).

PR #511, opened 2016-08-24
fix: echo supports -e option properly

### LINKED ISSUE #510: fix: echo -e should not print "-e"
any

master

### Operating system:
any

### Description of the bug:

In bash, `echo -e` just means to interpret backslashes specially (i.e. '\n' is a newline, instead of a literal '\n'). This is the default behavior for ShellJS `echo()` (since that's how `console.log()` works), so we should document it. Also, we should allow the `-e` option, and just make it a no-op.

```javascript
> echo('-e', 'Any message'); // This should not print '-e'
-e Any message
> echo('Any message'); // The previous command should do the same thing as this command
Any message
```

nfischer:
Fixes #510 

The only real difference in behavior is that `echo('-e', message)` won't print the literal `-e` (since Bash's `echo -e` wouldn't).

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,7 +459,10 @@ Searches for `command` in the system's PATH. On Windows, this uses the
 Returns string containing the absolute path to the command.
 
 
-### echo(string [, string ...])
+### echo([options,] string [, string ...])
+Available options:
+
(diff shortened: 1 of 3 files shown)


ariporad:
LGTM!

PR #513, opened 2016-08-27
chore: add downloads per month on README

nfischer:
This shows # of downloads per month, which I think is a useful stat to display on our README.


ariporad:
LGTM!

Issue #175, opened 2014-12-19
Command can't have \n


fenuks:
exec hang when executing command with /n. Its probably because wrong escaping.
E.g. this will hang: 

``` javascript
syncExec = require('shelljs').exec
var command = "python2 -c 'print \"first line\"\nprint \"second line\"'"
out = syncExec(command2)
```

Child_process.exec works just fine.

stevemao:
Is this problem solved?

stevemao:
You should probably use `\\n`?

stevemao:
Apparently travis does not like newline `$'\n'`

https://github.com/travis-ci/travis-ci/issues/4061

Have to use `child.exec` for now :(

eiabea:
I ran into the same issue :-(

Do you have any plans to work on this issue?

mtscout6:
I tried running this locally, though my understanding of python is limited. I get the following when just running the command outside of node:

```
➜  shelljs git:(master) ✗ python -c 'print "first line"\nprint "second line"'
  File "<string>", line 1
    print "first line"\nprint "second line"
                                          ^
SyntaxError: unexpected character after line continuation character
```

So it would appear to me that your example is an invalid python example?

danasilver:
I can reproduce this:

```
var sh = require('shelljs');

sh.exec('echo "\n"');

var childProcess = child.exec('echo \"
                              ^^^^^^^^

SyntaxError: Unexpected token ILLEGAL
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:414:25)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:134:18)
    at node.js:961:3
shell.js: internal error
Error: Command failed: "/Users/dsilver/.nvm/versions/node/v4.2.1/bin/node" /var/folders/vz/x698bjq50z55rz4xpm66fy000000gn/T/shelljs_9acc94d83ef8d84fbb8c
    at checkExecSyncError (child_process.js:441:13)
    at Object.execSync (child_process.js:481:13)
    at execSync (/Users/dsilver/Documents/workspace/shelljs/src/exec.js:81:11)
    at Object._exec (/Users/dsilver/Documents/workspace/shelljs/src/exec.js:214:12)
[...]

Issue #523, opened 2016-10-08
Feature request: allow `common.error()` to optionally not insert a prefix and optionally not print to console


nfischer:
Blocked by #522 

Our current implementation of `common.error()` works well for the common case, but could be extended to work much better for edge cases. I've been working on a replacement for `exec`, but I ran into some issues with getting the error calls just right.

If you look at how `exec()` is currently implemented, its error output looks like this:

``` js
> shell.exec('git status'); // git's natural error message makes sense
fatal: Not a git repository...
> console.log(shell.error()); // but this should ideally be the same
exec:
```

We can modify `exec()` to get things perhaps a little better, but without changing `common.error()`, the best we can get is this:

``` js
> shell.exec('git status'); // Now the error message prints twice
fatal: Not a git repository...
exec: fatal: Not a git repository...
> console.log(shell.error()); // We see the real error message, but with a prefix
exec: fatal: Not a git repository...
```

I think the necessary options would be:
- `prefix: false`: don't insert the command name as a prefix (it isn't natural to see the prefix `exec:` in front of the external command's error message)
- `print: false`: don't print the error message to the console (if the external command is already writing it to the console, this doesn't really make sense)

nfischer:
Fixed in #535

Issue #522, opened 2016-10-08
API request: allow `plugin.error()` to take an options parameter


nfischer:
The current API for `plugin.error`/`common.error` is this:

```js
common.error(message, code, _continue);
```

This API has been good enough for us, but we may want to modify it to take an options parameter for more flexibility:

```js
common.error(message, options);
```

We would migrate `continue` and `code` into the options (`code` already defaults to 1 and `continue` defaults to false).

ariporad:
I support having an options object, but I think that we should make the signature `common.error(message [, code] [, options])`;

nfischer:
I thought about that as well. The issue is that sometimes you might want the default error code while also wanting to specify an option.

PR #524, opened 2016-10-09
feat: new alternative to exec

### LINKED ISSUE #103: Allow multiple string arguments for exec()

### LINKED ISSUE #143: Escaping shell arguments to exec()
First of all, I just want to say thanks for writing shelljs. It cleans up my code a lot.

I need to invoke commands with `shell.exec()`, however, those commands will include some input from external sources, to be passed as arguments to system executables. Is there a "safe" way to do this? I was hoping for the classic array syntax, like:

``` javascript
shell.exec(["ls", "-l", "/some/path"])
```

Though this doesn't seem to be implemented. It doesn't appear nodejs itself even has a shell escape function neither. Any plans on this front?

### LINKED ISSUE #495: Add support for execFile-like interface to exec
It would be great if `exec` had an interface, perhaps like the following:

``` js
var fileName = 'foo.txt';
exec('git', 'add', fileName);
```

This is great because of the following:
- We can declare that each argument is its own non-splittable word (so `fileName = 'file1.txt file2.txt'` means the git call will search for one file named `file1.txt file2.txt`, not two separate files)
- We can implement this fairly easily with `child_process.execFile`
- This would allow us to do our own glob expansion, which would be good for Windows (Windows `exec()` won't glob-expand)
- This would be useful in `shx` (shelljs/shx#65 & shelljs/shx#68)
- Variables would have to be referenced as JavaScript variables just like in the rest of ShellJS, not as shell variables (`exec('echo', env.PATH)` vs. `exec('ls $PATH')`), which is the safer approach
- Doing our own globbing means that `exec` will respect a call to `set('-f')`
- This would give more security to [shelljs-exec-proxy](https://github.com/nfischer/shelljs-exec-proxy) for the same reasons

The issue with this interface is what to do when we have only 1 argument (do we use the old behavior or the new `execFile` implementation?). An alternative approach is `exec([arg1, arg2, arg3])`, which uses a list to remove ambiguity. This is easier to parse, but perhaps not as nice to work with.

---

**Edit:** to clarify, this is a **feature request**, not a security vulnerability. This new API would be easier to use safely than `shell.exec()`, but `shell.exec()` is not itself vulnerable and can in fact be used safely without much difficulty (see [security guidelines](https://github.com/shelljs/shelljs/wiki/Security-guidelines)).

nfischer:
Add a new command to launch external shell commands, providing a less vulnerable
alternative to exec(), with cross-platform globbing.

I'd love community input on the design of the new API, if this is an agreeable name, etc. This implementation would work well with shelljs-exec-proxy (see nfischer/shelljs-exec-proxy#2), which might be a better API alternative for Node v6+ users.

I believe this, in conjunction with `set('-f')`, should solve all the relevant security flaws surrounding command injection and unexpected globbing. Globbing is enabled by default in this implementation, for the sake of providing a consistent API across all commands.

This relies on a JS-based implementation of globbing, which is an improvement on `exec()`, which inconsistently provides globbing on unix but not on windows.

Fixes #143
Fixes #495
Fixes #103 

I believe this should also help with #175, #5, shelljs/shx#68, and probably some others.

### DIFF
--- a/src/child.js
+++ b/src/child.js
@@ -1,0 +1,45 @@
+var childProcess = require('child_process');
+var fs = require('fs');
+
+var stdoutFile = process.argv[2];
+var stderrFile = process.argv[3];
+var cmd = process.argv[4];
+var argString = process.argv[5];
+var optString = process.argv[6];
+var pipe = process.argv[7];
+
+var opts = JSON.parse(optString);
+var args = JSON.parse(argString);
+
+var c = childProcess.spawn(cmd, args, opts);
+
+// Wait to exit until all other work has been finished, like closing IO streams
+function delayedExit(code) {
+  setTimeout(function () {
+    process.exit(code); // let streams close before ending
+  }, 0);
+}
+
+c.on('error', function () {
+  delayedExit(127);
+});
+
+c.on('exit', function (code) {
+  delayedExit(code);
+});
+
+var stdoutStream = fs.createWriteStream(stdoutFile);
+var stderrStream = fs.createWriteStream(stderrFile);
+c.stdout.pipe(stdoutStream, { end: false });
+c.stderr.pipe(stderrStream, { end: false });
+
+if (!opts.silent) {
+  c.stdout.pipe(process.stdout);
+  c.stderr.pipe(process.stderr);
+}
+
+c.stdout.on('end', stdoutStream.end);
+c.stderr.on('end', stderrStream.end);
+
+// Handle piped input
+if (pipe) c.stdin.end(pipe);
--- a/src/child.js
+++ b/src/child.js
@@ -13,19 +13,12 @@ var args = JSON.parse(argString);
 
 var c = childProcess.spawn(cmd, args, opts);
 
-// Wait to exit until all other work has been finished, like closing IO streams
-function delayedExit(code) {
-  setTimeout(function () {
-    process.exit(code); // let streams close before ending
-  }, 0);
-}
-
 c.on('error', function () {
-  delayedExit(127);
+  process.exitCode = 127;
 });
 
 c.on('exit', function (code) {
-  delayedExit(code);
+  process.exitCode = code;
 });
 
 var stdoutStream = fs.createWriteStream(stdoutFile);
--- a/README.md
+++ b/README.md
@@ -171,6 +171,30 @@ Notable exceptions:
 + There is no "quiet" option since default behavior is to run silent.
 
 
+### cmd(command [, arguments...] [, options] [, callback])
+Available options (all `false` by default):
+
++ `silent`: Do not echo program output to console.
++ and any option available to NodeJS's
+  [child_process.spawnSync()](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options)
+
+Examples:
+
+```javascript
+var version = cmd('node', '--version', {silent:true}).stdout;
+
+cmd('git', 'checkout', 'master', '--', 'file.js');
+```
+
+An alternative for [exec()](#execcommand--options--callback), with better
+security around globbing, comamnd injection, and variable expansion. This is
+guaranteed to only run one external command, and won't handle special
+characters in unexpected and unsafe ways.
+
+By default, this performs globbing on all platforms (but you can disable
+this for extra security using `set('-f')`).
+
+
 ### cp([options,] source [, source ...], dest)
 ### cp([options,] source_array, dest)
 Available options:
--- a/commands.js
+++ b/commands.js
@@ -2,6 +2,7 @@ module.exports = [
   'cat',
   'cd',
   'chmod',
+  'cmd',
   'cp',
   'dirs',
   'echo',
--- a/src/cmd.js
+++ b/src/cmd.js
@@ -1,0 +1,136 @@
+var common = require('./common');
+var _tempDir = require('./tempdir');
+var path = require('path');
+var fs = require('fs');
+var child = require('child_process');
+
+common.register('cmd', _cmd, {
+  unix: false,
+  canReceivePipe: true,
+  wrapOutput: false,
+});
+
+// Similar to shell.exec(), this is a hack so that we can get concurrent output.
+function cmdSync(cmd, args, opts, pipe) {
+  var tempDir = _tempDir();
+  var stdoutFile = path.join(tempDir, common.randomFileName());
+  var stderrFile = path.join(tempDir, common.randomFileName());
+
+  opts = common.extend({
+    silent: common.config.silent,
+    cwd: process.cwd(),
+  }, opts);
+
+  if (common.existsSync(stdoutFile)) common.unlinkSync(stdoutFile);
+  if (common.existsSync(stderrFile)) common.unlinkSync(stderrFile);
+
+  // resolve to an absolute path, so if we cd in the child process, it's a no-op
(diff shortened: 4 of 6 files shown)


nfischer:
@ariporad Please go ahead and start reviewing. The fixes for the test failures should be pretty trivial (some of them are with the tests themselves). This way I can knock out all the fixes at once.

ariporad:
LGTM!

This is really great, I think that we should consider building a new exec on top of it.

nfischer:
I'll see if I can figure out why this is failing on unix. This will never pass on v0.10, so I think it's time to remove that from Travis (the version already wasn't supported by us).

This is currently blocked by #525 #523 #522 and shelljs/shx#81, so help with those issues would be appreciated.

detailyang:
Hi, @nfischer.

what's time your PR will be merged into master?

nfischer:
@detailyang thanks for the follow up. I'll try to put in some work on this PR this week. I *believe* most of the blocking issues are now resolved.

This just needs to be rebased on top of master and the tests need to be rewritten to use ava.

nfischer:
This is now rebased and tests are refactored.

This is currently blocked on #633, and I need to check that I've properly taken advantage of the PRs that this was previously blocked by.

nfischer:
TODO list (in no particular order):

 - [ ] possibly rename `realtimeOutput` to `interactive` (taken from [this comment](https://github.com/shelljs/shelljs/pull/432#issuecomment-273774783))
 - [ ] fix TODOs in the code
 - [ ] get rid of `nodeBinPath` (this was taken care of in #633)
 - [ ] make sure it handles interactive input nicely
 - [ ] clean up commented code
 - [ ] deprecate `exec`
 - [ ] rebase off `dev`
 - [ ] merge into `dev`
 - [ ] fix skipped tests
 - [ ] write tests for interactive output (maybe check that `ls --color=auto` prints colored output)
 - [ ] decide whether or not to return control characters in the return properties (like colors inside `result.stdout`)

zhiqunq:
Hi, @nfischer.

what's time your PR will be merged into master?

nfischer:
@zhiqunq this is still slated for v0.8. I haven't had time to pick it back up. I'm waiting to wrap this up so that I can be confident it fixes all of the major issues with `exec`. If you'd like to help, feel free to test this out and see if it meets your needs or if it has any issues.

whitecolor:
Hope to see this soon!

nfischer:
Slight update: we're [full-speed ahead on v0.8](https://github.com/shelljs/shelljs/pull/731). I just rebased PR off the latest master, so I can hopefully dedicate more focus to this.

I want to make sure we get this right before landing, so this might get bumped back to v0.9 if we want v0.8 released sooner. This will hopefully get finished up soon though.

jacquesdev:
Hi @nfischer - thanks for your work on this - appreciate it. Do you have any sort of a timeline/idea of when this will be released?

nfischer:
Unfortunately, I don't have a whole lot of time for OSS right now, so this will probably take a month or two.

src/child.js
```diff
@@ -0,0 +1,45 @@
...
+
+var opts = JSON.parse(optString);
+var args = JSON.parse(argString);
+
+var c = childProcess.spawn(cmd, args, opts);
+
+// Wait to exit until all other work has been finished, like closing IO streams
+function delayedExit(code) {
+  setTimeout(function () {
+    process.exit(code); // let streams close before ending
```
bruce-one: Should this use [`process.exitCode = code`](https://nodejs.org/api/process.html#process_process_exitcode) instead?

Based on the warning from https://nodejs.org/api/process.html#process_process_exit_code

    It is important to note that calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr.

(Which seems to imply it might not actually let the streams close/flush correctly?)

nfischer: This is better, thanks!

On second thought, I'm not completely sure if `delayedExit` would've worked. It posts a timeout to work around this, but that might not actually work if streams are buffered after the process finishes.

shaunbaruth:
@nfischer Thanks for your work on this. It's a very useful feature. Looking forward to it.

brennoo:
@nfischer thanks for good work here, we still good to have it in v0.8 ?

nfischer:
Update:

I'm trying to harden `shell.exec()` a bit for the v0.8 release (#782). It'll have some (not all) of the improvements from this PR (source code script instead of writing to a file, reducing file IO), but the goal is a backwards-compatible API.

I almost have that working, and it's cleaning up quite a bit of code. After we land that, there should be a little bit of clean-up work, and then it should be easy to revisit this PR (and build a secure-by-default API!).

Thanks for your patience!

nfischer:
Bumping this back to v0.9. I want to release the hardened exec changes under v0.8 (still a minor security improvement). This will give us a chance to fix any bugs, and move forward with the new API.

ev1stensberg:
Any update for this?

nfischer:
Updates:

 * I think #782 is wrapped up (I need to double check).
 * We want to do our v0.8 release without this new API, but including the improvements from #782. This will give us a chance to battle test the changes and make sure there are no issues.
 * After that, this PR can become a lot smaller and more manageable--and we can focus on merging it and pushing a release

nfischer:
Ok, so v0.8 is released! If you can, please try out the release, let us know if there are any new issues with `exec`. I'm going to dedicate time over the next couple weeks on this PR.

igrayson:
We're looking forward to switching over to this more secure execution method! Any update?

nfischer:
@igrayson I'm currently finishing up a local prototype. I will post a design doc in the next 2-ish weeks and upload a PR around the same time. Sorry this has dragged on so long, but this is indeed still on my radar (and thanks for asking about progress).

nfischer:
Thank you all for your patience. I'm currently polishing up my prototype (it is going to take a slightly different approach from this PR). I've written a [design doc](https://docs.google.com/document/d/1UFm10TONaNWok3aEPzUP_OjZ6lEvwlYqyJBUcugLfso/edit?usp=sharing), and I invite you all to comment on the design (in particular, please check the proposed design will work for your use case).

Next steps:

 * upload my prototype in a new PR, close this one
 * land the initial prototype, and implement the extra "modes" mentioned in the design doc in further PRs
 * the first shelljs release with shell.cmd() will be the first release we deprecate shell.exec()

nfischer:
New PR is up at #866! Closing this PR (and the other PR should be in a much more landable state than this one).

nfischer:
If you've found this PR, please **do not** consider this a security fix. ShellJS does not have a security "vulnerability" to resolve, we have a poorly-designed API (`shell.exec()`) that must be used with care to be considered secure. We've updated documentation to warn against improper usage and have provided [security guidelines](https://github.com/shelljs/shelljs/wiki/Security-guidelines) to help dependent modules use this safely. Please consult #945 for more information.

---

This PR is an obsolete attempt to provide a better API, but such a better API is not required for your modules to be secure.

PR #527, opened 2016-10-11
feat: cp -u option

### LINKED ISSUE #526: Add -u flag support for cp
Hello!

It would be great if someone could add -u flag support for cp command.

Right now output for shx is following:

cp: option not recognized: u

Thanks in advance.

nfischer:
Only update when the source file is older than the destination, or the
destination file is missing.

Fixes #526


ariporad:
LGTM!

PR #530, opened 2016-10-17
Fix: Remove default glob from shell.test (fixes #529)

### LINKED ISSUE #529: Inadvertent breaking change to shell.test()
See the discussion in #492 

`shell.test()` used to not perform shell globbing (it's not usually something you would want in this scenario). The above-mentioned PR accidentally introduced globbing to the command, which could break things for some users.

If anyone can examine that PR and figure out all behavior changes, feel free to submit a PR for this.

gyandeeps:
Regression from #492

### DIFF
--- a/src/test.js
+++ b/src/test.js
@@ -13,6 +13,7 @@ common.register('test', _test, {
     'S': 'socket',
   },
   wrapOutput: false,
+  allowGlobbing: false,
 });
 
 
--- a/test/test.js
+++ b/test/test.js
@@ -65,6 +65,12 @@ result = shell.test('-L', 'resources/file1');
 assert.equal(shell.error(), null);
 assert.equal(result, false);
 
+// regression #529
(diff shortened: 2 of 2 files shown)


nfischer:
LGTM. Thanks for the fix. @gyandeeps did you find any other commands that had a similar mistake? If not, I can merge this tomorrow.

gyandeeps:
This s the only I was sure about.

nfischer:
I can't find any other regressions. This should be great, thanks for catching this!

gyandeeps:
@nfischer When is the plan to make a release with this fix? Just wanted to know. Thanks

nfischer:
I need to take a look at our recent commits, but I think we can get this out as a patch release pretty soon. I also want to push out some plugin-related features to begin dogfooding them, but I should be wrapping those up soon. We can probably get this out within the week.

PR #534, opened 2016-10-20
New: `nodepath` option for the path

gyandeeps:
This what i am thinking for `nodepath` config option. This works as i have tried it. 
Its getting difficult to test since when you run `npm test` to run the unit test. NPM already adds the `./node_modules/.bin` to the $PATH. 

I will work on the tests but I wanted to know whether I am heading in the right direction.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -4,10 +4,22 @@
 
 var os = require('os');
 var fs = require('fs');
+var path = require('path');
 var glob = require('glob');
 var shell = require('..');
 
 var shellMethods = Object.create(shell);
+var PATH_IDENTIFIER = 'PATH';
+
+// windows calls it's path 'Path' usually, but this is not guaranteed.
+if (process.platform === 'win32') {
+  PATH_IDENTIFIER = 'Path';
+  Object.keys(process.env).forEach(function (e) {
+    if (e.match(/^PATH$/i)) {
+      PATH_IDENTIFIER = e;
+    }
+  });
+}
 
 // Module globals
 var config = {
@@ -16,7 +28,8 @@ var config = {
   verbose: false,
   noglob: false,
   globOptions: {},
-  maxdepth: 255
+  maxdepth: 255,
+  nodepath: false,
 };
 exports.config = config;
 
@@ -401,3 +414,15 @@ function _register(name, implementation, wrapOptions) {
   }
 }
 exports.register = _register;
+
+exports.getProcessEnv = function () {
+  var env = objectAssign({}, process.env);
+
+  if (config.nodepath) {
+    env[PATH_IDENTIFIER] += path.delimiter + path.join(process.cwd(), 'node_modules', '.bin');
+  }
+
+  return env;
+};
+
+exports.PATH_IDENTIFIER = PATH_IDENTIFIER;
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "eslint-config-airbnb-base": "^3.0.0",
     "eslint-plugin-import": "^1.11.1",
     "coffee-script": "^1.10.0",
+    "sinon": "^1.17.6",
     "shelljs-changelog": "^0.2.0",
     "shelljs-release": "^0.2.0",
     "travis-check-changes": "^0.2.0"
--- a/src/exec.js
+++ b/src/exec.js
@@ -29,7 +29,7 @@ function execSync(cmd, opts, pipe) {
   opts = common.extend({
     silent: common.config.silent,
     cwd: _pwd().toString(),
-    env: process.env,
+    env: common.getProcessEnv(),
     maxBuffer: DEFAULT_MAXBUFFER_SIZE
   }, opts);
 
@@ -171,7 +171,7 @@ function execAsync(cmd, opts, pipe, callback) {
   opts = common.extend({
     silent: common.config.silent,
     cwd: _pwd().toString(),
-    env: process.env,
+    env: common.getProcessEnv(),
     maxBuffer: DEFAULT_MAXBUFFER_SIZE
   }, opts);
 
--- a/src/which.js
+++ b/src/which.js
@@ -40,7 +40,7 @@ function checkPath(pathName) {
 function _which(options, cmd) {
   if (!cmd) common.error('must specify command');
 
-  var pathEnv = process.env.path || process.env.Path || process.env.PATH;
+  var pathEnv = common.getProcessEnv()[common.PATH_IDENTIFIER];
   var pathArray = splitPath(pathEnv);
   var where = null;
 
--- a/test/exec.js
+++ b/test/exec.js
@@ -4,6 +4,9 @@ var assert = require('assert');
 var util = require('util');
 var path = require('path');
 var os = require('os');
+var sinon = require('sinon');
+var sandbox = sinon.sandbox.create();
+var common = require('../src/common');
 
 shell.config.silent = true;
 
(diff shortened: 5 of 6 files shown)


src/common.js
```diff
@@ -4,11 +4,23 @@
...
 var fs = require('fs');
+var path = require('path');
 var glob = require('glob');
 var shell = require('..');
 
 var DEFAULT_ERROR_CODE = 1;
 var shellMethods = Object.create(shell);
+var PATH = 'PATH';
+
+// windows calls it's path 'Path' usually, but this is not guaranteed.
```
nfischer: I believe we can rely on `process.env` to have case-insensitive keys, so `process.env.PATH` should always work.

`process.env` isn't a real JS object. See atom/atom#11302 as an example of this.

src/common.js
```diff
@@ -17,7 +29,8 @@ var config = {
   verbose: false,
   noglob: false,
   globOptions: {},
-  maxdepth: 255
+  maxdepth: 255,
+  nodepath: false
```
nfischer: Could you add a trailing comma? We don't enforce it via linting, but it's something we're trying to move toward to keep diffs simple.

src/common.js
```diff
@@ -397,3 +410,17 @@ function _register(name, implementation, wrapOptions) {
   }
 }
 exports.register = _register;
+
+exports.getProcessEnv = function () {
+  var env = objectAssign({}, process.env);
+
+  if (config.nodepath) {
+    env[PATH] += (process.platform === 'win32' ? ';' : ':') + path.join(process.cwd(), 'node_modules', '.bin') + (process.platform === 'win32' ? ';' : ':');
```
nfischer: Could you change this to [path.delimiter](https://nodejs.org/api/path.html#path_path_delimiter)?

Also, I don't think we always want a trailing delimiter, so you can get rid of the second `(process.platform === 'win32' ? ';' : ':')`. Please correct me if Windows differs from Unix on this point.

src/common.js
```diff
@@ -397,3 +410,17 @@ function _register(name, implementation, wrapOptions) {
...
+  }
+
+  return env;
+};
+
+exports.getPathFromEnv = function (env) {
```
nfischer: I don't think we need this, since we can just use `process.env.PATH` to retrieve the value.

src/common.js
gyandeeps: I saw `npm` doing this that why I did it to make sure we are cross platform

src/common.js
gyandeeps: At some places in the code base we have used `process.env.path || process.env.Path || process.env.PATH;` to get the correct path. I just made a function to keep it at one spot.

nfischer: Yeah, I saw that. I think that wasn't actually necessary for us to do. We should probably convert everything to `process.env.PATH` for the sake of consistency.

Plus, we would run into issues if someone on unix ever did `export path=foo` (totally valid for a user to do, but it would provide incorrect results for `which()`).

gyandeeps: I am hearing get rid of this change and the above logic and just make it use `process.env.PATH`, correct? 
Will make the change then.

src/common.js
nfischer: Hmmm could you link me to the line in npm? I know that `process.env` is case-insensitive on my Windows machine. Perhaps it's not the case for everyone?

I would imagine things would get very broken in NodeJS if `process.env` were case-sensitive on Windows. Perhaps that's the case for older versions of Node?

gyandeeps: https://github.com/npm/npm/blob/master/lib/utils/lifecycle.js#L20-L28

src/common.js
nfischer: Yeah. Let's try that. If things break on Appveyor, then we can revert back to this.

src/common.js
gyandeeps: thumbsup meansd keep the change or remove it? because this change is tied with the below function which we are talking abt also.

nfischer: Hmmm interesting... I asked a clarifying question on nodejs/node#9157. They should be able to confirm if we should rely on the case-insensitivity or not.

gyandeeps: Will wait till tomorrow to see if we get a response. Will update Morrow then. Thanks

gyandeeps: @nfischer I saw the reply but dint quite understand. Can you plz decipher that let me know whether I should go with this strategy or remove the above logic. thanks

nfischer: I posted a follow up comment. It looks to me like they're making a PR to add this to the documentation (I didn't read the PR yet, so I could be mistaken). This seems to indicate that this is the intended behavior on Node on Windows.

My *guess* is that `process.env` has always had case-insensitive keys on Windows, but it was just never documented. I think it's ok to rely on that (unless we hear otherwise).

gyandeeps: @nfischer let me know whatever you thoughts are on this and I will react accordingly. Thanks

nfischer: @gyandeeps let's go for the simpler approach, `process.env.PATH`. Thanks

nfischer:
Let's add a test in `exec()` and in `which()`

src/common.js
gyandeeps: Apparently this is not true. I use windows 10 and when I use windows default cmd then its called `Path` but if I run using git bash then its called `PATH`.
Thats why I see test fail on windows.

gyandeeps:
@nfischer All the work has been done. Some of things to note:

1. Apparently `PATH` is not same across. I use windows 10 and when I use windows default cmd then its called Path but if I run using git bash then its called PATH. Thats why I see test fail on windows.
1. I have added `sinon` so that we can fake `process.env.PATH` for testing purposes.
1. I think we should soon move to using some kind of testing framework since its unit test are not well managed.

src/common.js
nfischer: @gyandeeps if this is not the case, then it would be good to chime in on nodejs/node#9157 and the corresponding pull request before Node 7 gets incorrect documentation.

As for this PR, let's go with your initial approach. Sorry for the confusion.

nfischer:
@gyandeeps 

1. See my comment in-line
2. `sinon` is fine
3. Please see #512. I'll try to rebase off master sometime this week. There's still quite a bit of work to do, however. If you're interested in helping, feel free to make a PR against that branch.

gyandeeps:
Thanks @nfischer . I think PR is ready. Let me know if you have any suggestions.

nfischer:
@gyandeeps I'll take a look later today. If you can, please chime in on nodejs/node#9157, since that just added got merged into official node documentation, and it disagrees with your observations on `process.env` case-sensitivity. It would be bad if the documentation was incorrect but got merged anyway.

src/common.js
```diff
@@ -4,10 +4,22 @@
...
+var path = require('path');
 var glob = require('glob');
 var shell = require('..');
 
 var shellMethods = Object.create(shell);
+var PATH_IDENTIFIER = 'PATH';
+
+// windows calls it's path 'Path' usually, but this is not guaranteed.
+if (process.platform === 'win32') {
+  PATH_IDENTIFIER = 'Path';
```
nfischer: Is this line necessary? It'll get overridden in the loop below.

nfischer:
@gyandeeps Sorry for the late reply. I got a bit distracted with #512 

Here's my thoughts on this PR:

`shell.env` (links to `process.env`) is already part of our API, so I'm concerned about moving away from that to a custom function. It feels a little clunky to use, especially with `common.PATH_IDENTIFIER`. I wanted to get your feedback on a slightly different approach.

Instead of having code to get `process.env.PATH` (regardless of how `PATH` is capitalized), what if we just use `process.env.PATH = process.env.PATH || process.env.Path || process.env.path`? Then we don't need a function or PATH_IDENTIFIER, we can just use `process.env.PATH`.

We can write a setter for `config.nodepath` to add (or remove) `node_modules/.bin` to the PATH.

Let me know what your thoughts are for this approach. The downside is that this modifies a global, but the option is likely not going to be used often, and I can't think of a use for this option where modifying the PATH that way would be dangerous.

Although admittedly, that approach wouldn't be anything the users wouldn't be able to do themselves, so perhaps it's best to leave the code as-is.

gyandeeps:
My thoughts:

* We can definitely change the `PATH` approach and do what ever you said.
* Why I create a copy of path is because I dont want to change the path inside the whole nodejs process. Since its single process, me changing the path would/might impact other things in the environment. That was my thought process when designing this.

nfischer:
@gyandeeps sorry for taking so long to get back. Let's hold off on this, since I'm not completely sure what the best approach is, or if this is something we should delegate to ShellJS users instead.

I'm marking this as something to look into once we get closer to v0.8.

@freitagbr feel free to chime in with your thoughts on this feature.

nfischer:
After some thought, this is probably the wrong approach to solve this problem. I think it's better to leave this to users, since it can easily be solved outside of ShellJS.

PR #535, opened 2016-10-24
feat: plugin.error() takes an options parameter

### LINKED ISSUE #522: API request: allow `plugin.error()` to take an options parameter

nfischer:
This implements the following options: `continue`, `code`, `prefix`, & `silent`.

Fixes #522, #523

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -7,7 +7,6 @@ var fs = require('fs');
 var glob = require('glob');
 var shell = require('..');
 
-var DEFAULT_ERROR_CODE = 1;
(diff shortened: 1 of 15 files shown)


nfischer:
LGTM

PR #543, opened 2016-11-02
Fix lint warning

### LINKED ISSUE #542: Lint warning
6.9.1
0.7.5
### Operating system:
OS X 10.11.6
### Description of the bug:
There's a lint warning on line 50 of `src/rm.js`

```
$ npm run lint

> shelljs@0.7.5 lint /Users/freitagb/src/javascript/shelljs
> eslint .

/Users/freitagb/shelljs/src/rm.js
  50:5  warning  Unexpected constant condition  no-constant-condition

✖ 1 problem (0 errors, 1 warning)
```

freitagbr:
Fixes #542

### DIFF
--- a/src/rm.js
+++ b/src/rm.js
@@ -47,6 +47,8 @@ function rmdirSyncRecursive(dir, force) {
   try {
     // Retry on windows, sometimes it takes a little time before all the files in the directory are gone
     var start = Date.now();
+
+    // TODO: replace this with a finite loop
     for (;;) {
       try {
         result = fs.rmdirSync(dir);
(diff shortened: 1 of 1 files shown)


nfischer:
@freitagbr Could you add a TODO for refactoring the infinite loop? We haven't received any bugs about it hanging, so we don't have to fix it right away, but it would be nice to eventually refactor so that we're confident it will never be a problem.

freitagbr:
Ok, will do.

nfischer:
LGTM. Feel free to merge when CI passes

freitagbr:
I think you'll have to do that, I don't have write access.

PR #545, opened 2016-11-02
Make the lint script more user-friendly

### LINKED ISSUE #544: User-friendly lint command
The `lint` script will cause npm to echo a verbose error message if there is a lint error. The lint script should be changed so that npm will not echo the error message.

freitagbr:
Fixes #544

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,10 +24,10 @@
   "homepage": "http://github.com/shelljs/shelljs",
   "main": "./shell.js",
   "scripts": {
-    "posttest": "npm run lint",
+    "posttest": "eslint .",
     "test": "node scripts/run-tests",
     "gendocs": "node scripts/generate-docs",
-    "lint": "eslint .",
+    "lint": "eslint . || true",
     "after-travis": "travis-check-changes",
     "changelog": "shelljs-changelog",
     "release:major": "shelljs-release major",


nfischer:
@freitagbr doesn't this mean that `npm run lint` looks like it succeeds, even when it actually fails? It'll show all error output, but I think it will still exit with status 0, which sounds confusing to me.

freitagbr:
Yes, that is correct. I think, in general, that `npm run lint` ought to be used to quickly check code for lint errors. If exiting with a 0 is needed, you could use `eslint .` instead.

nfischer:
I would argue that `eslint .` is a better way to quickly check for lint errors, minus the verbose output. I think it's better if `npm run lint` serves as a way of validating if the code passes our linting standards. How does that work for you?

freitagbr:
I can see your side of the argument, though here are the advantages to my method:
- If I don't have eslint installed globally, I have to type `node_modules/.bin/eslint .`
- If my global eslint is a different version than what shelljs has, then there may be different errors reported for the same set of rules. We can save confusion by always using the correct version of eslint.
- If linting is required as part of another script (for example, `posttest`), might as well just use `eslint .` It also prevents a nested invocation of npm.

nfischer:
Sure, but does this work on Windows? The `||` should, but I don't know if `true` is

freitagbr:
Oh, good point. I'll look into it.

freitagbr:
I'm going to close this. The only cross-platform way to prevent npm from displaying warnings is to use the `--silent` flat, which the user can use on their own.

PR #546, opened 2016-11-02
Safely exit by throwing an error

### LINKED ISSUE #483: Don't kill the node process upon unexpected error
#64 and #473 are examples of complaints about this.

Currently, ShellJS will [kill the process]() if it encounters an unexpected ("internal") error. Ideally, the commands should all catch these before they arise, but since nothing is ever perfect so sometimes utilities fail for various reasons (permissions errors are a big reason).

It'd be great if ShellJS threw an exception instead of killing the process, since this would at least allow users the chance to catch the exception in their code. If the exception is left uncaught at the top-level, this would kill the process anyway as it currently does. This could have significantly different behavior if thrown inside of a try-catch, however, so this would be a breaking change.

freitagbr:
Fixes #483.

Instead of exiting the process with `process.exit(1)`, simply throw an error. This way, the error can be caught by Node itself. If the error is not caught, the error is still printed, and the process exits with a `1`.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -348,8 +348,7 @@ function wrap(cmd, fn, options) {
       if (!state.error) {
         // If state.error hasn't been set it's an error thrown by Node, not us - probably a bug...
         console.error('ShellJS: internal error');
-        console.error(e.stack || e);
-        process.exit(1);
+        throw e;
       }
       if (config.fatal) throw e;
     }
--- a/src/common.js
+++ b/src/common.js
@@ -347,7 +347,7 @@ function wrap(cmd, fn, options) {
     } catch (e) {
       if (!state.error) {
         // If state.error hasn't been set it's an error thrown by Node, not us - probably a bug...
-        e.message = 'ShellJS Internal Error: ' + e.message;
+        e.name = 'ShellJSInternalError';
         throw e;
       }
       if (config.fatal) throw e;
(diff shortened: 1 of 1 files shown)


nfischer:
Is there a way to make the message be part of the new stack trace? That would be even better.

freitagbr:
Do you mean the `'ShellJS: internal error'` message?

nfischer:
> Do you mean the 'ShellJS: internal error' message?

Yeah, that's what I meant

nfischer:
@freitagbr is it standard to prepend to the message like that?

freitagbr:
Probably not, so I will suggest a different option: changing the message name to `ShellJSInternalError`.

nfischer:
@freitagbr that could work

freitagbr:
I toyed around with the idea of creating an internal error class, but in this instance, I think just changing the name should suffice.

nfischer:
LGTM. Moving this to the dev branch since it's a breaking change.

Hopefully we don't have to worry about unexpected scenarios where you can't assign to properties of exception is caught. This approach should be perfectly fine for any real exception.

nfischer:
@freitagbr it looks like we may have lost this commit on dev branch with a force push (I wish github supported auto merging). I'll just stick with manual merges instead of rebases to update dev branch.

Would you be able to redo this commit?

freitagbr:
Yeah, let me do redo this.

PR #232, opened 2015-09-10
Added -L follow sym link option to cp

charlesverge:
Added option to allow cp to follow sym links and copy source files rather than sym link.

PR #549, opened 2016-11-05
fix: maxdepth doesn't limit total number of copies

### LINKED ISSUE #547: bug: cp() doesn't always copy everything

v6.7.0 (probably any version though)

Tested this on 0.7, but it's probably the case on master still, since we haven't touched the code in a while.

### Operating system:

Ubuntu (but the bug was originally observed on OS X in shelljs/shx#94).

### Description of the bug:

This was originally filed as shelljs/shx#94. It's not specific to shx though, so I'm moving it upstream. I don't think this has to do with 72 being a hard limit on `cp()`, but I think there must be some bug that's causing us to exit early while processing long lists, or perhaps just for processing lots of data.

Although shelljs is missing some files, it consistently misses the same files each time I re-run the command (so I don't think it's a race condition).

Using [this zip](https://drive.google.com/file/d/0B3Hzs-LoXgdkUDl3NmVJVmxOSDQ/view) as the `node_modules` folder in the example:

```js
shell.cp('-r', 'node_modules', 'otherModules');
```

```bash
$ diff -r node_modules/ otherModules/ # observe that otherModules/ is missing some directories
```

Interestingly, if I run this command, I don't see any significant issues:

```js
shell.cp('-r', 'otherModules', 'otherModules2'); // copying the copy seems to get all the files
```

nfischer:
Prevent maxdepth from limiting the total number of copies, only allow it to
limit the maximum depth

Fixes #547

### DIFF
--- a/src/cp.js
+++ b/src/cp.js
@@ -152,6 +152,7 @@ function cpdirSyncRecursive(sourceDir, destDir, opts) {
       }
     }
   } // for files
+  opts.depth--;
 } // cpdirSyncRecursive
 
 function cpcheckcycle(sourceDir, srcFile) {
--- a/src/cp.js
+++ b/src/cp.js
@@ -84,18 +84,12 @@ function copyFileSync(srcFile, destFile, options) {
 //
 // Licensed under the MIT License
 // http://www.opensource.org/licenses/mit-license.php
-function cpdirSyncRecursive(sourceDir, destDir, opts) {
+function cpdirSyncRecursive(sourceDir, destDir, currentDepth, opts) {
   if (!opts) opts = {};
 
-  /* Ensure there is not a run away recursive copy. */
-  if (typeof opts.depth === 'undefined') {
-    opts.depth = 0;
-  }
-  if (opts.depth >= common.config.maxdepth) {
-    // Max depth has been reached, end copy.
-    return;
-  }
-  opts.depth++;
+  // Ensure there is not a run away recursive copy
+  if (currentDepth >= common.config.maxdepth) return;
+  currentDepth++;
 
   // Create the directory where all our junk is moving to; read the mode of the
   // source directory and mirror it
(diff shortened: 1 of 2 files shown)


nfischer:
Apparently the maxdepth feature added in #232 wasn't implemented correctly. It was compared against the total number of calls to the function, not to the depth of the function call stack. We incremented the value, but we forgot to decrement it. Because objects in JS are pass-by-reference, all invocations of the function reference the same depth counter.

nfischer:
@freitagbr will you have time to review in the next couple days? If not, I can merge myself.

freitagbr:
Yes, I will review this soon.

nfischer:
@freitagbr changes are finished

src/cp.js
```diff
@@ -84,18 +84,12 @@ function copyFileSync(srcFile, destFile, options) {
...
-    opts.depth = 0;
-  }
-  if (opts.depth >= common.config.maxdepth) {
-    // Max depth has been reached, end copy.
-    return;
-  }
-  opts.depth++;
+  // Ensure there is not a run away recursive copy
+  if (currentDepth >= common.config.maxdepth) return;
+  currentDepth++;
```
freitagbr: Well, ok. I was thinking that `depth` would be the last function parameter, so that you wouldn't have to explicitly call `cpdirSyncRecursive` with it. To initialize the depth:
```
if (!depth) depth = 0;
```
just like we initialize `opts`. Then you can `depth++` or call `cpdirSyncRecursive(..., depth + 1)`, whichever you think is clearer.

nfischer: As for the order, I usually leave options as the final parameter. `cpdirSyncRecursive` is totally internal, so requiring the `depth` parameter isn't so awful.

If we allow `depth` and `opts` to both be optional, then we have to account for the case when `depth` is missing but `opts` isn't, and vice-versa. Sounds very bug-prone to me, for very little external benefit.

freitagbr: Ok, I can see that. As long as it's internal, it shouldn't be a problem.

nfischer: Ok, just give it an LGTM when ready to merge

freitagbr:
LGTM

nfischer:
LGTM (forgot that you don't have lgtm permission yet)

Issue #525, opened 2016-10-09
Add shx as a dependency for testing


nfischer:
Tests in #524 are breaking because there's no good way to test the `cwd` option. There are also other parts where we try to use `grep`, but can't use it if it doesn't exist on Windows. If we import `shx` as a dev dependency, then we can get away with using `shx pwd` and `shx grep` for these tests.

nfischer:
Also, just realized that shx only supports v4+. I think the best thing is to just expand it with polyfills where necessary to support v0.11+. This might be a bit of work, but it should be achievable.

nfischer:
shx now has the proper support. We need to fix some of the remaining shx issues and then make a new release. Once ShellJS depends on that release, we can close this.

nfischer:
shx 0.2.0 has been released, so this is no longer blocked.

nfischer:
Fixed by #550

PR #550, opened 2016-11-06
Get pipe tests running on Windows.

binki:
I used the FIND command which ships with Windows as something to
pass data through.

The test checks that a POSIX-like find(1) exists and skips the tests
if this condition is true because Windows-provided FIND is incompatible
with find(1).

### DIFF
--- a/test/pipe.js
+++ b/test/pipe.js
@@ -50,18 +50,25 @@ assert.equal(shell.error(), null);
 assert.equal(result.toString(), shell.cat('resources/uniq/pipeSorted').toString());
 
 // Synchronous exec
+var findCommand;
 if (process.platform === 'win32') {
+  // On Windows, we use the FIND command. FIND is semantically similar
+  // to POSIX’s grep(1) except it only matches string literals. Also,
+  // FIND requires the first argument, the search string, to be
+  // wrapped in double quotes when accessed via GetCommandLine().
+  // https://technet.microsoft.com/en-us/library/bb490906.aspx
+
+  // Get the full path to FIND. If we just exec('find'), Windows will
+  // try to run ./find.js with Windows Script Host.
+  findCommand = '"' + shell.which('FIND') + '"';
+
   // Ensure the user does not have the wrong version of FIND. I.e.,
   // require the following to return error.
-  result = shell.exec('CD..&FIND . -name no');
+  result = shell.exec(findCommand + ' . -name no');
   if (!shell.error()) {
     console.error('Warning: Cannot verify piped exec: Found POSIX-like find(1) in PATH. This test assumes a Windows environment with its FIND command (fix your PATH, exit cygwin/mingw32/MSYS2).');
   } else {
-    // “CD..” is to avoid Windows running test/find.js with Windows
-    // Script Host. I tried using regex to remove ‘.’ from PATH but
-    // apparently exec sources the OS’s environment when executing which
-    // ends up re-adding ‘.’ to PATH.
-    result = shell.cat('resources/grep/file').exec('CD..&FIND "alph"');
+    result = shell.cat('resources/grep/file').exec(findCommand + ' "alph"');
     assert.ok(!shell.error());
     assert.equal(result, 'alphaaaaaaabeta\r\nalphbeta\r\n');
   }
@@ -79,14 +86,12 @@ if (process.platform === 'win32') {
 if (process.platform === 'win32') {
   // Ensure the user does not have the wrong version of FIND. I.e.,
   // require the following to return error.
-  result = shell.exec('CD..&FIND . -name no');
+  result = shell.exec(findCommand + ' . -name no');
   if (!shell.error()) {
     console.error('Warning: Cannot verify piped exec: Found POSIX-like find(1) in PATH. This test assumes a Windows environment with its FIND command (fix your PATH, exit cygwin/mingw32/MSYS2).');
     shell.exit(123);
   } else {
-    // “CD..” is to avoid Windows running test/find.js with Windows
-    // Script Host.
-    shell.cat('resources/grep/file').exec('CD..&FIND "alph"', function (code, stdout) {
+    shell.cat('resources/grep/file').exec(findCommand + ' "alph"', function (code, stdout) {
       assert.equal(code, 0);
       assert.equal(stdout, 'alphaaaaaaabeta\r\nalphbeta\r\n');
       shell.exit(123);
--- a/test/pipe.js
+++ b/test/pipe.js
@@ -49,63 +49,16 @@ result = shell.sort('resources/uniq/pipe').uniq('-c').sort('-n');
 assert.equal(shell.error(), null);
 assert.equal(result.toString(), shell.cat('resources/uniq/pipeSorted').toString());
 
-// Synchronous exec
-var findCommand;
-if (process.platform === 'win32') {
-  // On Windows, we use the FIND command. FIND is semantically similar
-  // to POSIX’s grep(1) except it only matches string literals. Also,
-  // FIND requires the first argument, the search string, to be
-  // wrapped in double quotes when accessed via GetCommandLine().
-  // https://technet.microsoft.com/en-us/library/bb490906.aspx
-
-  // Get the full path to FIND. If we just exec('find'), Windows will
-  // try to run ./find.js with Windows Script Host.
-  findCommand = '"' + shell.which('FIND') + '"';
-
-  // Ensure the user does not have the wrong version of FIND. I.e.,
-  // require the following to return error.
-  result = shell.exec(findCommand + ' . -name no');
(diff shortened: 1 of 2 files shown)


test/pipe.js
```diff
@@ -50,23 +50,54 @@ assert.equal(shell.error(), null);
...
 
 // Synchronous exec
-// TODO: add windows tests
-if (process.platform !== 'win32') {
+if (process.platform === 'win32') {
+  // Windows-specific
+
+  // Ensure the user does not have the wrong version of FIND. I.e.,
+  // require the following to return error.
+  result = shell.exec('CD..&FIND . -name no');
```
nfischer: Could we change this to `shell.which('find')`? That might be more readable.

test/pipe.js
```diff
@@ -50,23 +50,54 @@ assert.equal(shell.error(), null);
...
+  if (!shell.error()) {
+    console.error('Warning: Cannot verify piped exec: Found POSIX-like find(1) in PATH. This test assumes a Windows environment with its FIND command (fix your PATH, exit cygwin/mingw32/MSYS2).');
+  } else {
```
nfischer: Is there a clean way to do this without a nested if-else? It makes things a little confusing to read. One preferable alternative would be an else-if ladder.

If we wait for #512 to be merged, then you might be able to simplify this to:

```js
if (!shell.error()) {
  console.error('Cannot verify...');
  return; // exit early
}
// normal case....
```

This does not have to be blocked by #512 though if we can do a similarly readable alternative.

test/pipe.js
```diff
@@ -76,8 +107,6 @@ if (process.platform !== 'win32') {
     });
   } else {
     console.error('Warning: Cannot verify piped exec');
+    shell.exit(123);
```
nfischer: Why was this added? It looks like it should've been here already. Did something change here, or did we just forget this before?

test/pipe.js
```diff
@@ -50,23 +50,54 @@ assert.equal(shell.error(), null);
...
   }
-} else {
-  console.error('Warning: Cannot verify piped exec');
 }
 
 // Async exec
-// TODO: add windows tests
-if (process.platform !== 'win32') {
+if (process.platform === 'win32') {
+  // Windows-specified
```
nfischer: nit: you can delete this comment, it's implied by the conditional. Same thing for the section above.

test/pipe.js
binki: I’m going to guess that no one verified that the tests work on unix when grep is not available because, with grep being part of the POSIX spec, it’s hard to find such a unix system. This whole conditional on `shell.which('grep').stdout` should probably just be removed because you can assume grep exists when `process.platform !== 'win32'`.

test/pipe.js
binki: Ah, I was just imitating the `// unix-specific` comment. I’ll remove both and push a new commit to this PR.

test/pipe.js
binki: Well, could just wrap it all in an IEFE. I was just imitating how the unix code was conditional on `grep(1)`, though I inverted the conditional ;-). Also, not sure how one could use an else-if ladder without repeating expressions in the subsequent conditionals. Functions with returns or the current nesting are the natural ways to do this without repetition, I think. Please let me know how I should fix this!

binki:
@nfischer I find ss64 very useful. See [`FIND`](http://ss64.com/nt/find.html). It’s basically like `grep(1)` (and even the process description in Task Manager mentions grep on modern Windows). An interesting trait of `FIND` is that it relies on arguments being quoted a specific way, probably for historical reasons. An article which I didn’t fully read but touches on some of those things is [“Everyone quotes command line arguments the wrong way”](https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/). Basically, on Windows, programs [parse the commandline themselves instead of being passed a vector of arguments](http://stackoverflow.com/a/4094897/429091). Here’s an example:

```
C:\Users\ohnob>ECHO hi | FIND "hi"
hi

C:\Users\ohnob>ECHO hi | FIND "hey"

C:\Users\ohnob>ECHO hi | FIND hey
FIND: Parameter format not correct
```

FIND requires its first argument to be passed between double quotes and refuses to run otherwise. It also appears to only support exact substring matching—it doesn’t appear to have a pattern/glob/regex matching option. But for simply testing that piping in shelljs works, it’s sufficient and provided on all Windows machines.

However, an alternative approach would be to do something like `.exec(common.nodeBinPath + ' -e "process.stdin.pipe(process.stdout)"')` except with some logic that would actually mutate the stream. By just shelling out to node instead of a unix-specific built-in, we could get rid of the whole `if (process.platform)`. Want me to rewrite this PR using this idea instead? ^^

test/pipe.js
binki: I’m not sure how we would do that. The point is to detect people (like me) who like to put unix utilities such as [MSYS2](https://msys2.github.io/) in `PATH` and attempting to `npm test` instead of cleaning up their environment first. Basically, the test failure that you would get when you get `find(1)` instead of `FIND` would be quite confusing, so I think this check is worth it. `FIND` will error because the first argument is not wrapped in `""` whereas unix-like `find(1)` will exit successfully. Just finding the patch to `FIND` would return a path. I guess you could do something like trying to determine if the returned path is in `%WINDIR%\System32` or not, but that seems a lot more fragile than actually testing the behavior.

nfischer: Sorry, I didn't explain myself clearly. I mean, use `var find = shell.which('find')` to get a path to find. Now replace `CD..&FIND` with the `find` variable.

You'll still need to check to make sure it's the right `find`, but it should still guarantee that it's not referring to the local `find.js`. `CD..&FIND` is very confusing to non-Windows people, whereas `shell.which()` more intuitively returns a path to `find`.

Or if there's so much trouble with the `find` command on Windows, what if there's a more straightforward alternative?

test/pipe.js
nfischer: Hmmm... if you think this is the best format, let's leave it for now. When we switch to ava, I think we'll have more flexibility to refactor this for readability.

test/pipe.js
nfischer: Let's leave it, just to be safe. But I agree, `grep` is almost certainly available by default, and I would be extremely surprised if anyone intentionally deleted it. I've actually been meaning to refactor this to check for the existence of `grep` (windows or otherwise). Not necessary for this PR though.

binki:
Cool. I think I’ve addressed all of your outstanding review comments which was just removing the `// Windows-specific` comment. Please let me know if I missed anything, otherwise I assume you’re waiting for others to look at this and give feedback (if so, sorry for this bugspam xD).

nfischer:
@binki I'll give it another look a bit later and make sure everything is good, and then merge. Would you be able to add a brief explanation of the semantics of windows `find` in a comment in the code? Something like:

```
Windows `find` is semantically similar to Unix `grep`. It requires that its first argument be
surrounded by double quotes, and it only matches literal strings (not regex).
```

Add in whatever you think is useful in making the test-case more readable. Thanks!

binki:
Oh, I just noticed #525 now. That sounds like a much cleaner way of achieving the goal of this PR in the end.

test/pipe.js
```diff
@@ -50,18 +50,25 @@ assert.equal(shell.error(), null);
...
 if (process.platform === 'win32') {
+  // On Windows, we use the FIND command. FIND is semantically similar
+  // to POSIX’s grep(1) except it only matches string literals. Also,
+  // FIND requires the first argument, the search string, to be
+  // wrapped in double quotes when accessed via GetCommandLine().
+  // https://technet.microsoft.com/en-us/library/bb490906.aspx
+
+  // Get the full path to FIND. If we just exec('find'), Windows will
+  // try to run ./find.js with Windows Script Host.
+  findCommand = '"' + shell.which('FIND') + '"';
```
nfischer: Can you move this section out of the conditional? It'll make it easier when we merge this into #512.

nfischer:
@binki I just released the new `shx`. Would you be interested in refactoring this to use `shx grep` instead of `find`? Then we can get rid of the Windows-specific and unix-specific conditionals. That would be even more readable :smile:

binki:
Maybe I should rebase to get that ugly `FIND` stuff out  of the history?

nfischer:
> Maybe I should rebase to get that ugly FIND stuff out of the history?

I wouldn't worry about it. We squash commits upon merge, so that should take care of it.

LGTM once Travis passes

nfischer:
Fixes #525

binki:
Someone needs to retrigger the travis build because it “errored” rather than “failed” due to [badly timed Mac OS X timeouts](https://www.traviscistatus.com/incidents/06xg686x7wgw) ;-).

PR #555, opened 2016-11-13
Fix: copy recursively from read only location. (fixes #98)

### LINKED ISSUE #98: cp does not recursively copy from readonly location
shell.cp('-R', <src readonly> , dest) fails with the following error:

cp: copyFileSync: could not write to dest file (code=EACCES): <dest>

Fransz:
I think this fixes #98 
Tested on OS X, but not on windows/linux.

A test was added.

### DIFF
--- a/ava-test/cp.js
+++ b/ava-test/cp.js
@@ -647,3 +647,28 @@ test('Test with recursive option and symlinks.', t => {
     t.falsy(shell.test('-L', 'sym.lnk'));
   });
 });
+
+
+// cp -R should be able to copy a readonly src (issue #98).
+// On Windows, chmod acts VERY differently so skip these tests for now
+test('cp -R should be able to copy a readonly src. issue #98; (Non window platforms only)', t => {
+  if (common.platform !== 'win') {
+    shell.cp('-r', 'resources/cp', t.context.tmp);
+    shell.chmod('555', `${t.context.tmp}/cp/`);
+    shell.chmod('555', `${t.context.tmp}/cp/dir_a`);
+    shell.chmod('555', `${t.context.tmp}/cp/dir_b`);
+    shell.chmod('555', `${t.context.tmp}/cp/a`);
+
+    const result = shell.cp('-r', `${t.context.tmp}/cp`, `${t.context.tmp}/cp_cp`);
+    t.falsy(shell.error());
+    t.falsy(result.stderr);
+    t.is(result.code, 0);
+
+    t.is(shell.ls('-R', `${t.context.tmp}/cp`) + '', shell.ls('-R', `${t.context.tmp}/cp_cp`) + '');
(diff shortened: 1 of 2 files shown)


nfischer:
@Fransz thanks so much for the PR! Can you investigate the windows failure? This may be due to chmod not working correctly on Windows. If so, feel free to skip the tests on Windows.

 Can you rewrite the tests to not modify `resources`? Copy the files out first, then modify them, and then do a copy to test that read-only works.

Can you please add a test for an individual read-only file as well?

Fransz:
Tests are adjusted, and skipped on windows as asked by @nfischer 

I'm not sure to change the file permissions on the copied directories/files. 
The -p flag should be used for that really.

nfischer:
@Fransz can you also verify that the permissions are still correct after copying? This should do the trick:

```js
assert.equal(fs.statSync('tmp/cp_cp').mode & parseInt('777', 8), parseInt('555', 8));
```

Fransz:
I've added assertions for the permissions being correct after copying as @nfischer requested.
This doesn't pass travis however;

The test fails on a single job (not the complete travis build);
Repeating the build fails again, but now on a different job (i.e. different node version);

I've tried several times. The only thing in common seems to be that all failures happend on OS X jobs.

I'm new to travis, so i really do not know how to fix this.

nfischer:
@Fransz I'll look into Travis. Sounds like flakiness, either in the test or in Travis. I'll ping this if there's any action we need from you. Thanks for pointing it out!

nfischer:
To me, this doesn't look like a bug that is due to Travis. I can't reproduce it locally though, so I restarted the whole build. If it passes with no failures, we can merge now and fix it the next time the failure arises. Otherwise we'll investigate further.

I may mark the test with a TODO to examine the flakes later.

nfischer:
Just restarted the test again. Since it seems like the test keeps flaking, @Fransz would you mind redoing the tests after #565 is merged? Using a real framework may help with test reliability.

Fransz:
@nfischer, i rewrote the tests in ava. 
Unfortunately this doesn't seem to help. Tests keeps failing in an unpredicatable way.

nfischer:
@Fransz thanks for doing the refactor. It looks like other tests in `cp` are flaking now. Those flakes are probably due to our refactor to ava.

I'm going to see if I can resolve the general flakiness, and then I'll try to merge PR in.

nfischer:
Bumping this back to v0.8, since we're focusing efforts there.

nfischer:
Bumping to v0.9 milestone

nfischer:
Unfortunately, this has sat unmerged for a long time. Looking at the code, this doesn't seem to be too hard to cherry pick to a new branch. I'll try this out.

nfischer:
Close in favor of #870. Thanks for this fix!

PR #557, opened 2016-11-16
Add newline to output of echo

### LINKED ISSUE #476: Echo doesn't return value ending in a trailing newline
### Node version:

v6.2.0

0.7.0
### Operating system:

OSX 10.11.5
### Description of the bug:

Using `toEnd` doesn't prepend a newline when chained to `echo`.

``` javascript
shell.echo(`A`).toEnd(`test.txt`);
shell.echo(`B`).toEnd(`test.txt`);
```

test.txt becomes: `AB`

``` bash
echo A >> test.txt
echo B >> test.txt
```

test.txt becomes:

```
A
B
```

freitagbr:
Fixes #476

### DIFF
--- a/ava-test/echo.js
+++ b/ava-test/echo.js
@@ -5,6 +5,14 @@ import utils from './utils/utils';
 
 shell.config.silent = true;
 
+test.beforeEach(t => {
+  t.context.tmp = utils.getTempDir();
+});
+
+test.afterEach.always(t => {
+  shell.rm('-rf', t.context.tmp);
+});
+
 //
 // Valids
 //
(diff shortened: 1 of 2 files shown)


nfischer:
@freitagbr Can you also add a test?

freitagbr:
@nfischer Sure. I am thinking however that the behavior of this should be changed. Currently, `console.log` is used, and a newline is automatically placed at the end of the output. Instead, I think `process.stdout.write` should be used, as it only writes to stdout what is literally passed in.

nfischer:
@freitagbr unix `echo` adds a newline by default. Probably best to keep using `console.log()` for this scenario.

You raise a good point though, so I filed #557

nfischer:
Looks good. Let's wait until we ava-ify the tests for `echo` though, to avoid callback hell.

freitagbr:
Ok. Is it worth adding support for `-n` in the meantime? Or should I wait until this is merged?

nfischer:
You can do that in a new PR, it shouldn't necessarily be blocked by this. Either way, let's wait to get the `echo` tests ava-ified.

freitagbr:
@nfischer I rewrote the test for this branch.

nfischer:
LGTM, make sure we land this on dev, not master.

PR #558, opened 2016-11-16
Update sed documentation regarding capture groups

### LINKED ISSUE #507: Sed should allow a replacement string to contain `\1` for match groups
Sed should accept replacement strings that contain `\1` and `\2` and the like, for the sake of doing replacements. Currently, this is probably supported by `$1`, `$2`, etc. (but that's JS-syntax, not Unix-syntax).

``` js
sed(/my(.*)regex$/, '\\1 works', 'file.txt');
// probably the closest we can get to Unix sed's syntax without a breaking change:
// sed 's/my.*regex$/\1 works/' file.txt
```

This ties in closely with `shx sed`, since it doesn't currently support substitutions like this.

freitagbr:
Fixes #507

### DIFF
--- a/src/sed.js
+++ b/src/sed.js
@@ -28,15 +28,11 @@ common.register('sed', _sed, {
 //@
 //@ Note:
 //@
-//@ Unix `sed` specifies capture groups using `\n` syntax. In ShellJS (and JavaScript, in general),
-//@ capture groups are specified using the `$n` syntax. The following are equivalent:
+//@ Like unix `sed`, ShellJS `sed` supports capture groups. Capture groups are specified
+//@ using the `$n` syntax:
 //@
 //@ ```javascript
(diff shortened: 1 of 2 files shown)


src/sed.js
```diff
@@ -25,6 +25,19 @@ common.register('sed', _sed, {
 //@
 //@ Reads an input string from `files` and performs a JavaScript `replace()` on the input
 //@ using the given search regex and replacement string or function. Returns the new string after replacement.
+//@
+//@ Note:
+//@
+//@ Unix `sed` specifies capture groups using `\n` syntax. In ShellJS (and JavaScript, in general),
```
nfischer: Don't use `\n`, since that sounds like the newline character. It might suffice to just provide the javascript example and skip the shell code, for brevity's sake.

nfischer:
@freitagbr LGTM once you regenerate docs

nfischer:
LGTM

Issue #560, opened 2016-11-17
Switch to the ava test framework


nfischer:
This is a bug to track the progress of switching to the ava test framework.

Here is the proposed approach:

 - [x] make a directory called `ava-test/`
 - [x] copy the resources to `ava-test/resources/` (leave a copy where it is now, too)
 - [x] one by one, remove a file from `test/` and put its ava equivalent in `ava-test/`
 - [x] modify `npm test`: instead of `node scripts/run-tests`, turn it into `node scripts/run-tests && ava --serial ava-test/*.js`
 - [x] remove the old `test/` folder, and run `git mv ava-test/ test/`
 - [x] remove the `run-tests` script and change `npm test` to run only `ava --serial test/*.js`

nfischer:
I also set up a github project for this (see the projects tab). We can use that if we find it helpful.

nfischer:
Reassigning to @freitagbr so that he can finish up the last two things bits (should be fine to do in one PR). Parallelizing the tests isn't necessary, since it actually runs slower on my machine and serial tests seem pretty reliable.

nfischer:
Resolved in #630

PR #561, opened 2016-11-17
test: set up ava and move cd.js

nfischer:
Set up ava as the new test framework. Migrate the tests for the cd command to
use the ava framework.

This is the first commit in the plan I suggested in [this comment on #512](https://github.com/shelljs/shelljs/pull/512#issuecomment-261162455).

Fixes bullet points 1, 2, and 4 of #560


ava-test/cd.js
```diff
@@ -0,0 +1,103 @@
+import test from 'ava';
+import shell from '..';
+import path from 'path';
+import common from '../src/common';
+import fs from 'fs';
+import utils from './utils/utils';
```
freitagbr: Can we group these imports together? I like to group node stdlib imports, then dependency imports, then local imports. Something like:
```javascript
import fs from 'fs';
import path from 'path';
import test from 'ava';
import shell from '..';
import common from '../src/common';
import utils from './utils/utils';
```

package.json
```diff
@@ -43,10 +43,12 @@
     "rechoir": "^0.6.2"
   },
   "devDependencies": {
+    "ava": "^0.16.0",
+    "coffee-script": "^1.10.0",
     "eslint": "^2.0.0",
     "eslint-config-airbnb-base": "^3.0.0",
     "eslint-plugin-import": "^1.11.1",
-    "coffee-script": "^1.10.0",
+    "jscodeshift": "^0.3.28",
```
freitagbr: Is this dependency necessary? I thought it was used to convert our tests to ava initially, not sure if it's necessary anymore.

nfischer: Ah, you're probably right. Thanks for pointing this out. I'll remove it and verify.

ava-test/cd.js
nfischer: SGTM. I'll probably also put newline breaks between sections of the imports

nfischer: New format:

- core modules (alphabetized)
- npm modules (alphabetized)
- local files (alphabetized, but with `import shell from '..';` first because it needs to be loaded before the others)

nfischer:
@freitagbr PTAL

freitagbr:
LGTM

PR #564, opened 2016-11-19
test: refactor cat tests to ava

nfischer:


nfischer:
@freitagbr I'm pretty confident that all tests have been properly converted. Feel free to review this for style.

ava-test/cat.js
```diff
@@ -0,0 +1,61 @@
+import fs from 'fs';
+
+import test from 'ava';
+
+import shell from '..';
+
+test.beforeEach(() => {
+  shell.config.silent = true;
+});
```
freitagbr: Is this necessary? In the old code, this is only configured once, not reconfigured before every test. I think this can be simplified to just this:
```
test.before(() => {
    shell.config.silent = true;
});
```
Or even more simply to this:
```
shell.config.silent = true;
```

nfischer: Doing it each time should be harmless. But since it slows things down, it should be fine to move it out to just once before.

nfischer:
@freitagbr PTAL

freitagbr:
LGTM

nfischer:
LGTM

nfischer:
Merging, since it looks like lgtm.co is broken...

PR #572, opened 2016-11-23
test: refactor dirs tests to AVA

nfischer:
This is an automated PR, so double check that I got all the test cases. I'll follow up once I've verified that this translates everything correctly.


ava-test/dirs.js
```diff
@@ -0,0 +1,39 @@
+import path from 'path';
+
+import test from 'ava';
+
+import shell from '..';
+
+test.beforeEach(() => {
+  shell.config.silent = true;
```
freitagbr: This line doesn't need to be in `beforeEach`.

nfischer: Let's leave it here. It's harmless to set it to silent repeatedly, and it's safer in case someone accidentally unsets it in one of the tests.

As long as there's some set up to be done in the `beforeEach`, I think it's better to just leave lines like this in there as well (unless they need to only be performed once).

freitagbr: Ok, sure.

freitagbr:
LGTM

nfischer:
LGTM

PR #580, opened 2016-11-24
test: refactor 'which' tests to AVA

nfischer:
@freitagbr ready for review. This should be good to go.


ava-test/which.js
```diff
@@ -0,0 +1,47 @@
...
+  t.falsy(result);
+});
+
+//
+// Valids
+//
+
+// TODO(nate): make sure this does not have a false negative if 'git' is missing
+test('basic usage', t => {
+  const git = shell.which('git');
```
freitagbr: Why not use `node`? Surely if this test is running on a system, that system has `node` installed...

nfischer: Node can be installed as `nodejs` on Linux. It's always installed as `node.exe` on Windows, though. We can count on git to be always be installed on a dev's box. That's the rationale. We could probably change the next one to `git` as well if we wanted to be consistent

freitagbr: That's a good point, let's keep it `git` then.

freitagbr:
LGTM

PR #582, opened 2016-11-24
test: refactor pwd tests to AVA

nfischer:
@freitagbr this should be good


ava-test/pwd.js
```diff
@@ -0,0 +1,41 @@
...
+  shell.rm('-rf', t.context.tmp);
+});
+
+
+//
+// Valids
+//
+
+test('initial directory', t => {
+  const _pwd = shell.pwd();
```
freitagbr: `pwd` instead of `_pwd`?

PR #586, opened 2016-11-24
test: refactor rm tests to AVA

nfischer:
@freitagbr ready for review

I added a new test to `rm` for a case we weren't checking. Turns out our `rm` doesn't pass the test case (`rm -rf on a symbolic link to a dir deletes its contents`), so I'm skipping it for now and marked it with a TODO. I'll file an issue for it.


freitagbr:
LGTM, let's fix that issue soon though.

PR #588, opened 2016-11-24
Fix: errors on working without permission 

HyperUa:


nfischer:
Thanks for the PR! Would you be able to add tests for these cases? Tests are either in `test/` or `ava-test/`

nfischer:
@HyperUa can you fix the lint errors?

HyperUa:
@nfischer About testing: I can't make cross platform decision. 
I tried to work with files with other owner and got errors, that's why I added the code. In my opinion it is better to say that user doesn't have permission than throwing error.

nfischer:
@HyperUa if you can't make a cross-platform test, can you at least make a unix-specific or windows-specific test? We haven't received any bugs about this, so I have no idea what this code is fixing. We have a general policy against merging untested code unless there's a good reason why it can't be tested.

nfischer:
@HyperUa any follow up on my previous comments? It would be great if you could either add some tests or explain how I can reproduce the issues that this PR solves. Thanks.

nfischer:
Closing this, due to lack of a response. If someone can show me the cases this code handles, I'll reopen and take a look.

PR #590, opened 2016-11-28
Add "-n" option to echo

### LINKED ISSUE #559: feature: echo -n
Add support for the `-n` option for the `echo` command. This removes the default trailing newline.

Ex.

```sh
$ echo foo bar # notice the extra space after "foo bar"
foo bar

$ echo -n foo bar # notice no extra space after "foo bar"
foo bar
$
```

This can be implemented using `process.stdout.write()` fairly easily, just remember to concatenate all arguments into a single string to pass to `process.stdout.write()`

freitagbr:
Fixes #559

### DIFF
--- a/ava-test/echo.js
+++ b/ava-test/echo.js
@@ -79,3 +79,44 @@ test.cb('piping to a file', t => {
     t.end();
   });
 });
+
+test.cb('-n option', t => {
+  const script = "require('../global.js'); echo('-n', 'message');";
+  utils.runScript(script, (err, stdout) => {
+    t.falsy(err);
+    t.is(stdout, 'message');
+    t.end();
+  });
+});
+
+test.cb('-ne option', t => {
+  const script = "require('../global.js'); echo('-ne', 'message');";
+  utils.runScript(script, (err, stdout) => {
+    t.falsy(err);
+    t.is(stdout, 'message');
+    t.end();
+  });
+});
+
+test.cb('-en option', t => {
+  const script = "require('../global.js'); echo('-en', 'message');";
+  utils.runScript(script, (err, stdout) => {
+    t.falsy(err);
+    t.is(stdout, 'message');
+    t.end();
+  });
+});
+
+test.cb('piping to a file with -n', t => {
+  // see issue #476
+  shell.mkdir(t.context.tmp);
+  const tmp = `${t.context.tmp}/echo.txt`;
+  const script = `require('../global.js'); echo('-n', 'A').toEnd('${tmp}'); echo('-n', 'B').toEnd('${tmp}');`;
+  utils.runScript(script, (err, stdout) => {
+    const result = shell.cat(tmp);
+    t.falsy(err);
+    t.is(stdout, 'AB');
+    t.is(result.toString(), 'AB');
+    t.end();
+  });
+});
--- a/ava-test/echo.js
+++ b/ava-test/echo.js
@@ -107,6 +107,15 @@ test.cb('-en option', t => {
   });
 });
 
+test.cb('-en option with escaped characters', t => {
+  const script = "require('../global.js'); echo('-en', '\\tmessage\\n');";
+  utils.runScript(script, (err, stdout) => {
+    t.falsy(err);
+    t.is(stdout, '\tmessage\n');
+    t.end();
+  });
+});
+
 test.cb('piping to a file with -n', t => {
   // see issue #476
   shell.mkdir(t.context.tmp);
--- a/ava-test/echo.js
+++ b/ava-test/echo.js
@@ -129,3 +129,9 @@ test.cb('piping to a file with -n', t => {
     t.end();
   });
 });
+
+test('stderr with unrecognized options is null', t => {
+  const ret = shell.echo('-asdf');
+  t.falsy(ret.stderr);
+  t.is(ret.stdout, '-asdf\n');
+});
--- a/ava-test/echo.js
+++ b/ava-test/echo.js
@@ -131,7 +131,7 @@ test.cb('piping to a file with -n', t => {
 });
 
 test('stderr with unrecognized options is null', t => {
-  const ret = shell.echo('-asdf');
-  t.falsy(ret.stderr);
-  t.is(ret.stdout, '-asdf\n');
+  const result = shell.echo('-asdf');
+  t.falsy(result.stderr);
+  t.is(result.stdout, '-asdf\n');
 });
--- a/ava-test/echo.js
+++ b/ava-test/echo.js
@@ -130,7 +130,8 @@ test.cb('piping to a file with -n', t => {
   });
 });
 
-test('stderr with unrecognized options is null', t => {
+test('stderr with unrecognized options is empty', t => {
+  // TODO: console output here needs to be muted
   const result = shell.echo('-asdf');
   t.falsy(result.stderr);
   t.is(result.stdout, '-asdf\n');
--- a/src/echo.js
+++ b/src/echo.js
@@ -1,3 +1,5 @@
+var format = require('util').format;
+
 var common = require('./common');
 
 common.register('echo', _echo, {
@@ -31,13 +33,13 @@ function _echo(opts, messages) {
   if (option >= 0) {
     // ignore options
     messages.shift();
-    output = messages.join(' ');
+    output = format.apply(null, messages);
     if (option === 0) {
       // add newline if -n is not passed
       output += '\n';
     }
   } else {
-    output = messages.join(' ') + '\n';
+    output = format.apply(null, messages) + '\n';
   }
 
   process.stdout.write(output);
--- a/src/echo.js
+++ b/src/echo.js
@@ -30,17 +30,15 @@ function _echo(opts) {
 
   // If the first argument starts with '-', parse it as options string.
   // If parseOptions throws, it wasn't an options string.
-  if (opts[0] === '-') {
-    try {
-      options = common.parseOptions(opts, {
-        'e': 'escapes',
-        'n': 'no_newline'
-      }, {
-        silent: true
-      });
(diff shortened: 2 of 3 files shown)


ava-test/echo.js
```diff
@@ -79,3 +79,44 @@ test.cb('piping to a file', t => {
...
+  const script = "require('../global.js'); echo('-n', 'message');";
+  utils.runScript(script, (err, stdout) => {
+    t.falsy(err);
+    t.is(stdout, 'message');
+    t.end();
+  });
+});
+
+test.cb('-ne option', t => {
+  const script = "require('../global.js'); echo('-ne', 'message');";
```
nfischer: Can we also test `echo -en 'foo\n'` to make sure that `-e` still works?

src/echo.js
```diff
@@ -23,12 +27,23 @@ function _echo(opts, messages) {
...
+    output = format.apply(null, messages);
+    if (option === 0) {
+      // add newline if -n is not passed
+      output += '\n';
+    }
+  } else {
+    output = format.apply(null, messages) + '\n';
   }
 
-  console.log.apply(console, messages);
```
nfischer: Let's only use `process.stdout.write` for when `-n` is passed. Calling `process.exit` immediately after `process.stdout.write` has some issues.

src/echo.js
```diff
@@ -23,12 +27,23 @@ function _echo(opts, messages) {
   // allow strings starting with '-', see issue #20
   messages = [].slice.call(arguments, opts ? 0 : 1);
 
-  if (messages[0] === '-e') {
-    // ignore -e
+  var option = ['-e', '-n', '-ne', '-en'].indexOf(messages[0]);
+  var output;
+
+  if (option >= 0) {
```
nfischer: It might be easier to just check:

 - is the first argument a real option?
 - do the options contain `-n`?

And just switch the behavior on that condition. That way we can keep using `console.log` by default, which has some nicer behavior (see my comment below).

src/echo.js
freitagbr: I don't think it matters, because `console.log` is just a wrapper around `process.stdout.write`. See [this](https://github.com/nodejs/node/blob/master/lib/console.js#L43).

ava-test/echo.js
freitagbr: Sure, I can add a test for that.

src/echo.js
freitagbr: I think we can use `common.parseOptions` to parse the first argument. If we catch the `option not recognized` error, then just print that string.

src/echo.js
freitagbr: For reference, it has been that way since at least [v0.7.4](https://github.com/nodejs/node/blob/v0.7.4-release/lib/console.js#L25).

src/echo.js
nfischer: Yeah, but it won't throw the error as an exception, it'll return early from the `echo` function. You can hack around it by setting `config.fatal` to have it throw an exception. A clearer way is to add an option to `common.error` to throw an exception instead of exiting (regardless of `config.fatal`).

The approach you have here works if we only have two options, but it gets cumbersome if we ever support `-E` (which we may want to support in the future).

src/echo.js
nfischer: Hmmm ok, let's use this approach then.

src/echo.js
freitagbr: Ok, I've added a function to `echo.js` that parses options just for `echo`. This means adding `-E` later on will be easier.

src/echo.js
```diff
@@ -9,26 +11,79 @@ common.register('echo', _echo, {
...
   }
 
-  console.log.apply(console, messages);
-  return messages.join(' ') + '\n';
+  process.stdout.write(output);
+
+  return output;
 }
+
+function parseEchoOptions(opts) {
```
nfischer: Actually, let's use `common.parseOptions()`. It actually does throw an exception upon error regardless of `config.fatal`. It behaves exactly how we want, and it'll be much cleaner to reuse the code.

freitagbr: `parseOptions` will still log to stdout when it doesn't recognize an option:
```
 > try {
... common.parseOptions('-b', {a: 'test'})
... } catch (_) {
... console.log('whoops')
... }
shell.js: option not recognized: b
whoops
```
This is not good if someone tries to echo a string that happens to start with `'-'`.

nfischer: Ah, I see the issue. I think it may be better to either:

 - hack around this by temporarily setting `config.silent = true` (document what we're doing in comments)
 - add a way to silence `parseOptions` (exception only, no stdio)

I think the approach we take depends on if we want to expose this feature in `parseOptions` in the plugin API.

@freitagbr What do you think?

freitagbr: I would prefer an extra `silence` argument or option for `parseOptions`. Plugins could probably make use of the option, too.

freitagbr:
@nfischer I refactored `echo` to use `common.parseOptions`, can you take a look when you get a chance?

nfischer:
@freitagbr please rebase off `dev`

freitagbr:
Done.

src/echo.js
```diff
@@ -9,26 +11,55 @@ common.register('echo', _echo, {
...
+        silent: true
+      });
+    } catch (_) {}
+  }
 
-  if (messages[0] === '-e') {
-    // ignore -e
+  if (options) {
+    // first argument was options string,
+    // so do not print it
```
nfischer: Join these lines?

src/echo.js
```diff
@@ -9,26 +11,55 @@ common.register('echo', _echo, {
...
   // allow strings starting with '-', see issue #20
-  messages = [].slice.call(arguments, opts ? 0 : 1);
+  var messages = [].slice.call(arguments, opts ? 0 : 1);
+  var options;
+  var output;
+
+  // If the first argument starts with '-',
+  // parse it as options string.
+  // If parseOptions throws,
+  // it wasn't an options string.
```
nfischer: Join these lines?

src/echo.js
```diff
@@ -9,26 +11,55 @@ common.register('echo', _echo, {
...
+    // first argument was options string,
+    // so do not print it
     messages.shift();
+    output = format.apply(null, messages);
+    if (!options.no_newline) {
+      // add newline if -n is not passed
+      output += '\n';
+    }
+  } else {
+    output = format.apply(null, messages) + '\n';
```
nfischer: We're appending `\n` in this branch, but we're also appending it above. Can we flatten-out this logic? Something like:

```js
try {
  options = common.parseOptions(messages[0], { ... }, { silent: true });
  messages.shift();
} catch (e) {
  options = {};
}
output = format.apply(null, messages);

if (!options.no_newline) {
  output += '\n';
}

process.stdout.write(output);
```

freitagbr: I think we still need to check if the first argument starts with `'-'`, because we don't know if it needs to be printed or not otherwise.

nfischer: Ok, sure. `parseOptions`  has useless behavior if you pass it a string that doesn't start with `-`.

freitagbr:
I'll wait for the pull request I submitted ([#615](https://github.com/shelljs/shelljs/pull/615)) to be merged before moving on here.

freitagbr:
There's an issue here. When passing a string starting with `-`, but containing unrecognized options, `stderr` on the return object is set:
```
> echo('-hi')
-hi
{ [String: '-hi\n']
  stdout: '-hi\n',
  stderr: 'echo: option not recognized: h',
  code: 1,
  cat: [Function: bound ],
  head: [Function: bound ],
  tail: [Function: bound ],
  to: [Function: bound ],
  toEnd: [Function: bound ],
  sed: [Function: bound ],
  sort: [Function: bound ],
  uniq: [Function: bound ],
  grep: [Function: bound ],
  exec: [Function: bound ] }
```
This is obviously not how `echo` behaves in the shell.

nfischer:
@freitagbr can you manually clear out `common.state.error`? That should do the trick.

I doubt there's a great way to test it, but maybe you could add a test like:

```js
const ret = shell.echo('-n', ''); // echo nothing
t.falsy(ret.stderr);
t.is(ret.stdout, '');
t.is(ret.code, 0);
```

This case wouldn't add stderr anyway, but it'll save us from huge future regressions while also not polluting the output.

freitagbr:
@nfischer I added the fix, though there is already a test case that checks that `stderr === ''`

nfischer:
@freitagbr it needs to check for `ret.stderr`. The `.stderr` isn't ever being outputted (that's what the `runScript` checks), but it was being returned as a property (so we need to check function return value).

freitagbr:
Ok, test added. PTAL

ava-test/echo.js
```diff
@@ -129,3 +129,9 @@ test.cb('piping to a file with -n', t => {
     t.end();
   });
 });
+
+test('stderr with unrecognized options is null', t => {
+  const ret = shell.echo('-asdf');
```
nfischer: @freitagbr will this output to the console? If so, let's at least add a TODO to mute console output

ava-test/echo.js
```diff
@@ -129,3 +129,9 @@ test.cb('piping to a file with -n', t => {
     t.end();
   });
 });
+
+test('stderr with unrecognized options is null', t => {
```
nfischer: s/null/empty/ or s/null/falsy/

We may want to switch stderr from `null` to the empty string in a future release.

ava-test/echo.js
freitagbr: Yes it does. I suppose we can fix this along with mocking `process.stdout` like you suggested for #622.

freitagbr:
@nfischer PTAL

PR #593, opened 2016-12-01
Use Object.assign if possible

### LINKED ISSUE #592: refactor: objectAssign should refer to Object.assign if it exists, or the internal polyfill otherwise
Inside `common.js`, we export either `objectAssign()` or `Object.assign()` as `common.extend()`. However, we still use our internal implementation inside `common.js`, and we should prefer `Object.assign()` when it's available.

freitagbr:
Fixes #592

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -245,7 +245,7 @@ function randomFileName() {
 exports.randomFileName = randomFileName;
 
 // objectAssign(target_obj, source_obj1 [, source_obj2 ...])
-// Polyfill for Object.assign
+// "Ponyfill" for Object.assign
 //    objectAssign({A:1}, {b:2}, {c:3}) returns {A:1, b:2, c:3}
 var objectAssign = typeof Object.assign === 'function' ?
   Object.assign :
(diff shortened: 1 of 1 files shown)


src/common.js
```diff
@@ -245,19 +245,21 @@ function randomFileName() {
 exports.randomFileName = randomFileName;
 
 // objectAssign(target_obj, source_obj1 [, source_obj2 ...])
-// Ponyfill for Object.assign
+// Polyfill for Object.assign
```
nfischer: nit: isn't this technically still a [ponyfill](https://ponyfoo.com/articles/polyfills-or-ponyfills) since it doesn't modify the global `Object`? Doesn't matter either way to me, just curious on the terminology. Either way, I agree with the implementation.

nfischer:
LGTM

src/common.js
freitagbr: Wow, I had no idea that was a thing. Let me change it back but put quotes around it.

PR #598, opened 2016-12-08
Fix: rm behavior regarding symlinks

### LINKED ISSUE #519: Feature request: Add "shelljs.unlink"
On my Mac (macOS Sierra) `shelljs.rm('/symlink/dir')` of a symlink causes a "permission denied" error, but `shelljs.exec('unlink /symlink/dir')` works fine. Just thinking you could add `unlink()` for symmetry to `ln()`.

freitagbr:
Fixes #519

### DIFF
--- a/src/rm.js
+++ b/src/rm.js
@@ -123,7 +123,7 @@ function _rm(options, files) {
     }
 
     // If here, path exists
-    if (stats.isFile() || stats.isSymbolicLink()) {
+    if (stats.isFile()) {
       // Do not check for file writing permissions
       if (options.force) {
         common.unlinkSync(file);
@@ -148,6 +148,11 @@ function _rm(options, files) {
     // Recursively remove existing directory
     if (stats.isDirectory() && options.recursive) {
       rmdirSyncRecursive(file, options.force);
+    }
+
+    // Path is a symbolic link
+    if (stats.isSymbolicLink()) {
+      common.unlinkSync(file);
     }
   }); // forEach(file)
   return '';
--- a/src/rm.js
+++ b/src/rm.js
@@ -124,34 +124,20 @@ function _rm(options, files) {
 
     // If here, path exists
     if (stats.isFile()) {
-      // Do not check for file writing permissions
-      if (options.force) {
-        common.unlinkSync(file);
-        return;
-      }
-
-      if (isWriteable(file)) {
+      if (options.force || isWriteable(file)) {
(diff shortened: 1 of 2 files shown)


src/rm.js
```diff
@@ -149,6 +149,11 @@ function _rm(options, files) {
     if (stats.isDirectory() && options.recursive) {
       rmdirSyncRecursive(file, options.force);
     }
+
+    // Path is a symbolic link
+    if (stats.isSymbolicLink()) {
```
nfischer: Maybe we should turn this into an else-if ladder?

Also, what happens if this is a different type of file, such as a fifo (like `mkfifo` creates)?

nfischer: Also, we have lots of code duplication here that should probably get cleaned up. We call `common.unlinkSync()` in two different branches for `isFile`, and we should probably move the `isDirectory` branches into one:

```js
if (isFile) {
  if (options.force || isWriteable()) {
    // remove
  } else {
    // error
  }
} else if (isDirectory()) {
  if (options.recursive) {
     // remove
  } else {
     // error
  }
} else if (isLink()) {
  // remove
}
```

freitagbr: Ok, that makes sense.

Regarding other file types, here are the types that `fs.lstatSync` recognizes:
```
stats.isFile()
stats.isDirectory()
stats.isBlockDevice()
stats.isCharacterDevice()
stats.isSymbolicLink()
stats.isFIFO()
stats.isSocket()
```
Out of these, I think it only really makes sense to unlink files, directories, symlinks, and maybe fifos. Maybe we should stick to those 4, and ignore anything else?

nfischer: We can add support for the others later. In this PR, we need to make sure that we have an `else` case that performs whatever would have been done before (assuming it was reasonable behavior) to avoid a breaking change.

nfischer:
@freitagbr what was the original behavior on the other file types, such as fifos? It looks like this approach will ignore those files, which probably isn't what we want. It may be sufficient to have one `else` to catch all of them (if that's consistent with previous behavior).

src/rm.js
freitagbr: I think the original behavior has been maintained: noop for anything that isn't a directory, file or symlink.

nfischer:
Let's add support for removing other file types in another PR. @freitagbr can you file the issue?

freitagbr:
Sure, I will do that.

PR #609, opened 2016-12-08
test: refactor ls to use AVA

nfischer:
Will update when I verify all tests are copied


nfischer:
Ready for review

ava-test/ls.js
```diff
@@ -0,0 +1,481 @@
...
+  let result = shell.ls('-l', 'resources/ls');
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  let idx = -1;
+  for (let k = 0; k < result.length; k++) {
+    if (result[k].name === 'file1') {
+      idx = k;
+      break;
+    }
+  }
```
freitagbr: How about this:
```js
const idx = result.map(r => r.name).indexOf('file1');
```

ava-test/ls.js
```diff
@@ -0,0 +1,481 @@
...
+  let result = shell.ls('-lR', 'resources/ls/');
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  let idx = -1;
+  for (let k = 0; k < result.length; k++) {
+    if (result[k].name === 'a_dir/b_dir') {
+      idx = k;
+      break;
+    }
+  }
```
freitagbr: Same here:
```js
const idx = result.map(r => r.name).indexOf('a_dir/b_dir');
```

ava-test/ls.js
nfischer: Done

ava-test/ls.js
nfischer: Done

nfischer:
PTAL

PR #613, opened 2016-12-08
chore: switch to files attribute from npmignore

nfischer:
Prefer the `files` attribute over `.npmignore` to minimize the bytes users download when installing ShellJS. This cuts the size down from ~1.2M (measured in `v0.7.5`) to ~300K (measured via a local install).


freitagbr:
LGTM

PR #616, opened 2016-12-09
refactor: list all commands in commands.json

nfischer:
This also changes the order of commands in the docs, due to how I changed
the doc-generating script.

Let me know if you think we should enforce some sort of order on the list of commands. I just mostly kept the order from before.

This should be safe to put in master. I've tested to make sure that the `files` attribute is properly updated.


freitagbr:
I think it would be helpful to alphabetize them , especially because the readme is generated from that list.

nfischer:
Done

Issue #618, opened 2016-12-11
rm -rf doesn't work if the directory contains an asar archive in Electron


tgds:
### Node version (or tell us if you're using electron or some other framework):
electron v1.4.8
node v7.1.0

### ShellJS version (the most recent version/Github branch you see the bug on):
shelljs v.0.7.5

### Operating system:
macOS Sierra 10.12.1
Windows 8.1

### Description of the bug:
Getting `rm: could not remove file (code ENOTDIR)` errors when running recursive rm in Electron to remove a directory and the directory happens to contain an asar archive. 

I believe that this is caused by the fact that Electron patches fs [to treat asar files as virtual directories](https://github.com/electron/electron/blob/master/docs/tutorial/application-packaging.md#node-api). Thus, ShellJS sees the .asar file as a directory and attempts to remove all individual files as given by fs.readdir('example.asar'). 

Monkey patching rm.js to require original-fs seems to solve it. 
I expected similar problems with cp -r, but doing a quick test that seems to work without issues.

Now, I am very new in the industry, so I'd need some guidance on crafting a good PR and discuss what makes the most sense to do here. Would checking if original-fs is available and using it in rm.js be the best solution, or should this be extended globally? Do you think this could potentially have some unwanted side effects or can think of a better solution? 

Thank you!

### Example ShellJS command to reproduce the error:

```javascript
// Somewhere in main or renderer Electron js
const shelljs = require('shelljs')
shelljs.rm('-rf', '/PathToADirectoryContainingAsarArchive')
```

nfischer:
@tgds Thanks for reporting this!

This could be a tricky thing to fix. I'll need to learn more about electron and asar archives. We already have some issues with electron support (see #480), so we have quite a few things to work on.

Could you point me toward docs on original-fs? I'm assuming this is an electron-specific thing to copy over the real `fs` to a different package name.

tgds:
@nfischer Thanks! Yeah, you're assuming correctly. It looks like it should be straightforward in the end. I should pay more attention when reading the [docs](https://github.com/electron/electron/blob/master/docs/tutorial/application-packaging.md) next time. Looks like [this switch](https://github.com/electron/electron/blob/master/docs/tutorial/application-packaging.md#treating-an-asar-archive-as-a-normal-file) should do it: 
```javascript
process.noAsar = true
``` 
Will test this and send a PR later on today, I hope. 

[...]

PR #628, opened 2016-12-19
Turn off Electron asar archive support.

tgds:
ShellJS fails to remove asar archives when run in Electron due to
Electron fs patches that treat asar archives as virtual directories.
See issue #618 .
To overcome this, we set process.noAsar to true.

### DIFF
--- a/src/rm.js
+++ b/src/rm.js
@@ -105,6 +105,11 @@ function isWriteable(file) {
 //@
 //@ Removes files.
 function _rm(options, files) {
+  // Switch Electron asar support off
+  // Prevents unexpected behaviour when attempting to remove an asar archive
+  var noAsar = process.noAsar;
+  process.noAsar = true;
+
   if (!files) common.error('no paths given');
 
   // Convert to array
@@ -141,6 +146,8 @@ function _rm(options, files) {
       common.unlinkSync(file);
     }
   }); // forEach(file)
+
+  process.noAsar = noAsar;
   return '';
 } // rm
 module.exports = _rm;


nfischer:
@tgds Thanks for the patch. I agree that electron support is something we need to work on, and I'd like to move toward that in future releases.

From the looks of it, this seems like something end users might prefer to do on their own. I have reservations about modifying globals, especially since ShellJS is often used with other packages. This has led in the past to some hard-to-find bugs.

Although we're resetting noAsar's value here, it's still not a total perfect reset, since things subtly change (fails `'key' in obj` check):

```js
> // in regular node
> 'noAsar' in process
false
> var noAsar = process.noAsar
undefined
> process.noAsar = noAsar
undefined
> 'noAsar' in process // this value has changed, which end users might not expect
true
```

Even if we perfectly restore the state, there's always a chance that `rm()` will change in the future and might unexpectedly skip the cleanup steps, which would be hard to diagnose.

In the interest of kicking off support for electron, what if we instead document this in a new ShellJS wiki page, specifically for electron issues? That way ShellJS users know what to do and won't be surprised if this change breaks some other behavior.

tgds:
@nfischer Yeah, certainly doesn't feel like the cleanest solution. Very much appreciate your feedback,  I still need to develop the sense of knowing what one can safely afford to do. 
How about requiring original-fs instead of fs when it's available? I know this would again be a little messy since we would be using slightly different fs library at one place.

nfischer:
Is there a reason why requiring users to use `process.noAsar = true;` before calling `shell.rm()` is not feasible?

Using `original-fs` might be the right approach. This approach is certainly safer, but we would probably need to add electron as an environment we test against in CI, which probably won't happen for a while.

Are there any scenarios where `rm` is useful if `noAsar` isn't modified? Also, does this asar stuff make `cat`, `cp`, etc. more useful or less useful? I wouldn't want to do anything until we're sure we aren't breaking a valid use case.

nfischer:
@tgds I started this [wiki page](https://github.com/shelljs/shelljs/wiki/Electron-compatibility) for documenting electron support. I think I can hack something out pretty quickly to expose a workaround for `exec()`. If you find any other issues with electron support, let me know and I'll see if we can find reasonable workarounds.

Depending on how difficult electron support is, this may be slated for our v0.8 or v0.9 release, but we should probably stick with reasonable workarounds at least until then.

nfischer:
Closing, for the reasons noted in #618. We can reopen if documentation isn't sufficient.

PR #635, opened 2016-12-21
fix: allow non-normalized paths as input to mkdir

### LINKED ISSUE #634: mkdir fails with non-normalized path

```javascript
shelljs = require('shelljs');
shelljs.mkdir('-p', 'asdf/./');
```

Expected result is that no error should be thrown and certainly the process should not terminate.  `'-p'` flag should not fail for any existing path or part of path.

nfischer:
Adds tests to make sure that non-normalized paths (i.e. path/to/./dir) are
valid for a few commands, including mkdir() which previously failed when given
the -p flag.

Fixes #634

### DIFF
--- a/src/mkdir.js
+++ b/src/mkdir.js
@@ -75,7 +75,7 @@ function _mkdir(options, dirs) {
 
     try {
       if (options.fullpath) {
-        mkdirSyncRecursive(dir);
+        mkdirSyncRecursive(path.resolve(dir));
       } else {
         fs.mkdirSync(dir, parseInt('0777', 8));
       }
--- a/test/cp.js
+++ b/test/cp.js
@@ -259,7 +259,6 @@ test(
 );
 
 test('recursive, everything exists, no force flag', t => {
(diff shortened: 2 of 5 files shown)


src/mkdir.js
```diff
@@ -75,7 +75,7 @@ function _mkdir(options, dirs) {
 
     try {
       if (options.fullpath) {
-        mkdirSyncRecursive(dir);
+        mkdirSyncRecursive(path.resolve(dir));
```
freitagbr: It seems like `fs.mkdirSync` normalizes the path passed to it anyway. Is that why `path.resolve` is not needed below?

nfischer: Yes, that's correct.

It's needed here because, for each recursive call, we process another piece of the path. This can lead us to processing something that isn't valid for `mkdir`, like `.` or `..`. It can also lead us to processing the same directory twice (ex. `foo/../foo/`) which also isn't valid.

nfischer:
@freitagbr LGTY?

freitagbr:
Yes, LGTM.

PR #636, opened 2016-12-21
refactor: create common.execPath

### LINKED ISSUE #633: Move execPath into common
We should replace `process.execPath` with something like `common.execPath`. While `process.execPath` does not always point to NodeJS (like if it's running within electron), `common.execPath` ideally should, allowing us to reliably use the within `exec` and other utilities.

At startup, we can set `common.execPath` to:

 - `process.execPath` if not within electron
 - otherwise use `which('node')` if it returns something
 - otherwise use `which('nodejs')` if it returns something
 - otherwise we can leave it null and fail early whenever we try to use it (like in `shell.exec`), displaying a descriptive error message (at least an improvement over the current cryptic error message).

Also, this would allow electron users to manually override `common.execPath` with the hardcoded path of their NodeJS binary if the above approach fails. While not ideal, that should at least be a suitable workaround until we can figure out a solid way to reliably get a path to the current node instance.

Linking this to #480

nfischer:
Switch to using `common.execPath` instead of `process.execPath` directly and warn
electron users if we were unable to find the correct path to NodeJS.

This should be safe to merge into master because exec didn't work before anyway.

Fixes #633

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -26,7 +26,10 @@ var objectAssign = typeof Object.assign === 'function' ?
   };
 exports.extend = objectAssign;
 
-// Module globals
+// Check if we're running under electron
+var isElectron = Boolean(process.versions.electron);
+
+// Module globals (assume no execPath by default)
 var DEFAULT_CONFIG = {
   fatal: false,
   globOptions: {},
@@ -34,16 +37,22 @@ var DEFAULT_CONFIG = {
   noglob: false,
   silent: false,
   verbose: false,
+  execPath: null,
 };
+
 var config = {
   reset: function () {
     objectAssign(this, DEFAULT_CONFIG);
+    if (!isElectron) {
+      this.execPath = process.execPath;
+    }
   },
   resetForTesting: function () {
-    objectAssign(this, DEFAULT_CONFIG);
+    this.reset();
     this.silent = true;
   },
 };
+
 config.reset();
 exports.config = config;
 
--- a/src/exec.js
+++ b/src/exec.js
@@ -19,6 +19,10 @@ common.register('exec', _exec, {
 // Node is single-threaded; callbacks and other internal state changes are done in the
 // event loop).
 function execSync(cmd, opts, pipe) {
+  if (!common.config.execPath) {
+    common.error('Unable to find a path to the node binary. Please manually set config.execPath');
+  }
+
   var tempDir = _tempDir();
   var stdoutFile = path.resolve(tempDir + '/' + common.randomFileName());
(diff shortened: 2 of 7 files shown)


nfischer:
@freitagbr PTAL

src/exec.js
```diff
@@ -19,6 +19,10 @@ common.register('exec', _exec, {
 // Node is single-threaded; callbacks and other internal state changes are done in the
 // event loop).
 function execSync(cmd, opts, pipe) {
+  if (!common.execPath) {
+    common.error('Unable to find a path to the node binary. Please manually set common.execPath');
+  }
```
freitagbr: What about `execAsync`? Maybe this should go in `_exec`.

nfischer: @freitagbr `execAsync` doesn't actually depend on it. It uses `child_process.exec()` directly, instead of our `execSync` which uses something like `child_process.execSync(process.execPath + ' script.js')`. You can do a simple search for `process.execPath` and you'll see it only shows up in `execSync`.

src/common.js
```diff
@@ -409,3 +409,9 @@ function _register(name, implementation, wrapOptions) {
   }
 }
 exports.register = _register;
+
+// Expose the path to the NodeJS binary to utilities
+var isElectron = Boolean(process.versions.electron);
+if (!isElectron) {
+  exports.execPath = process.execPath;
```
freitagbr: Maybe `common.execPath` should be configurable through `shell.config`?

src/exec.js
freitagbr: Ok, that makes sense.

src/common.js
nfischer: Probably the better spot for it. We'll have to be careful with the name we come up for in #534 (which is a different feature, that currently sounds very similar).

nfischer:
Blocked on #641. I'll update this after that's merged.

freitagbr:
#641 is merged, so this can go through now.

src/common.js
freitagbr: Should we wait for #534 to implement that then?

nfischer:
Updates:

 - `common.execPath` ➡️ `common.config.execPath`
 - Error message for invalid execPath is changed: `Please manually set common.execPath` ➡️ `Please manually set config.execPath`

📜 I didn't document the change in the README because it affects very few people. I'll document it in the wiki page for electron support after merge

freitagbr:
CI failures are unrelated, so LGTM.

src/common.js
nfischer: We'll just wait on #534. I'm not convinced if that's the best approach, since it's something that can easily be implemented by users of our module.

PR #638, opened 2016-12-27
chore: set up test coverage

### LINKED ISSUE #347: Improve test coverage
This is a general issue that our unit tests should be more comprehensive. I've noticed that sometimes unit tests don't actually test correct behavior (this was the issue with some tests in `touch.js` that I fixed. I'm sure there are also many options/edge cases for functions that are not tested very comprehensively.

As we work toward future releases, we should also work on improving our test framework, to prevent regressions. Also, we should probably add more sanity checks in the `run-tests.js` script, since because this uses ShellJS to test ShellJS, a major break (for example: if `ls()` returns the empty array every time) can cause the test runner to give a false positive.

I'm putting the "help wanted" label on here, so if anyone is looking for something to contribute on, more test cases would be welcome.

If you'd like to contribute, please try to isolate your changes to one file per PR (it makes it much easier to review). Also, it helps if you tag this issue number in your PR. We're also interested in refactoring test cases (so as long as you know JS, you'd be able to help) :+1:

nfischer:
Use `nyc` for test coverage and ignore all nyc-related files in git & eslint.
This increases testing time (~35sec -> ~60sec on my 4-core machine).

Fixes #347

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,6 @@
-ava-test/resources/
-ava-test/tmp/
+.nyc_output/
+coverage/
+make.js
+node_modules/
 test/resources/
 test/tmp/
-node_modules/
-make.js
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 npm-debug.log*
 node_modules
 tmp
+coverage/
+.nyc_output/
 
 # Linux
 *~
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   ],
   "scripts": {
     "posttest": "npm run lint",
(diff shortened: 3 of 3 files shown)


nfischer:
This is a regression in testing time, so I'll leave it to @freitagbr's discretion as to whether to merge. The real benefit will be once we set up coveralls or codecov for this project. Codecov seems to be mostly broken for shx right now, and I haven't had time to look into it yet, so I have no issues with using coveralls if that's preferred.

freitagbr:
I think the code coverage report is useful, and the increase in testing time is not that great. Let's go for it.

PR #641, opened 2017-01-05
refactor: add config.reset() and .resetForTesting()

nfischer:
This adds `.reset()` and `.resetForTesting()` to `shell.config` and use `.resetForTesting()` as a standard set-up for unit tests.

This also adds `npm run test-no-coverage`, since testing with coverage every time gets pretty slow.

`.reset()` is implicitly tested by checking default values for `shell.config`. `.resetForTesting()` is relied upon by all unit tests.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -9,15 +9,42 @@ var shell = require('..');
 
 var shellMethods = Object.create(shell);
 
+// objectAssign(target_obj, source_obj1 [, source_obj2 ...])
+// "Ponyfill" for Object.assign
+//    objectAssign({A:1}, {b:2}, {c:3}) returns {A:1, b:2, c:3}
+var objectAssign = typeof Object.assign === 'function' ?
+  Object.assign :
+  function objectAssign(target) {
+    var sources = [].slice.call(arguments, 1);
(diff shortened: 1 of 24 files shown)


nfischer:
@freitagbr this should be passing all tests now.

Let me know if you think we should document `shell.config.reset()` as part of the public API.

src/common.js
```diff
@@ -10,14 +10,24 @@ var shell = require('..');
...
+  noglob: false,
+  silent: false,
+  verbose: false,
+};
+var config = {
+  reset: function () {
+    Object.assign(this, DEFAULT_CONFIG);
+  },
+  resetForTesting: function () {
+    Object.assign(this, DEFAULT_CONFIG);
```
freitagbr: These should be `objectAssign`.

nfischer: Whoops, forgot to push the fix. Yeah, this should work now

freitagbr:
Let's also add documentation for `shell.config.reset`. Something simple like:
```
## shell.config.reset()
Reset the configuration to the following defaults:
```
{
    foo: 'bar',
    ...
}
```
Other than that, LGTM.

PR #648, opened 2017-01-17
Chore drop 0.12

### LINKED ISSUE #647: Drop support for v0.12
According to the [NodeJS LTS schedule](https://github.com/nodejs/LTS), v0.12 support was dropped by Node in December 2016. This means it's probably reasonable to drop the support in our v0.8 release.

This has some benefits for us:

 - we can cut v0.11 and v0.12 from our Travis builds for faster CI (maybe even iojs too, which sometimes has false negatives)
 - we can significantly trim down on the code inside `src/exec.js` (there's a whole [v0.10-specific else-branch](https://github.com/shelljs/shelljs/blob/4c48631d745ef4cb3e9d198b37e5772fb57ce905/src/exec.js#L128))
 - we can [clean up tests](https://github.com/shelljs/shelljs/blob/4c48631d745ef4cb3e9d198b37e5772fb57ce905/test/exec.js#L116)
 - we can trim down `src/tempdir.js` a [bit as well](https://github.com/shelljs/shelljs/blob/4c48631d745ef4cb3e9d198b37e5772fb57ce905/src/tempdir.js#L43)
 - we can remove our `Object.assign()` polyfill (if we remove iojs as well)

@freitagbr what are your thoughts on dropping v0.12 support? What about dropping iojs support? We'll get the biggest CI benefits if we go with iojs, but I don't know if many people still rely on it (my guess is iojs users have moved on to v4+).

nfischer:
Fixes #647

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,6 @@
 language: c++
 sudo: false
 env:
-  - NODE_VERSION="0.11"
-  - NODE_VERSION="0.12"
-  - NODE_VERSION="iojs-v1"
-  - NODE_VERSION="iojs-v2"
-  - NODE_VERSION="iojs-v3"
   - NODE_VERSION="4"
   - NODE_VERSION="5"
   - NODE_VERSION="6"
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Node.js API. You can use it to eliminate your shell script's dependency on Unix 
 its familiar and powerful commands. You can also install it globally so you can run it from outside
 Node projects - say goodbye to those gnarly Bash scripts!
 
-ShellJS is proudly tested on every node release since `v0.11`!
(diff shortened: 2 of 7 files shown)


nfischer:
@freitagbr I updated the README to mention v4+. I don't think it's necessary to mention this corresponds with v0.8, since that'll be the version once we merge dev into master.

nfischer:
And yes, dev branch contains changes for v0.8. Once we get close to a release, we can merge into master and do the finishing touches on the master branch. Postponing the merge is helpful, otherwise v0.8 docs will show up on the README before it's released.

freitagbr:
CI passed, so LGTM.

freitagbr:
LGTM, needs a rebase off of `dev` though.

nfischer:
Updated with Github UI. Should be fine, since it all squashes anyway.

PR #655, opened 2017-02-04
feature: add -a option for which command

termosa:
With a `-a` option it will return an array of all matched binaries. Stdout is formatted as needed (joined with `\n` all binaries and `\n` is added to the end).
I've tested it on Mac OSX and Windows.

### DIFF
--- a/test/which.js
+++ b/test/which.js
@@ -46,9 +46,15 @@ test('Windows can search with or without a .exe extension', t => {
   }
 });
 
-test('Search all of matched binaries', t => {
+test('Search all of matched binaries shoud return an array', t => {
   const result = shell.which('-a', 'node');
   t.falsy(shell.error());
   t.truthy(result);
   t.truthy(result.length);
+});
+
+test('Search all for not exist command should return an empty array', t => {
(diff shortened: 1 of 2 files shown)


test/which.js
```diff
@@ -45,3 +45,10 @@ test('Windows can search with or without a .exe extension', t => {
     t.is(node.toString(), nodeExe.toString());
   }
 });
+
+test('Search all of matched binaries', t => {
+  const result = shell.which('-a', 'node');
+  t.falsy(shell.error());
+  t.truthy(result);
+  t.truthy(result.length);
```
nfischer: Can you check that `result[0] === shell.which('node')`?

termosa: @nfischer this is a good idea. I'll get it done.

termosa:
@nfischer hi!

`shell.which(command)` returns `String(path)`. Is it an expected result?
Currently, I've added a test that compares two strings but uses `.toString` for both before the compare, but it doesn't look well for me.

So let me know what do you think about `String` wrapper so I will either remove it or add it to the results of `shell.which('-a', command)`.

nfischer:
LGTM. I'll merge after CI passes

termosa:
Thank you, I'll be waiting for release.

nfischer:
I think we can make a v0.7.x release for this

PR #659, opened 2017-02-10
refactor: reduce repeated code in which()

### LINKED ISSUE #656: Refactor which() (too many repeated code blocks)
Self-assigning. I'll take this up when I have time. This is a pure refactor to reduce repeated code blocks.

I noticed we have 3 blocks of code where we push the match (`where.push(attempt)...where = attempt`): 2 inside of the `win` section, 1 in the unix section. We can easily combine these.

1. Make the path-ext array, but initialize it to `[]` for unix and to `[''].concat(splitPath())` for Windows
2. Always iterate over the path-ext array. The only member is the empty string on unix, so nothing should change.

nfischer:
Use path.delimiter instead of explicit conditional. Try to reuse
repeated code blocks. Remove logical negation at the end
of the function to improve readability.

Blocked on #655 

Fixes #656

### DIFF
--- a/src/which.js
+++ b/src/which.js
@@ -15,13 +15,7 @@ var XP_DEFAULT_PATHEXT = '.com;.exe;.bat;.cmd;.vbs;.vbe;.js;.jse;.wsf;.wsh';
 
 // Cross-platform method for splitting environment PATH variables
 function splitPath(p) {
-  if (!p) return [];
-
-  if (common.platform === 'win') {
-    return p.split(';');
-  } else {
-    return p.split(':');
-  }
+  return p ? p.split(path.delimiter) : [];
 }
 
 function checkPath(pathName) {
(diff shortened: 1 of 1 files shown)


src/which.js
```diff
@@ -13,12 +13,7 @@ var XP_DEFAULT_PATHEXT = '.com;.exe;.bat;.cmd;.vbs;.vbe;.js;.jse;.wsf;.wsh';
 // Cross-platform method for splitting environment PATH variables
 function splitPath(p) {
   if (!p) return [];
-
-  if (common.platform === 'win') {
-    return p.split(';');
-  } else {
-    return p.split(':');
-  }
+  return p.split(path.delimiter);
```
freitagbr: Or a good ol' ternary statement:
```js
return p ? p.split(path.delimiter) : [];
```

src/which.js
```diff
@@ -45,7 +40,17 @@ function _which(options, cmd) {
...
+    var pathExtArray = [''];
+    if (common.platform === 'win') {
+      // In case the PATHEXT variable is somehow not set (e.g.
+      // child_process.spawn with an empty environment), use the XP default.
+      var pathExtEnv = process.env.PATHEXT || XP_DEFAULT_PATHEXT;
+      pathExtArray = splitPath(pathExtEnv.toUpperCase());
+    }
+
     // Search for command in PATH
     pathArray.forEach(function (dir) {
```
freitagbr: Here, we're looping through the path array, and setting `where` under certain conditions, then returning from the `forEach` callback. On the next call, we check if `which` was set, and if it was, we don't continue. FInally, at the end of `checkPath`, we return `where` if it was set. To me, it seems like a regular `for` loop would be a better choice here, because we could return `where` from the function immediately.

src/which.js
```diff
@@ -54,46 +59,35 @@ function _which(options, cmd) {
...
+  // No relative path was specified, so we return what we found from the search
+  if (where) return where;
 
-  where = where || path.resolve(cmd);
+  // Otherwise, return the absolute path of the path specified
+  if (checkPath(cmd)) return path.resolve(cmd);
 
-  return where;
+  // Command not found anywhere?
+  return null;
```
freitagbr: Hmm... Maybe an empty string should be returned here? All of the other paths return strings, and `''` is falsy like `null` is.

src/which.js
nfischer: Yeah, I agree. Index-based for loops kind of suck in JS, but it's probably still better than this exit-early-if-we-found-it approach. I will adjust.

src/which.js
nfischer: `null` is better here because it won't get wrapped by ShellString. The empty string would get wrapped, and `ShellString('')` is actually truthy, unlike `''`.

freitagbr: Ok, that makes sense.

nfischer:
@freitagbr PTAL. Significantly refactored things, hopefully this is a cleaner approach, with less code duplication.

src/which.js
```diff
@@ -45,74 +39,67 @@ function _which(options, cmd) {
...
+    add: function (file) { this.value.push(file); },
+    // Return the first result only unless options.all is specified
+    getValue: function () {
+      return options.all ? this.value : this.value[0];
+    },
+    hasFoundSomething: function () { return Boolean(this.value.length); },
+    // This is the value to return if we haven't found any matches
+    noMatches: options.all ? [] : null,
+    value: [],
+  };
```
freitagbr: This is better than it was before, but this is still a little complicated. On the one hand, this generalizes the two options more clearly, but on the other hand, dealing with one-off objects is a bit messy.

src/which.js
```diff
@@ -45,74 +39,67 @@ function _which(options, cmd) {
...
-        // Assume it's Unix-like
-        if (checkPath(attempt)) {
-          if (all) {
-            where.push(attempt);
-          } else {
-            where = attempt;
-            return;
-          }
+        // Otherwise, see if it exists without the extension
+        var newAttempt = attempt + ext;
```
freitagbr: This is a bit confusing. We're checking if it exists "without the extension", but then we add the extension?

nfischer: Yeah, I'll see if I can make that better. The basic idea is this:

 1. Handle the case where they call `which('node.exe')`
 2. If they instead did `which('node')`, then check for `node.exe`, `node.bat`, `node.cmd`, etc.

src/which.js
nfischer: Sure, I'll take away some of the abstraction

nfischer:
@freitagbr please see if this next set of changes improves clarity

src/which.js
```diff
@@ -45,74 +39,60 @@ function _which(options, cmd) {
...
-            } else {
-              where = attempt;
-              return;
-            }
-          }
-        }
-      } else {
-        // Assume it's Unix-like
+      var match = attempt.match(/\.[^<>:"/\|?*.]+$/);
+      if (match && pathExtArray.search(match[0]) >= 0) { // this is Windows-only
```
freitagbr: I think `indexOf` should be used here, instead of `search`.

nfischer: Ah, this was a mistake. `Array.prototype.search` isn't even defined... `indexOf` is what I was looking for. Thanks

nfischer: Fixed

nfischer:
@freitagbr PTAL

freitagbr:
Ok, this is much better. LGTM.

PR #665, opened 2017-02-23
ls -L (follow symlinks)

### LINKED ISSUE #563: Feature request: ls -L option
Hello! Do you think it would be interesting to add an option for `ls -RAL`? Just like `ls -RA` but also following symlinks. I think `ls -RA` is supported right now but there is no `L` for `ls`.
Thanks!

frandiox:
This closes #563

I added some tests for `-L` option and fixed the other.

@nfischer  If I'm not mistaken, there is a difference between bash's `ls -R` and shelljs's related to dir symlinks. Bash only shows the symlink itself but shelljs shows the direct children as well (1 level). In this case, `ls -RL` will show all the levels, not only the first one.

By the way, I think [this test](https://github.com/shelljs/shelljs/blob/master/test/ls.js#L295) is a mistake since it is just a copy of [another one](https://github.com/shelljs/shelljs/blob/master/test/ls.js#L274), so I removed it.


frankdiox:
@nfischer  Perhaps the difference I pointed out only happens in my local environment? Tests pass:

```
  463 passed
   1 skipped
```

Not sure how to proceed :sweat_smile:

src/ls.js
```diff
@@ -60,7 +62,7 @@ function _ls(options, paths) {
       relName = relName.replace(/\\/g, '/');
     }
     if (options.long) {
-      stat = stat || fs.lstatSync(abs);
+      stat = stat || options.link ? fs.statSync(abs) : fs.lstatSync(abs);
```
freitagbr: I think you need to put the ternary in parentheses:
```js
stat = stat || (options.link ? fs.statSync(abs) : fs.lstatSync(abs));
```
Otherwise the condition for the ternary is `stat || options.link`.

frankdiox:
@freitagbr @nfischer Appveyor is using a Windows environment, now I understand the difference :sweat_smile: 
I'm not sure how the tests should work for Windows since there are no symlinks there.
By the way, about the issue I mentioned in the OP

> If I'm not mistaken, there is a difference between bash's ls -R and shelljs's related to dir symlinks. Bash only shows the symlink itself but shelljs shows the direct children as well (1 level). In this case, ls -RL will show all the levels, not only the first one.

I just reported it with a screenshot on #666
This issue is related to the failing tests as well because since ShellJS follows symlinks 1 level even without `-L`, I had to add +3 entries to some tests while normally only +1 would be enough (the symlink itself, not its children). That would pass in Windows I guess.

nfischer:
@frankdiox this is a relatively small change that affects a lot of test cases. Do you think there's any way we could make this affect fewer test cases?

What if we do `ls -RAL resources/rm/link_to_a_dir/` instead? We won't have to change any resource files, so we won't have to touch any other test cases. Would that work for testing this feature?

nfischer:
That change would fix most of the breaking test cases on Windows.

I would expect we'll still have one breaking test case on Windows, because it handles symlnks differently than unix does. If that's the case, would it make sense to skip this test only for Windows? [Here's an example](https://github.com/shelljs/shelljs/blob/master/test/rm.js#L276) of how we test unix-only code.

frankdiox:
@nfischer Yep, I didn't see there was another symlink in `resources/rm` so I just created a new one under `resources/ls`. I reverted the latest test commit and made a new one using the existing symlink. It should be fine now :+1:

nfischer:
Deferring to @freitagbr for final approval

freitagbr:
LGTM too.

PR #668, opened 2017-02-26
fix: switch commands.json -> commands.js

### LINKED ISSUE #667: Use with webpack broken in 0.7.6
Electron

0.7.6
### Operating system:
OSX
### Description of the bug:

    TypeError: __webpack_require__(...).forEach is not a function
        at eval (webpack:///./~/shelljs/shell.js?:26:25)

```javascript
import { shell } from 'shelljs'
```

Webpack is simply confused by this [commands.json trick](https://github.com/shelljs/shelljs/blob/master/shell.js#L24) introduced in https://github.com/shelljs/shelljs/pull/616.

Works in 0.7.5 before the trick.

nfischer:
WebPack has issues with importing JSON directly and using JavaScript
methods on it. For this reason, using the `.forEach()` method on the
imported JSON array caused a method-not-found error.

Instead, we can make this a real JavaScript array by keeping it in a
JavaScript file and exporting it as a field.

Fixes #667


commands.js
```diff
@@ -0,0 +1,29 @@
+exports.list = [
```
freitagbr: `exports` is initially assigned to `module.exports`. This is why reassigning `exports` doesn't export anything, because `exports` is now whatever you assigned it to, and not `module.exports`.

Anyways, I think it makes more sense to use `module.exports` here, because all we want to export is the array.

scripts/generate-docs.js
```diff
@@ -11,7 +11,7 @@ var docs = grep('^//@', 'shell.js');
 
 // Insert the docs for all the registered commands
 docs = docs.replace(/\/\/@commands\n/g, function () {
-  return require('../commands.json').map(function (commandName) {
+  return require('../commands').list.map(function (commandName) {
```
freitagbr: If `module.exports` is used in `commands.js`, this will become `require('../commands').map(...)`.

shell.js
```diff
@@ -21,7 +21,7 @@ var common = require('./src/common');
 //@commands
 
 // Load all default commands
-require('./commands.json').forEach(function (command) {
+require('./commands').list.forEach(function (command) {
```
freitagbr: Same here.

commands.js
nfischer: Ah, ok. That explanation makes sense. I'll try that instead.

nfischer:
@freitagbr PTAL

freitagbr:
LGTM, CI failures are unrelated.

PR #669, opened 2017-02-26
Fix 666

### LINKED ISSUE #666: Difference between bash ls -R and ShellJS ls -R with symlinks
Node 7.1.0

0.7.6

### Operating system:
Arch Linux 64-bit + bash/zsh

### Description of the bug:
shelljs-lsr

The output of a normal `ls -R` (bash) is different from ShellJS' `ls -R`. I have created a symlink to `a_dir` directory called `a_dir-link`. Bash does not follow that symlink with `ls -R`, it just shows a symlink entry. However, ShellJS follows that link only 1 level. I think the two highlighted lines on the left side should not appear.

```javascript
ls -R
```

frandiox:
This PR continues #665 and fixes #666.

From `node-glob` repo:

>Note that symlinked directories are not crawled as part of a **, though their contents may match against subsequent portions of the pattern. This prevents infinite loops and duplicates and the like.

Therefore, the current pattern (`/**/*`) takes the first level of files inside symlinked directories. This PR changes it to `/**` which does not follow symlinks.

Now both `ls` and `find` only show the symlink directory as a single entry. `ls -RL` follows it.


nfischer:
@frankdiox Let's wait until #665 is merged. Then please rebase this off master and I'll take another look.

frandiox:
@nfischer :+1: 
I made a new PR with this commit: #676
Closing this one now.

nfischer:
@frandiox for future reference, you can rebase locally and do a `git push -f` to the same PR 😄

frandiox:
@nfischer Thanks, just wanted to make it easier to follow in a new PR. The commit is cherry picked from this one anyway :sweat_smile:

Issue #671, opened 2017-02-26
Set up Codecov for the project


nfischer:
Our test coverage is actually pretty decent, so now would be a good time to set up Codecov or coveralls for the project.

Some things we need to take a look at:

 - [x] We should mark hard-to-test code blocks with `/* istanbul ignore next */`
 - [x] `echo` tests appear to have horrible coverage because of how the tests are run. We'll need to resolve #622 to make test coverage more accurate
 - [x] Some lines are untested but should be tested. These cases are pretty trivial, I can take these on.
 - [ ] Some lines of code are actually dead code that are never reached. We might consider removing these at this point.
 - [x] Actually add Codecov to the project 😄 

Blocked on #622

nfischer:
Before marking this as resolved, we should improve test coverage for the commands mentioned in [these issues](https://github.com/shelljs/shelljs/labels/test%20coverage).

PR #672, opened 2017-02-26
test: don't count hard-to-test lines for coverage

nfischer:
No change in logic.

Add `/* istanbul ignore next */` lines for hard-to-test lines so that
they don't count against us during code coverage.

I've also adjusted comments that I found confusing, and changed some
formatting.

Partial fix for #671

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -73,6 +73,7 @@ exports.platform = platform;
 var pipeMethods = [];
 
 function log() {
+  /* istanbul ignore next */
   if (!config.silent) {
     console.error.apply(console, arguments);
   }
@@ -259,6 +260,7 @@ function unlinkSync(file) {
     fs.unlinkSync(file);
   } catch (e) {
     // Try to override file permission
+    /* istanbul ignore next */
     if (e.code === 'EPERM') {
       fs.chmodSync(file, '0666');
       fs.unlinkSync(file);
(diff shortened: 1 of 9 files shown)


src/common.js
```diff
@@ -367,11 +369,13 @@ function wrap(cmd, fn, options) {
           if (e.msg === 'earlyExit') {
             retValue = e.retValue;
           } else {
+            /* istanbul ignore next */
```
freitagbr: I think `/* istanbul ignore else */` above the `if` statement also works.

nfischer: Yeah, either way. I just usually stick with `ignore next`. Do you think I should change it?

freitagbr: With `ignore next`, I think istanbul will mark the `else` route as not taken. `ignore else` should improve test coverage slightly.

nfischer: I'll try it tonight and see if I see a difference in either line or branch coverage

nfischer: @freitagbr done. Same line coverage, but `ignore else` slightly improves branch coverage.

nfischer:
@freitagbr PTAL. I also moved one `ignore next` above the `if` so we get slightly better branch coverage. Branch coverage isn't a priority right now though, so let's just focus on line coverage.

freitagbr:
LGTM

PR #673, opened 2017-02-26
test: add misc. tests to improve coverage

nfischer:
No change in production logic.

This adds missing tests to improve test coverage. This does not change
ShellJS behavior at all--all test cases are testing already-working
functionality.

One test case has been renamed for clarity.

For the "omit directory if missing recursive flag" case, we were
actually already testing that in another case, but we were testing
multiple things in that test case. It's better to test this one error
condition explicitly in its own case.

When adding real tests for `parseOptions()`, we need to explicitly clear
`common.state.error` because we're testing an internal function, not a
wrapped command.

Partial fix for #671

### DIFF
--- a/test/cp.js
+++ b/test/cp.js
@@ -71,7 +71,7 @@ test('source does not exist', t => {
   t.is(result.stderr, 'cp: no such file or directory: asdfasdf');
 });
 
-test('sources does not exist', t => {
+test('multiple sources do not exist', t => {
   const result = shell.cp('asdfasdf1', 'asdfasdf2', t.context.tmp);
   t.truthy(shell.error());
   t.is(result.code, 1);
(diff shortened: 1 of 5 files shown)


test/cp.js
```diff
@@ -112,6 +112,32 @@ test('dest already exists', t => {
   t.is(shell.cat('resources/file2').toString(), oldContents);
 });
 
+test('-nR does not overwrite an existing file at the destination', t => {
+  // Create tmp/new/cp/a
+  const dest = `${t.context.tmp}/new/cp`;
```
freitagbr: I feel like we should be using `path.join` to join paths everywhere, but its probably not an issue for now.

freitagbr: Actually, it looks like one of the `mv` tests failed on windows for this exact reason. Maybe we should create another issue for this?

nfischer: ShellJS should be able to handle paths using `/`, regardless of which platform we're on

nfischer: Yup, looks like we caught a bug. We should probably be converting the output to use a `/` instead of a `\`.

PR #674, opened 2017-02-26
refactor(parseOptions): better handle errors

nfischer:
For the case where `parseOptions()` is passed a string that does not
start with a hyphen, we should reject this string instead of returning
the same value. This has no change on any other test cases, and should
not affect any commands since we are careful about what input we pass to
`parseOptions()`

This also adjust how we were handling error cases in the function. We
were previously using two different calls to `common.error()`, one for
if `errorOptions` is passed, and one without. This hurts readability,
because it reads like "if we have errorOptions, then this is an error,
and if we don't then it's also an error," instead of the more
appropriate, "we reached an error case and should use errorOptions if
it's available, otherwise we should signal an error without using it."

We now reject invalid input at the top of the function. These are regular
`Error`s instead of calls to `common.error()` because they represent misuse
of an internal function.

Added test cases bring this function to 100% test coverage.

Partial fix for #671

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -195,11 +195,12 @@ function parseOptions(opt, map, errorOptions) {
   });
 
   if (!opt) return options; // defaults
+  var hasErrorOpts = typeof errorOptions === 'object';
 
   var optionName;
   if (typeof opt === 'string') {
     if (opt[0] !== '-') {
-      return options;
+      error("Must start the options string with a '-'");
     }
 
     // e.g. chars = ['R', 'f']
@@ -213,10 +214,10 @@ function parseOptions(opt, map, errorOptions) {
         } else {
           options[optionName] = true;
         }
-      } else if (typeof errorOptions === 'object') {
-        error('option not recognized: ' + c, errorOptions);
       } else {
-        error('option not recognized: ' + c);
+        error('option not recognized: ' + c, hasErrorOpts
+          ? errorOptions
+          : undefined);
       }
     });
   } else if (typeof opt === 'object') {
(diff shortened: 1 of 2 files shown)


src/common.js
```diff
@@ -183,11 +183,12 @@ function parseOptions(opt, map, errorOptions) {
   });
 
   if (!opt) return options; // defaults
+  var hasErrorOpts = typeof errorOptions === 'object';
 
   var optionName;
   if (typeof opt === 'string') {
     if (opt[0] !== '-') {
-      return options;
+      error("Must start the options string with a '-'");
```
freitagbr: `error("Options string must start with a '-'")`?

src/common.js
```diff
@@ -183,11 +183,12 @@ function parseOptions(opt, map, errorOptions) {
   });
 
   if (!opt) return options; // defaults
+  var hasErrorOpts = typeof errorOptions === 'object';
```
freitagbr: I would be careful here, because `typeof null === 'object'`. This check exists elsewhere in this function (and in `wrap`).

nfischer:
@freitagbr PTAL

src/common.js
```diff
@@ -172,59 +172,57 @@ exports.getUserHome = getUserHome;
 // Returns {'reference': 'string-value', 'bob': false} when passed two dictionaries of the form:
 //   parseOptions({'-r': 'string-value'}, {'r':'reference', 'b':'bob'});
 function parseOptions(opt, map, errorOptions) {
-  if (!map) error('parseOptions() internal error: no map given');
+  // Validate input
+  if (typeof opt !== 'string' && !(opt instanceof Object)) {
```
freitagbr: `Object.create(null) instanceof Object === false`, but `Object.create(null)` indeed creates an object. This is the correct way to check that something is an object: 
```js
typeof a === 'object' && a !== null
```
Maybe make a function for this.

src/common.js
```diff
@@ -172,59 +172,57 @@ exports.getUserHome = getUserHome;
...
         }
       } else {
-        error('option not recognized: ' + c, hasErrorOpts
-          ? errorOptions
-          : undefined);
+        error('option not recognized: ' + c, errorOptions || {});
       }
     });
-  } else if (typeof opt === 'object') {
+  } else { // opt instanceof Option
```
freitagbr: `opt is object`?

src/common.js
nfischer: Done

src/common.js
nfischer: Done

nfischer:
PTAL

freitagbr:
LGTM, just need to resolve the merge conflicts.

nfischer:
Should be resolved now

PR #679, opened 2017-03-04
Fix cp overwriting identical files

### LINKED ISSUE #678: *CRITICAL data loss* shell.cp() Content of file is erased when trying to copy it to the folder it already belongs to
7.5.0
0.7.5
### Operating system:
Mac OS X
### Description of the bug:
If you try to copy a file to the folder it already belongs to, the content of the file is erased.

```bash
echo "blah" > /home/test
```

```javascript
shell.cp('/home/test', '/home')
```

```bash
cat /home/test
```

`/home/test` is now empty.

freitagbr:
If `cp` is passed a destination that is location of a source file, in effect, copying a file to itself, the source file will become empty. This prevents this situation form happening by verifying that the destination is a different location than the source.

Fixes #678

### DIFF
--- a/src/cp.js
+++ b/src/cp.js
@@ -267,7 +267,7 @@ function _cp(options, sources, dest) {
       if (path.relative(src, thisDest) === '') {
         // a file cannot be copied to itself, but we want to continue copying other files,
         // so save the error for later
-        identicalFiles.push([thisDest, src]);
+        identicalFiles.push("'" + thisDest + "' and '" + src + "' are the same file");
         return;
       }
 
@@ -277,11 +277,9 @@ function _cp(options, sources, dest) {
 
   if (identicalFiles.length > 0) {
     // if any of the copies were identical, print an error for each and then exit
-    identicalFiles.forEach(function (files, i) {
-      var thisDest = files[0];
-      var src = files[1];
+    identicalFiles.forEach(function (err, i) {
       var cont = i < identicalFiles.length - 1; // exit on the last file
-      common.error(thisDest + ' and ' + src + ' are identical (not copied).', { continue: cont });
+      common.error(err, { continue: cont });
     });
   }
 
--- a/src/cp.js
+++ b/src/cp.js
@@ -213,6 +213,7 @@ function _cp(options, sources, dest) {
 
   var destExists = fs.existsSync(dest);
   var destStat = destExists && fs.statSync(dest);
+  var identicalFiles = [];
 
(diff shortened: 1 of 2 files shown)


freitagbr:
Same bug from #673 is affecting the windows tests here. Let's see if we can implement a fix.

src/cp.js
```diff
@@ -260,9 +261,25 @@ function _cp(options, sources, dest) {
         return; // skip file
       }
 
+      if (path.relative(src, thisDest) === '') {
+        // a file cannot be copied to itself, but we want to continue copying other files,
+        // so save the error for later
+        identicalFiles.push("'" + thisDest + "' and '" + src + "' are the same file");
```
freitagbr: Will this still set the `result.stderr`? And `shell.error()` to `true`?

test/cp.js
```diff
@@ -653,3 +653,21 @@ test('recursive, with a non-normalized path', t => {
...
+  t.is(result.stderr, "cp: 'resources/file1' and 'resources/file1' are the same file");
+});
+
+test('copy mutliple files to same location', t => {
+  const result = shell.cp('resources/file1', 'resources/file2', 'resources');
+  t.truthy(shell.error());
+  t.is(result.code, 1);
+  t.is(
+    result.stderr,
+    "cp: 'resources/file1' and 'resources/file1' are the same file\n" +
```
freitagbr: Issue #6? Or issue #681?

freitagbr: This test was fixed by rebasing off of master.

test/cp.js
```diff
@@ -653,3 +653,21 @@ test('recursive, with a non-normalized path', t => {
   t.falsy(result.stderr);
   t.is(result.code, 0);
 });
+
+test('copy file to same location', t => {
+  const result = shell.cp('resources/file1', 'resources');
```
freitagbr: Yes, so I added a test for this.

src/cp.js
nfischer: It sets `result.stderr` to the error output (it's a `+=` operation, so it all collects up). `shell.error()` is some truthy value (it's currently the same string as `result.stderr`, but this is no longer guaranteed as part of the API and it may later change to an error code).

test/cp.js
nfischer: I meant 681. But this comment can be ignored now 😄

freitagbr: I figured. Once this is approved it should be good to merge.

PR #682, opened 2017-03-04
chore: add codecov

nfischer:
Add codecov to ShellJS

Partial fix for #671


nfischer:
@freitagbr from this PR, it looks like codecov integration is a success! This should be safe to merge (it should immediately start tracking coverage on master branch).

PR #684, opened 2017-03-05
fix: convert error output to be consistent cross-platform

### LINKED ISSUE #681: Error output should be consistent across all platforms.
Some ShellJS commands have error output which varies between unix and Windows. File paths are a big issue, since Windows will output `path.join('foo', 'bar')` as `foo\bar`, whereas unix is `foo/bar`.

I think the best approach is to write a helper in `common` which correctly translates `\` to `/` where appropriate (and handles any necessary escaping). Commands should be automatically opted into this behavior. We should discuss if it's responsible to provide an opt-out mechanism (do we really want to promote inconsistent behavior across platforms?).

If we can make this change without modifying any existing test cases, then I think we should treat this as a non-breaking bug and include it in a patch. Otherwise, we'll need to treat it as a breaking change.

This would make testing much easier, so I'm marking this as blocking #671

nfischer:
The error output produced by `shell.error()` or `result.stderr` should
not be inconsistent between platforms. This ensures that path separators
are always printed by ShellJS as `/` instead of as `\` on Windows. This
should allow scripts using ShellJS to be more consistent cross-platform.

We had only one test case which relied on the old behavior, and it's in
`pushd`, which we don't any issues for. Since it's only the error output
that's changed, it's probably a reasonable fix to make.

Fixes #681

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -80,6 +80,17 @@ function log() {
 }
 exports.log = log;
 
+// Converts strings to be equivalent across all platforms. Primarily responsible
+// for making sure we use '/' instead of '\' as path separators, but this may be
+// expanded in the future if necessary
+function convertErrorOutput(msg) {
+  if (typeof msg !== 'string') {
+    throw new TypeError('input must be a string');
+  }
+  return msg.replace(/\\/g, '/');
+}
+exports.convertErrorOutput = convertErrorOutput;
+
 // Shows error message. Throws if config.fatal is true
 function error(msg, _code, options) {
   // Validate input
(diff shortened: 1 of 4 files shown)


nfischer:
> It would be better to just sanitize the paths themselves, but this would be complicated to implement.

Why would that be better? You mean to change `\` ➡️ `/` before passing strings to commands? This would have the same effect on output.

---

For what it's worth, our docs don't actually specify what these error messages may be. We should decide if we really want to consider these part of our public API. In fact, `shell.error()` is only guaranteed to be truthy or falsy, and `.stderr` is only guaranteed to be some sort of error message.

The only real motivation we have for this right now is to land tests, which probably isn't enough to justify something on the fence like this. So feel free to push the milestone back to v0.8.

freitagbr:
I think you've convinced me. In the error messages, \ characters should only occur in the paths. And because the error messages are not documented, there should be no reason to rely on them. This looks good so I'll go ahead and merge it.

freitagbr:
Er, there are merge conflicts.

nfischer:
Rebased off master. Will land this tonight.

nfischer:
I also fixed a TODO which depended on this being landed.

PR #685, opened 2017-03-05
Refactor tests to improve readability

nfischer:
No change in production logic.


nfischer:
Note to self: I killed the redundant appveyor job. One appveyor job is enough.

PR #686, opened 2017-03-05
chore: add codecov script to appveyor CI

nfischer:
Invoke codecov on appveyor builds too, so that we can get coverage for
Windows-specific lines of code.

Partial fix for #671

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,3 +24,6 @@ build: off
 
 test_script:
   - npm test
+
+on_success:
+  - npm run codecov -- -f coverage/lcov.info
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "lint": "eslint .",
     "after-travis": "travis-check-changes",
     "changelog": "shelljs-changelog",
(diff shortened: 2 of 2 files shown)


nfischer:
This is a test to see if codecov is additive: will running codecov on both Windows and Unix take the union of the lines covered? This would be very beneficial for us, so that we can cover lines which handle different versions of Node.js and which handle different platforms.

If this is successful, then we should see that code coverage improves, particularly for commands with lots of Windows-specific code, such as `which()`. Once codecov is done, we should be able to go to the shelljs page and view coverage for this particular branch.

nfischer:
Looks like the codecov package relies on the unix `find` command, which is very different than the Windows `FIND` command. Looks like codecov has disappointed me once again... Hopefully this is something they're working on.

nfischer:
Ok, I think appveyor works now. The tricks I needed were:

 - wrap the `codecov` command with `npm run codecov` (avoids a command-not-found error)
 - specify the `lcov.info` file with codecov's `-f` flag

PR #687, opened 2017-03-07
Add support for removing fifos

### LINKED ISSUE #617: Add support for other file types in rm
`rm` is only able to delete files, directories, and symbolic links. `fs.lstatSync` is able to detect several other file types, however:
```
stats.isBlockDevice()
stats.isCharacterDevice()
stats.isFIFO()
stats.isSocket()
```
Support for these file types should be added (at least for the ones it makes sense to run `rm` on).

freitagbr:
Adds support for removing fifos.

Fixes #617

### DIFF
--- a/test/utils/utils.js
+++ b/test/utils/utils.js
@@ -36,3 +36,13 @@ function sleep(time) {
   child.execFileSync(common.config.execPath, ['resources/exec/slow.js', time.toString()]);
 }
 exports.sleep = sleep;
+
+function mkfifo(dir) {
+  if (process.platform !== 'win32') {
+    const fifo = `${dir}/fifo`;
+    child.execFileSync('mkfifo', [fifo]);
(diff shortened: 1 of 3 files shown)


test/utils/utils.js
```diff
@@ -36,3 +36,13 @@ function sleep(time) {
   child.execFileSync(common.config.execPath, ['resources/exec/slow.js', time.toString()]);
 }
 exports.sleep = sleep;
+
+function mkfifo(dir) {
+  if (process.platform !== 'win32') {
+    const fifo = `${dir}/fifo`;
```
nfischer: Can you use regular string concatenation? This doesn't work for v0.12

freitagbr: Fixed.

freitagbr:
CI failures are unrelated.

PR #688, opened 2017-03-08
Remove contents of symlink to dir with rm -rf

### LINKED ISSUE #587: rm -rf on a symbolic link to a dir deletes its contents
Currently, ShellJS differs from the unix implementation. I added a test in #586 (currently being skipped) to cover this. Once that's merged, we should write a fix for this.

blocked by #586

freitagbr:
Fix the behavior of `rm` such that:

1. `rm('-rf', 'link_to_dir/')` deletes contents of real dir, but keeps dir and symlink
2. `rm('-rf', 'link_to_dir')` deletes the symlink itself

Fixes #587

### DIFF
--- a/src/rm.js
+++ b/src/rm.js
@@ -1,5 +1,6 @@
 var common = require('./common');
 var fs = require('fs');
+var path = require('path');
 
 common.register('rm', _rm, {
   cmdOptions: {
@@ -17,7 +18,7 @@ common.register('rm', _rm, {
 //
 // Licensed under the MIT License
 // http://www.opensource.org/licenses/mit-license.php
-function rmdirSyncRecursive(dir, force) {
+function rmdirSyncRecursive(dir, force, fromSymlink) {
   var files;
 
   files = fs.readdirSync(dir);
@@ -42,6 +43,10 @@ function rmdirSyncRecursive(dir, force) {
       }
     }
   }
+
+  // if was directory was referenced through a symbolic link,
+  // do not remove the directory itself
+  if (fromSymlink) return;
 
   // Now that we know everything in the sub-tree has been deleted, we can delete the main directory.
   // Huzzah for the shopkeep.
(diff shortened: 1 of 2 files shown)


src/rm.js
```diff
@@ -91,6 +96,56 @@ function isWriteable(file) {
...
+    common.error('path is a directory', { continue: true });
+  }
+}
+
+function handleSymbolicLink(file, options) {
+  var stats;
+  try {
+    stats = fs.statSync(file);
+  } catch (e) {
+    // symlink is bad, so just remove it
```
nfischer: Change this to something like "symlink is broken, so only delete the link, not its contents"

src/rm.js
```diff
@@ -91,6 +96,56 @@ function isWriteable(file) {
...
+    if (file[file.length - 1] === '/') {
+      // trailing separator, so remove the contents, not the link
+      if (options.recursive) {
+        // -r was passed, so directory can be removed
+        var link = file.slice(0, -1); // remove the '/' character for readlinkSync
+        var dir = fs.readlinkSync(link);
+        var dirpath = path.resolve(path.dirname(link), dir);
+        var fromSymlink = true;
+        rmdirSyncRecursive(dirpath, options.force, fromSymlink);
+      } else {
```
nfischer: I think you can ignore the `options.recursive` check and use `handleDirectory` instead.

Although this won't catch the "link ➡️ link ➡️ directory" case, it only catches the "link ➡️ directory" case.

src/rm.js
```diff
@@ -43,6 +44,10 @@ function rmdirSyncRecursive(dir, force) {
     }
   }
 
+  // if was directory was referenced through a symbolic link,
+  // do not remove the directory itself
+  if (fromSymlink) return;
```
nfischer: Can you specify that while we do not remove the directory itself, we do in fact remove its contents?

src/rm.js
nfischer: Also, do we really need to do all this work here? Does it really not work to call `fs.readdirSync()` on the link itself? That works for me when I tried on Node v7. Then we probably don't need the `fromSymlink` stuff.

freitagbr: You're right, `fs.readdirSync()` works when passed symlinks, so that can be simplified. I think the `fromSymlink` logic still needs to be there, so that `rmdirSyncRecursive` knows not to delete the directory.

freitagbr:
@nfischer Can you help diagnose the failing test on windows? I don't have a windows machine to test with.

nfischer:
@freitagbr probably has to do with how Windows handles sym links differently. I'd recommend skipping the tests on Windows.

PR #690, opened 2017-03-14
fix(grep, sed, sort, uniq): Split only on newline characters

### LINKED ISSUE #645: Why does sed split files into an array, call replace on each line and rejoin?
v6.4.0

0.7.6

### Operating system:
Mac OS 10.12.2

### Description of the bug:
Using `shelljs.sed` over a folder full of files, if one of the files is a binary file like a png… sed will corrupt the images.

*More details:*
So say a folder has files of the following types (_.js_, _.css_, _.json_, _.png_, _.jpg_). I wanted to use `sed` to replace a string that occurs only in the js, css, json files. I am 100% certain this string would not exist in the binary formats for the png/jpg files so there should be no harm in just running sed over the entire folder. However, because sed attempts to break every file (including the binary files) into an array, split them up on line breaks and merge it back together the binary files come out the other end completely corrupted.

Looking through the history I see that this file splitting/joining happened in a PR that says it is meant to add support for multiple files. I'm not sure that the splitting of individual files should have been part of that or should be part of the code at all. Is this how normal `sed` works? If so then maybe it is appropriate, if not then I can quickly create a PR to resolve this.

```javascript
var buildFiles = shelljs.ls("build/**").filter(f => f.substr(0, 6) === "build/");
shelljs.sed("-i", /apples/g, "oranges", buildFiles);
```

freitagbr:
Splits and joins on newlines only.

Fixes #645 .

### DIFF
--- a/src/grep.js
+++ b/src/grep.js
@@ -47,12 +47,12 @@ function _grep(options, regex, files) {
     }
 
     var contents = file === '-' ? pipe : fs.readFileSync(file, 'utf8');
-    var lines = contents.split('\n');
     if (options.nameOnly) {
       if (contents.match(regex)) {
         grep.push(file);
       }
     } else {
+      var lines = contents.split('\n');
       lines.forEach(function (line) {
         var matched = line.match(regex);
         if ((options.inverse && !matched) || (!options.inverse && matched)) {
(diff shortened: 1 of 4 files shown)


nfischer:
@freitagbr can you see what the issue is with code coverage? If we're uncovering any lines, we should see if we can add tests.

freitagbr:
The diff is +7/-9, so overall -2, so the coverage is going to decrease. No lines are becoming uncovered.

freitagbr:
I think it could be breaking. Just to be safe, let's change the target branch to dev so this can land in v0.8.

PR #702, opened 2017-04-11
Fix #631 throw error when overwriting recently created file

### LINKED ISSUE #631: feat: cp & mv should not overwrite recently created files
If you have a copy command like:

```sh
$ cp file1.txt out/file1.txt destination/
```

You should get an error that says something like:

```
cp: will not overwrite just-created ‘destination/file1.txt’ with ‘out/file1.txt’
```

We're currently silently overwriting files that we created in that same copy command, but should instead provide a warning and refuse to copy the second file.

The behavior should stay the same with `-f` and `-r`, but `cp` does not give the warning with directories or their contents.

```sh
$ cp dir1 out/dir1 destination/ # never has this error
$ ls destination/dir1/ # shows the merged contents of both ./dir1/ and out/dir1/
```

The warning is never present with the `-n` flag.

This warning should also be presented for the `mv` command.

uttpal:
Fixes #631 
If you have a copy command like:

`$ cp file1.txt out/file1.txt destination/`
You should get an error that says something like:

`cp: will not overwrite just-created ‘destination/file1.txt’ with ‘out/file1.txt’`

same behavior goes for mv with -R, -f flags.


test/mv.js
```diff
@@ -206,3 +206,36 @@ test('dest exists, but -f given', t => {
...
+  let result = shell.mv('-n', 'file1', 'cp/file1', 't/');
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+
+  // Ensure First file is copied
+  t.is(shell.cat('t/file1').toString(), 'test1');
+
+  result = shell.mv('t/file1', 'file1'); // revert
+  t.truthy(fs.existsSync('file1'));
+  t.truthy(fs.existsSync('cp/file1'));
```
nfischer: Is this all to manually revert? I think our beforeEach step should be sufficient

src/cp.js
```diff
@@ -150,6 +150,14 @@ function cpdirSyncRecursive(sourceDir, destDir, currentDepth, opts) {
   } // for files
 } // cpdirSyncRecursive
 
+// Checks if cureent file was created recently
+function checkRecentCreated(sources, index) {
```
uttpal: @nfischer Thanks for suggestions.
How about we use the timestamp to check whether a file was created recently or not?
we can set a variable with current time when cp operation starts and then only overwrite files that were created before that.

nfischer: This sounds trickier--if someone is playing with timestamps on their files, we might get different behavior. The goal is to just catch files created by the current `cp` command. I just suggested building a list of files we create, as we create them. Fairly confident this is how GNU `cp` does this.

nfischer: It appears like relying on the parameter list is consistent with GNU `cp`, I must have been mistaken when I tried it

test/cp.js
```diff
@@ -713,3 +713,39 @@ test('copy mutliple files to same location', t => {
...
+  t.is(shell.cat(`${t.context.tmp}/file1`).toString(), 'test1');
+  t.is(
+    result.stderr,
+    `cp: will not overwrite just-created '${t.context.tmp}/file1' with 'resources/cp/file1'`
+  );
+});
+
+
+test('should not overwrite recently created files (in recursive Mode)', t => {
+  const result = shell.cp('-R', 'resources/file1', 'resources/cp/file1', t.context.tmp);
```
nfischer: Never mind, I don't think we need this test

test/mv.js
nfischer: LGTM % this comment

nfischer:
Thanks!

PR #708, opened 2017-04-27
Echo test mocks

### LINKED ISSUE #622: Echo tests unnecessarily run tests in own process
To prevent writing to the process' own stdout, the `echo` tests were written to run each test in its own process, and the stdout of that process was captured. Because ava runs each test in its own process anyway, the `echo` tests no longer need to run in separate processes.

freitagbr:
Fixes #622 

Adds mocks for `process.stdout.write` and `process.stderr.write` that allows the written values to be extracted. This way, we can inspect the output of `echo` without actually outputting to stdout.

### DIFF
--- a/test/echo.js
+++ b/test/echo.js
@@ -51,17 +51,6 @@ test("using null as an explicit argument doesn't crash the function", t => {
   t.is(stderr, '');
 });
 
-test.skip('simple test with silent(true)', t => {
-  // already using shell.config.silent = true
-  const result = shell.echo(555);
-  const stdout = mocks.stdout();
-  const stderr = mocks.stderr();
-  t.falsy(shell.error());
-  t.is(result.code, 0);
-  t.is(stdout, '555\n');
-  t.is(stderr, '');
-});
-
 test('-e option', t => {
   const result = shell.echo('-e', '\tmessage');
   const stdout = mocks.stdout();
--- a/test/echo.js
+++ b/test/echo.js
@@ -2,18 +2,18 @@ import test from 'ava';
 
 import shell from '..';
 import utils from './utils/utils';
-import mocks from './utils/mocks';
+import Mocks from './utils/mocks';
 
 shell.config.silent = true;
 
 test.beforeEach(t => {
   t.context.tmp = utils.getTempDir();
-  mocks.init();
+  t.context.mocks = new Mocks();
 });
 
 test.afterEach.always(t => {
   shell.rm('-rf', t.context.tmp);
-  mocks.restore();
+  t.context.mocks.restore();
 });
 
 //
@@ -22,8 +22,8 @@ test.afterEach.always(t => {
 
 test('simple test with defaults', t => {
   const result = shell.echo('hello', 'world');
-  const stdout = mocks.stdout();
-  const stderr = mocks.stderr();
+  const stdout = t.context.mocks.stdout();
+  const stderr = t.context.mocks.stderr();
   t.falsy(shell.error());
   t.is(result.code, 0);
   t.is(stdout, 'hello world\n');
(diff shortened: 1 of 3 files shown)


test/echo.js
```diff
@@ -2,137 +2,161 @@ import test from 'ava';
...
 });
 
-test.cb('simple test with silent(true)', t => {
-  const script = 'require(\'../global.js\'); config.silent=true; echo(555);';
-  utils.runScript(script, (err, stdout) => {
-    t.falsy(err);
-    t.is(stdout, '555\n');
-    t.end();
-  });
+test.skip('simple test with silent(true)', t => {
```
nfischer: Any reason why we're skipping this?

test/utils/mocks.js
```diff
@@ -0,0 +1,73 @@
+/* eslint-disable prefer-rest-params, strict */
```
nfischer: Why are we disabling linting?

test/utils/mocks.js
```diff
@@ -0,0 +1,73 @@
...
+  // arg
+  if (val === undefined) return stdinValue;
+  stdinValue = val;
+  return null;
+}
+exports.stdin = stdin;
+
+function init() {
+  resetValues();
+  console.log = consoleLog;
```
nfischer: Do we need to mock `console.log` and `console.error`? What happens if we just mock `process.stdout` & `.stderr` objects? That might give us more reliable results overall.

freitagbr: Right, that's probably a better thing to do in this scenario. I actually just copied this over from the shx mocks. I'll reduce this down to just the `process.stdout` and `process.stderr`.

nfischer: Ok, SGTM

test/echo.js
freitagbr: I think because this is testing with `config.silent = true;`, but we already have set `shell.config.silent = true` above on line 7, so this test is redundant.

nfischer: In that case, why not delete it?

test/utils/mocks.js
```diff
@@ -1,73 +1,29 @@
...
-exports.init = init;
-
-function restore() {
-  console.log = oldConsoleLog;
-  console.error = oldConsoleError;
-  process.stdout.write = oldStdoutWrite;
-  process.exit = oldProcessExit;
-}
-exports.restore = restore;
+function Mocks() {
```
nfischer: Any reason we're using mocks as a prototype? I like the idea in general, but we're mocking global values. Using prototypes gives the illusion that multiple mocks can be created, which wouldn't work very well (the second mock would be mocking the first one).

freitagbr: This is a good point. I implemented it as a prototype because I though it was a good way to implement this. But, if you have several mocks going at once, the true `process.stdout.write` will get swapped around in them, and it may be hard to figure out where it went, unless you `restore` in the right order. This seems like it could be problematic, so I will revert this to the global version.

test/echo.js
freitagbr: Sure, I will delete this test.

freitagbr:
Yes, the tests still output during these tests, even with the mocks. In fact, that spinny thing is still shown, so I don't think the mocks interfere with anything.

test/utils/mocks.js
nfischer: Might be good to enforce that `init()` can only be called once though (since calling it twice will lose the real `process.stdout`). Maybe we can just implement this as 2 streams for `process.stdout` and `process.stdin`, then we don't have to reinvent the stream logic.

freitagbr: I've re-implemented the mocks using globals, and calling `init()` or `restore()` any number of times will not lose the real `process.stdout.write`, nor the mock functions

freitagbr:
Also, the reason I merged this into dev instead of master is because of the extra unit tests added in the newline echo branch. I figured it would just be easier to refactor all the tests at once, instead of fixing merge conflicts later on.

PR #709, opened 2017-04-28
Fix common.expand error

### LINKED ISSUE #664: Passing empty string to cp throws internal error
v6.8.0
0.7.6
### Operating system:
Mac OS X
### Description of the bug:
Passing empty string to `cp` command throws an internal error.  Would be much easier to deal with an exception or error return.

```
ShellJS: internal error
Error: must provide pattern
    at new GlobSync (node_modules/glob/sync.js:31:11)
...
   at node_modules/shelljs/src/common.js:246:22
```

```javascript
shelljs.cp('', 'dest');
```

freitagbr:
Fixes #664

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -170,6 +170,24 @@ test('broken links still expand', t => {
   t.deepEqual(result, ['resources/badlink']);
 });
 
+test('empty array', t => {
+  const result = common.expand([]);
+  t.falsy(shell.error());
+  t.deepEqual(result, []);
+});
+
+test('empty string', t => {
+  const result = common.expand(['']);
+  t.falsy(shell.error());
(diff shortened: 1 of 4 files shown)


test/common.js
```diff
@@ -170,6 +170,24 @@ test('broken links still expand', t => {
...
+
+test('empty string', t => {
+  const result = common.expand(['']);
+  t.falsy(shell.error());
+  t.deepEqual(result, ['']);
+});
+
+test('non-string', t => {
+  const result = common.expand([5]);
+  t.falsy(shell.error());
```
nfischer: No need to check `shell.error()`, it's never set by this function (it's internal). Maybe just file a bug to clean up these tests later.

PR #721, opened 2017-05-14
fix(mkdir): improve error handling around files

### LINKED ISSUE #720: ShellJS: internal error on shelljs.mkdir('myFile/myDir')
6.9.2

0.7.7

### Operating system:
Ubuntu 16.04

### Description of the bug:
ShellJS throws the following error when calling `shelljs.mkdir('myFile/myDir')` with an existing file `myFile`:

    ShellJS: internal error
    Error: ENOTDIR: not a directory, mkdir 'myFile/myDir'

This cannot be caught by try...catch, but should be.
Inspired by #473 

```javascript
shelljs.touch('myFile');
shelljs.mkdir('myFile/myDir');
```

nfischer:
In particular, this fixes:

 - if we try to overwrite a file with a mkdir
 - if we try to create a subdirectory of a file
 - adds `continue: true` in spots where we missed it

Fixes #720

### DIFF
--- a/src/mkdir.js
+++ b/src/mkdir.js
@@ -57,9 +57,11 @@ function _mkdir(options, dirs) {
 
   dirs.forEach(function (dir) {
     try {
-      fs.lstatSync(dir);
+      var stat = fs.lstatSync(dir);
       if (!options.fullpath) {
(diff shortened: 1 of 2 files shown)


freitagbr:
LGTM.

PR #725, opened 2017-05-28
Deprecate common.getUserHome, advise using os.homedir instead

### LINKED ISSUE #683: Switch to os.homedir() when we move to v4+
NodeJS added support for `os.homedir()` to get the user's home directory in a reliable way. When we drop support for v4, let's switch to that to reduce the amount of code we need to maintain.

freitagbr:
Deprecates `common.getUserHome` and provides a warning about the deprecation. Also replaces all of the internal uses of `common.getUserHome` with `os.homedir`.

Fixes #683

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -4,7 +4,6 @@
 
 var os = require('os');
 var fs = require('fs');
-var util = require('util');
 var glob = require('glob');
 var shell = require('..');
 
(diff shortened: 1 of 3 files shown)


src/common.js
```diff
@@ -157,6 +158,8 @@ function ShellString(stdout, stderr, code) {
 
 exports.ShellString = ShellString;
 
+// DEPRECATED: Use os.homedir instead
```
nfischer: Don't deprecate internal functions, just remove them. We only explicitly export the functions listed here:

https://github.com/shelljs/shelljs/blob/master/plugin.js#L7

PR #727, opened 2017-05-31
Copying fifos hangs.

bruce-one:
shelljs' `cp` will hang if trying to copy a fifo.

This PR skips copying more non-file-like entities to avoid this issue.

I can't quite decide whether this fix is going to work for a symlink pointing to a fifo? :-s

### DIFF
--- a/src/cp.js
+++ b/src/cp.js
@@ -34,7 +34,9 @@ function copyFileSync(srcFile, destFile, options) {
     // If we're here, destFile probably doesn't exist, so just do a normal copy
   }
 
-  if (fs.lstatSync(srcFile).isSymbolicLink() && !options.followsymlink) {
+  var srcFileStat = fs.lstatSync(srcFile);
+
+  if (srcFileStat.isSymbolicLink() && !options.followsymlink) {
     try {
       fs.lstatSync(destFile);
       common.unlinkSync(destFile); // re-link it
@@ -44,7 +46,7 @@ function copyFileSync(srcFile, destFile, options) {
 
     var symlinkFull = fs.readlinkSync(srcFile);
     fs.symlinkSync(symlinkFull, destFile, isWindows ? 'junction' : null);
-  } else {
+  } else if (srcFileStat.isSymbolicLink() || srcFileStat.isFile() || srcFileStat.isCharacterDevice() || srcFileStat.isBlockDevice()) {
     var buf = common.buffer();
     var bufLength = buf.length;
     var bytesRead = bufLength;
@@ -76,6 +78,8 @@ function copyFileSync(srcFile, destFile, options) {
     fs.closeSync(fdw);
 
     fs.chmodSync(destFile, fs.statSync(srcFile).mode);
+  } else {
+    common.log('copyFileSync: skipping non-file (' + srcFile + ')');
   }
 }
 
--- a/src/cp.js
+++ b/src/cp.js
@@ -46,7 +46,9 @@ function copyFileSync(srcFile, destFile, options) {
 
     var symlinkFull = fs.readlinkSync(srcFile);
     fs.symlinkSync(symlinkFull, destFile, isWindows ? 'junction' : null);
-  } else if (srcFileStat.isSymbolicLink() || srcFileStat.isFile() || srcFileStat.isCharacterDevice() || srcFileStat.isBlockDevice()) {
+  } else if (srcFileStat.isSymbolicLink()) {
+    copyFileSync(fs.realpathSync(srcFile), destFile);
+  } else if (srcFileStat.isFile() || srcFileStat.isCharacterDevice() || srcFileStat.isBlockDevice()) {
     var buf = common.buffer();
     var bufLength = buf.length;
     var bytesRead = bufLength;
--- a/test/cp.js
+++ b/test/cp.js
@@ -772,3 +772,24 @@ test('should not attempt to copy fifos', t => {
   t.is(shell.cat(`${t.context.tmp}/cp/file1`).toString(), 'test1');
   t.falsy(fs.existsSync(`${t.context.tmp}/cp/fifo`));
 });
+
+test('should not attempt to copy fifos via symlinks', t => {
+  shell.mkdir(`${t.context.tmp}/dir`);
+  try {
+    shell.exec(`mkfifo ${t.context.tmp}/dir/fifo`);
+  } catch (e) {
+    console.warn('Exception trying to create fifo. Skipping fifo copy test.');
+    return;
+  }
+  shell.cp('resources/file1', `${t.context.tmp}/dir`);
+  t.truthy(fs.existsSync(`${t.context.tmp}/dir/fifo`));
+  shell.pushd(`${t.context.tmp}/dir`);
+  fs.symlinkSync('fifo', 'symFifo');
+  shell.popd();
+  const result = shell.cp('-rL', `${t.context.tmp}/dir`, `${t.context.tmp}/cp`);
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  t.is(shell.cat(`${t.context.tmp}/cp/file1`).toString(), 'test1');
+  t.falsy(fs.existsSync(`${t.context.tmp}/cp/fifo`));
+  t.falsy(fs.existsSync(`${t.context.tmp}/cp/symFifo`));
+});
--- a/test/cp.js
+++ b/test/cp.js
@@ -755,3 +755,20 @@ test('should not overwrite recently created files (not give error no-force mode)
   // Ensure First file is copied
   t.is(shell.cat(`${t.context.tmp}/file1`).toString(), 'test1');
 });
+
+test('should not attempt to copy fifos', t => {
+  shell.mkdir(`${t.context.tmp}/dir`);
+  try {
+    shell.exec(`mkfifo ${t.context.tmp}/dir/fifo`);
+  } catch (e) {
+    console.warn('Exception trying to create fifo. Skipping fifo copy test.');
+    return;
+  }
+  shell.cp('resources/file1', `${t.context.tmp}/dir`);
+  t.truthy(fs.existsSync(`${t.context.tmp}/dir/fifo`));
+  const result = shell.cp('-r', `${t.context.tmp}/dir`, `${t.context.tmp}/cp`);
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  t.is(shell.cat(`${t.context.tmp}/cp/file1`).toString(), 'test1');
+  t.falsy(fs.existsSync(`${t.context.tmp}/cp/fifo`));
+});


bruce-one:
Hmm, intriguing.

`cp -r` also seems to act differently to a straight copy? (I was copying a directory, and it was what lead me to creating this PR)

eg:

    mkdir a
    mkfifo a/fifo
    cp -r a b # no hang
    ls b # fifo

So this PR isn't properly replicating cp, no... (Although, afaik, node doesn't support creating fifo's; except I guess if this shells out? :-s (Or uses some native bindings))

    execve("/usr/bin/cp", ["cp", "-r", "a", "b"], 0x7ffd9a4e0168 /* 76 vars */) = 0
    ...
    lstat("a/fifo", {st_mode=S_IFIFO|0644, st_size=0, ...}) = 0
    mknod("b/fifo", S_IFIFO|0644)           = 0

bruce-one:
When copying the file directly (eg `cp fifo newFifo`); cp does:

    stat("newFifo", 0x7fffdfb9e570)         = -1 ENOENT (No such file or directory)
    stat("fifo", {st_mode=S_IFIFO|0644, st_size=0, ...}) = 0
    stat("newFifo", 0x7fffdfb9e300)         = -1 ENOENT (No such file or directory)
    open("fifo", O_RDONLY)                  = 3
    fstat(3, {st_mode=S_IFIFO|0644, st_size=0, ...}) = 0
    open("newFifo", O_WRONLY|O_CREAT|O_EXCL, 0644) = 4
    fstat(4, {st_mode=S_IFREG|0644, st_size=0, ...}) = 0
    fadvise64(3, 0, 0, POSIX_FADV_SEQUENTIAL) = -1 ESPIPE (Illegal seek)
    mmap(NULL, 139264, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f0cb3a6f000
    read(3, "q\n", 131072)                  = 2
    write(4, "q\n", 2)                      = 2

vs the following for shelljs:

    open("fifo", O_RDONLY|O_CLOEXEC)        = 9
    open("newFifo", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 10
    pread64(9, 0x26697d0, 65536, 0)         = -1 ESPIPE (Illegal seek)


(The point being read vs pread; wherein `pread` says "The file referenced by fd must be capable of seeking." [1] which is likely the issue here? :-s (This comes from uv by the look of it? [2])

[1]: http://man7.org/linux/man-pages/man2/pread.2.html
[2]: https://github.com/nodejs/node/blob/master/deps/uv/src/unix/fs.c#L318

(I'm just braindumping as I follow some of the rabbit hole :-) )

bruce-one:
Okay, so I think I've fixed the scenario you've referenced :-)

(Setting that offset to `-1` (was zero) means that libuv uses read rather than pread :-) )

freitagbr:
@bruce-one is there documentation on setting the offset to `-1`? I'm not seeing any references to that in the node documentation or source code.

bruce-one:
Ah, no, I just found it in https://github.com/libuv/libuv/blob/v1.x/src/unix/fs.c#L306 (and when looking through the libuv tests -1 is used for the offset quite a bit -- was trying to replicate it in straight uv and stumbled across the difference)

bruce-one:
That switch has been there for a very long time; and it kinda makes sense (I think?) because it's the way to say "no need to worry about offset, so just read not pread":

https://github.com/libuv/libuv/commit/74999f8f9974970e4d71a0df521e178e3149f3df#diff-6a16903c26af4b4035eda9922a73ecc9R160 (via  a few steps in the middle)

bruce-one:
(Sorry for the noise; trying to fix the tests on Windows. It looks like Appvoyer's environment (cygwin?) supports `mkfifo`, but `fs.existsSync` fails to recognise that fifo's exist, I'm assuming it's something to do with how cygwin(?) fakes fifo's, whilst node is trying to go direct?)

test/cp.js
```diff
@@ -797,7 +797,7 @@ test('should not attempt to copy fifos via symlinks in directories', t => {
 test('should copy fifos directly', t => {
   try {
     shell.exec(`mkfifo ${t.context.tmp}/fifo`);
-    shell.exec(`echo -n test1 > ${t.context.tmp}/fifo`, { async: true });
+    shell.exec(`printf test1 > ${t.context.tmp}/fifo`, { async: true });
```
nfischer: I'd prefer to see `shx echo -n test1` here. Appveyor may have the command, but it may not be consistent with unix `printf`, and I don't think it's available by default on Windows systems.

bruce-one: Iirc, osx's echo doesn't support the `-n`... (Which was why I changed it initially place)

Also worth noting named pipes aren't supported directly by Windows (only via cygwin etc, which didn't appear reliable)

(Could match the `\n` though if echo is preferred)

nfischer: > Iirc, osx's echo doesn't support the -n... (Which was why I changed it initially)

`shx echo` is platform agnostic, so it runs on OS X, Win, and Linux. The behavior is guaranteed to be the same across all machines, which is what you want out of an external dependency.

nfischer:
@bruce-one are we confident that `isFIFO()` always returns `false` for Windows, or is this something we need to investigate? I'm hesitant to add features that go untested on Windows--if there's any chance we hit those code paths in production, then we have unpredictable behavior. If we can guarantee that we don't hit that code for that platform, I feel a bit more comfortable.

test/cp.js
bruce-one: Oh yeah :-) good point :-)

Is it possible to easily shell out to shelljs's echo implementation? (I've shelled out here because the event loop is blocked by the cp, so need a way to get data into the fifo in the background; and atm this is shelling out to the /bin/sh impl (iirc?) and using it's echo/printf (although maybe there's a better way?)).

bruce-one:
From what I know/can ascertain (some but not extensive research), Windows can't put named pipes on a filesystem? (a la `mkfifo x; ls # x` Hence I don't know of a situation where isFIFO could return true on Windows.

Hence support for this in native windows isn't possible? (Cygwin emulates pipes, but unless node is built for Windows, I don't think node will report them - from my testing.)

I'm no Windows expert though... :-s

bruce-one:
Another outstanding question with this is in the recursive type scenario: should we call mkfifo or something to create the fifo to better replicate `cp`? (afaict, node can't do that natively)

To illustrate:

    mkdir a
    mkfifo a/fifo
    cp -r a b
    ls b # fifo ; with shx this fifo wouldn't exist as it stands in this pr

nfischer:
> Another outstanding question with this is in the recursive type scenario: should we call mkfifo or something to create the fifo to better replicate cp? (afaict, node can't do that natively)

Seems like the answer would be yes. Our goal here is compatibility with POSIX behavior.

bruce-one:
Have added support for acting like `-R` has been specified (when recursive is true).

From some reading (eg https://unix.stackexchange.com/questions/18712/difference-between-cp-r-and-cp-r-copy-command) `-r` (by contrast to `-R`) is not well defined in this context, however based on the existing code and the GNU implementation I've just made `-r` act like `-R`. I don't know whether adding a `--copy-contents` flag becomes useful now? (It's not posix... From skimming the posix definition there's no equivalent of that GNU flag?)

Unfortunately creating character and block devices requires root permissions... I didn't know how to handle that. (Running ava with sudo and a grep for `root` (ish) after running it as non-root seemed like a tempting solution? But I wasn't sure what was preferred.)

So for now those two tests just don't run on CI :-s (And I haven't even manually tested them on OSX...)

nfischer:
@bruce-one thanks for following up on this

> Unfortunately creating character and block devices requires root permissions...

I would like to steer clear of root-only features in ShellJS core.

After thinking a bit, your initial suggestion may have been right after all: skip these file types by default. We can indicate an error occurred via `common.error()` and copy all the other files.

Since this is a great PR though, I think this would be a good candidate for a plugin. We're adding a plugin API for v0.8, allowing users to extend ShellJS. We could create a plugin to add back support for fifo and other specialized file types, overriding the default behavior.

```js
shell.cp('blockDeviceName', 'dest'); // default behavior
// cp: block device is not supported: 'blockDeviceName'

require('shelljs-plugin-cp-special-files');
shell.cp('blockDeviceName', 'dest');
// Now it succeeds! (assuming we're on Unix and have root permission)
```

ShellJS proper won't need to have any more platform-dependent code, and we can still provide support for the rare users who need it. This sounds like a win. What do you think?

src/cp.js
```diff
@@ -153,6 +171,42 @@ function cpdirSyncRecursive(sourceDir, destDir, currentDepth, opts) {
...
+  args.push('"' + destFile + '"');
+  if (stat.isFIFO()) {
+    args.push('p');
+  } else {
+    if (stat.isCharacterDevice()) args.push('c');
+    if (stat.isBlockDevice()) args.push('b');
+    args.push(major(stat.rdev), minor(stat.rdev));
+  }
+  try {
+    var cmd = exec(args.join(' '), { silent: true });
```
nfischer: For security, I recommend `child_process.execFileSync()` here. Just wrap it in a try-catch, since it throws an exception if the command fails.

```js
try {
  child_process.execFileSync('mknod', argsArray);
} catch (e) {
  failureMessage = e.stderr;
}
```

src/cp.js
```diff
@@ -153,6 +171,42 @@ function cpdirSyncRecursive(sourceDir, destDir, currentDepth, opts) {
...
+function minor(rdev) {
+  return ((rdev >> 12) & 0xffffff00) | (rdev & 0x000000ff);
+}
+
+// attempts to create a dev (character or block) by shelling out to mknod.
+function mknod(destFile, stat) {
+  var created;
+  var failureMessage;
+  var args = ['mknod'];
+  args.push('"' + destFile + '"');
```
nfischer: At first glance this looks wrong. What if `destFile` is `hello world"; rm -rf *; echo "`? Looks like you'll be running:

```sh
$ mknod "hello world"; rm -rf *; echo "" p
```

`JSON.stringify()` is the way we usually perform quote escaping, but you can avoid this entirely if you take my suggestion below.

src/cp.js
```diff
@@ -153,6 +171,42 @@ function cpdirSyncRecursive(sourceDir, destDir, currentDepth, opts) {
   } // for files
 } // cpdirSyncRecursive
 
+function major(rdev) {
```
nfischer: An explanatory comment might be good, since this is pretty niche and not very obvious

bruce-one:
I've done that cleanup based on the feedback from your review (thanks! :-) ).

I agree that making this a plugin does seem reasonable :-)

Hence, should this PR be changed to just log the error message and then independently create a plugin package?

Although, it might be nice to mention the plugin in the error/warning message? Hence maybe make the plugin, then change PR, then merge? Or are you not that keen on "centralising" plugins quite that much? (e.g. if it were called out explicitly (a la `cp: block device is not supported: 'blockDeviceName'` and at the end of the cp run `Special files were encountered during this operation. Add the cp-special-files plugin if support for these is required.`) then it _might_ be nice to have it as shelljs/plugin-cp-special-files? Absolutely no issue if that's not how you want to handle it though :-)

One of the reasons I ask that is because a "shelljs/plugin-posix" could be useful (from past experience) because I know I personally use shelljs solely on posix-like OSs, and find it's consistency excellent (and still very useful), but being able to do "all the posix things" in one plugin could be nice?

nfischer:
> One of the reasons I ask that is because a "shelljs/plugin-posix" could be useful (from past experience) because I know I personally use shelljs solely on posix-like OSs, and find it's consistency excellent (and still very useful), but being able to do "all the posix things" in one plugin could be nice?

Sounds like a good idea. Let's discuss the plugin over at #748. We'll leave this issue for only the cp-specific changes.

src/cp.js
```diff
@@ -54,14 +53,12 @@ function copyFileSync(srcFile, destFile, options) {
...
-      } catch (e) {
-        common.error('copyFileSync: could not set ownership for dest file (' + srcFile + ')');
-      }
-    }
+    var type = 'unknown file';
+    if (srcFileStat.isFIFO()) type = 'fifo';
+    if (srcFileStat.isCharacterDevice()) type = 'block device';
+    if (srcFileStat.isBlockDevice()) type = 'character device';
+    common.error('copyFileSync: ' + type + ' is not supported (' + srcFile + ')', { continue: true });
+    common.logLater('copyFileSync: Special files were encountered during this operation. Please investigate shelljs plugins if you would like to add support for these.');
```
nfischer: I think the `common.error()` log message is sufficient. Let's get rid of `logLater` and move something like this into the docs. You can edit the `//@ ` section in this file, and then run `npm run gendocs`.

test/cp.js
```diff
@@ -757,7 +757,7 @@ test('should not overwrite recently created files (not give error no-force mode)
   t.is(shell.cat(`${t.context.tmp}/file1`).toString(), 'test1');
 });
 
-test('should copy fifos contents', t => {
+test('should copy fifo contents', t => {
```
nfischer: I think this case should also produce a warning. We'll completely drop fifo support and add it back in the plugin.

test/cp.js
```diff
@@ -770,19 +770,19 @@ test('should copy fifos contents', t => {
...
-    t.is(result.code, 0);
-    t.truthy(fs.statSync(`${t.context.tmp}/newFifo`).isFIFO());
+    t.truthy(shell.error());
+    t.is(result.code, 1);
+    t.falsy(fs.existsSync(`${t.context.tmp}/newFifo`));
   }
 });
 
-test('should copy fifos contents via symlinks', t => {
+test('should copy fifo contents via symlinks', t => {
```
nfischer: For consistency, should this also be a warning?

test/cp.js
bruce-one: This fits in the "act like cp" sense.

Being, without the `recursive` flag cp will treat these files as just "dumb files" rather than special files.

From what I'd read, that's the context of where this came from in posix's cp; i.e. originally it would just treat everything as a file and do a dumb copy (creating a normal-file rather than a special-file), but then later on support was added for being clever with special-files. Hence, my brain goes "native shelljs can treat them like normal-files" and "plugin'd shelljs gains the knowledge to treat them as special-files" parallels "normal cp is dumb" and "modern cp -R is smart". If that helps frame why I'd left this in?

There's a conflict between "act like cp" and "pull all the unix out" :-s
I was inclined to "act like cp" (and treat them like normal files, and hence hang while copying) because it's simple and native? Although that could just be me :-)

bruce-one:
(btw, more than happy to just log and not copying when in non-recursive mode, was just explaining why I'd originally left that aspect alone because it feels like sound logic (hence why I'd left it :-p ), but I'm more than happy make the change :-) )

test/cp.js
nfischer: > There's a conflict between "act like cp" and "pull all the unix out" :-s

I agree these goals are in conflict. I think in this case we should try to stay internally consistent. We should output a warning when copying any fifos and delegate to the plugin to provide actual posix behavior.

If the plugin aims for posix compatibility, what arguments does it need passed? It sounds like it needs `options.recursive`, `srcfile`, & `destFile`, is that correct?

bruce-one: Yep, sounds right :-) (Assuming the srcFile symlink has already been resolved if `options.followsymlink === true`.) Perhaps it's worth passing the whole `options` object though? (Thinking in terms of flexibility, eg if more flags are added (eg `-p`))

(And I agree with your perspective on this: the whole reason I opened this PR in the first place was because it tripped me up, and staying internally consistent would have saved me whereas "acting like cp" wouldn't. Just hadn't quite come to that realisation yet :-p )

PR #743, opened 2017-06-18
test(head): improve coverage

### LINKED ISSUE #671: Set up Codecov for the project

### LINKED ISSUE #741: Improve coverage for head()
See [the coverage report](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/head.js)

We're currently missing line coverage for:

 - [Unable to open file](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/head.js#L22)
 - Right-hand-side of a pipe ([here](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/head.js#L70) and [here](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/head.js#L89))

Blocks #671

nfischer:
This adds a test for `head()` on the right-hand side of a pipe. This also
removes the try-catch surrounding `fs.openSync()`, because it was unreachable
code. `fs.existsSync()` guarantees that the file exists, and `fs.openSync()`
only throws if the file does not exist, according to official documentation.

I see 100% line coverage for `src/head.js` when running:

```sh
$ nyc --reporter=text --reporter=lcov ava --serial test/head.js`
```

Fixes #741

### DIFF
--- a/src/head.js
+++ b/src/head.js
@@ -8,20 +8,14 @@ common.register('head', _head, {
   },
 });
 
-// This reads n or more lines, or the entire file, whichever is less.
+// Reads |numLines| lines or the entire file, whichever is less.
 function readSomeLines(file, numLines) {
   var buf = common.buffer();
   var bufLength = buf.length;
   var bytesRead = bufLength;
   var pos = 0;
-  var fdr = null;
 
-  try {
-    fdr = fs.openSync(file, 'r');
-  } catch (e) {
-    common.error('cannot read file: ' + file);
-  }
-
+  var fdr = fs.openSync(file, 'r');
   var numLinesRead = 0;
   var ret = '';
   while (bytesRead === bufLength && numLinesRead < numLines) {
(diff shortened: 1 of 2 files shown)


nfischer:
No functional change, so self-LGTM

Issue #742, opened 2017-06-18
Improve coverage for exec()


nfischer:
See [the coverage report](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/exec.js)

We're currently missing line coverage for:

 - [Using exec without options.silent](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/exec.js#L84)
 - Coverage for #536 ([here](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/exec.js#L138))
 - Commands with non-zero return value/stderr (async exec) ([here](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/exec.js#L142) and [here](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/exec.js#L155))

Blocks #671

PR #746, opened 2017-06-18
chore: add skipOnWin and skipOnUnix test helpers

nfischer:
This adds `skipOnWin` and `skipOnUnix` to help us manage our platform-dependent
tests. These methods give a nice warning message when we skip tests. We may also
consider adding warnings when running platform-dependent tests.

Part of the motivation for this is if we ever update to AVA v0.19. This version
requires at least one assertion per test case. While this could be disabled with
an AVA setting, we instead benefit from warnings for any case when we
unintentionally skip assertions.

This adds chalk as a dev dependency to enable colored messages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
   },
   "devDependencies": {
     "ava": "^0.16.0",
+    "chalk": "^1.1.3",
     "codecov": "^1.0.1",
     "coffee-script": "^1.10.0",
     "eslint": "^2.0.0",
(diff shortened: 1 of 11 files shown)


nfischer:
The coverage report is wrong, this doesn't actually change `exec` coverage. I just didn't rebase off master after #742 landed.

nfischer:
self-LGTM since this doesn't affect production logic (and doesn't actually affect coverage)

PR #760, opened 2017-08-11
refactor(test): update AVA and refactor tests

nfischer:
This updates tests for `AVA` 19.0.0+.

`AVA` 0.18.0 has a breaking change which changes the current working directory
for tests. As a result, we need to change 'resources' -> 'test/resources' (and
 similar path changes).

`AVA` 0.19.0 has a breaking change that all tests must run at least one assert.
This breaking change was already resolved by #746, so no change was necessary in
this PR.

This updates to `AVA` 0.21.0, since there are no other breaking changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "rechoir": "^0.6.2"
   },
   "devDependencies": {
-    "ava": "^0.16.0",
+    "ava": "^0.21.0",
     "chalk": "^1.1.3",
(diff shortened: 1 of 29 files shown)


nfischer:
Thanks. Merging, since something seems broken with appveyor/pr...

@freitagbr make sure you wipe out your `node_modules/` folder in your checkout, otherwise tests won't work with the older versions of `AVA`.

PR #763, opened 2017-08-23
Add "encoding" option to exec

### LINKED ISSUE #675: Cant get encoding buffer to work on exec
7.6.0

"version": "0.7.6"

### Operating system:
Mac OS

### Description of the bug:
Im trying to run 
```git archive --remote repo HEAD keystores/keystore.jks | tar -xO```
But the encoding get fucked up when running through shelljs synchronous command.
I need to have buffer as encoding.

```exec(`git archive --remote=${url} ${branch} ${path}${file} | tar -xO`)```
If you have a keystore anywhere, try that
The keystore becomes invalid.

I have noe used node directly where "buffer" works as encoding

### PR DESCRIPTION
**Pull request by freitagbr**
Fixes #675 by adding support for the `'encoding'` option to `exec`. It technically supported the `'encoding'` option before, but the return values of `stdout` and `stderr` did not honor the encoding. Now, they will.

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -105,11 +105,17 @@ function execSync(cmd, opts, pipe) {
     code = parseInt(fs.readFileSync(codeFile, 'utf8'), 10);
   }
 
-  // fs.readFileSync uses buffer encoding by default, so set
-  // the encoding only if opts.encoding wasn't set to 'buffer'
-  var encoding = opts.encoding !== 'buffer' ? opts.encoding : null;
-  var stdout = fs.readFileSync(stdoutFile, encoding);
-  var stderr = fs.readFileSync(stderrFile, encoding);
+  // fs.readFileSync uses buffer encoding by default, so call
+  // it without the encoding option if the encoding is 'buffer'
+  var stdout;
+  var stderr;
+  if (opts.encoding === 'buffer') {
+    stdout = fs.readFileSync(stdoutFile);
+    stderr = fs.readFileSync(stderrFile);
+  } else {
+    stdout = fs.readFileSync(stdoutFile, opts.encoding);
+    stderr = fs.readFileSync(stderrFile, opts.encoding);
+  }
 
   // No biggie if we can't erase the files now -- they're in a temp dir anyway
   try { common.unlinkSync(scriptFile); } catch (e) {}
--- a/src/exec.js
+++ b/src/exec.js
@@ -160,11 +160,13 @@ function execAsync(cmd, opts, pipe, callback) {
 
 //@
 //@ ### exec(command [, options] [, callback])
-//@ Available options (all `false` by default):
+//@ Available options:
 //@
 //@ + `async`: Asynchronous execution. If a callback is provided, it will be set to
-//@   `true`, regardless of the passed value.
-//@ + `silent`: Do not echo program output to console.
+//@   `true`, regardless of the passed value (default: `false`).
+//@ + `silent`: Do not echo program output to console (default: `false`).
+//@ + `encoding`: Character encoding to use. Affects the returned stdout and stderr values, and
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/exec.js [hunk 1]
```diff
@@ -104,8 +105,11 @@ function execSync(cmd, opts, pipe) {
     code = parseInt(fs.readFileSync(codeFile, 'utf8'), 10);
   }
 
-  var stdout = fs.readFileSync(stdoutFile, 'utf8');
-  var stderr = fs.readFileSync(stderrFile, 'utf8');
+  // fs.readFileSync uses buffer encoding by default, so set
+  // the encoding only if opts.encoding wasn't set to 'buffer'
+  var encoding = opts.encoding !== 'buffer' ? opts.encoding : null;
```
nfischer: What happens if we get rid of the ternary and replace with `encoding = opts.encoding`? It seems weird to specify `'buffer'` if what it actually turns into is `null`

test/exec.js
```diff
@@ -163,6 +163,14 @@ test('exec returns a ShellString', t => {
   t.is(result.toString(), result.stdout);
 });
 
+test('encoding option works', t => {
+  const result = shell.exec(`${JSON.stringify(shell.config.execPath)} -e "console.log(1234);"`, { encoding: 'buffer' });
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  t.truthy(Buffer.isBuffer(result.stdout));
+  t.is(result.stdout.toString(), '1234\n');
```
nfischer: Check `stderr` too?

src/exec.js
freitagbr: Perhaps this would be clearer:
```
var stdout;
var stderr;
if (opts.encoding === 'buffer') {
    stdout = fs.readFileSync(stdoutFile);
    stderr = fs.readFileSync(stderrFile);
} else {
    stdout = fs.readFileSync(stdoutFile, opts.encoding);
    stderr = fs.readFileSync(stderrFile, opts.encoding);
}
```

test/exec.js
freitagbr: Yeah, I will add a `stderr` check here.

src/exec.js
nfischer: I think this is the way to go:

```js
var stdout = fs.readFileSync(stdoutFile, opts.encoding);
var stderr = fs.readFileSync(stderrFile, opts.encoding);
```

freitagbr: That won't work, because the `'buffer'` encoding is only valid for the `child_process` methods:
```
> fs.readFileSync('./package.json', 'buffer')
Error: Unknown encoding: buffer
    at assertEncoding (fs.js:88:11)
    at Object.fs.readFileSync (fs.js:394:3)
    at repl:1:4
    at REPLServer.defaultEval (repl.js:272:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:441:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:203:10)
```
If you need to get a buffer out of the `fs` methods, `encoding` needs to be `null` or `undefined`.

nfischer: Oh, ok. Nevermind, [your proposal](https://github.com/shelljs/shelljs/pull/763#discussion_r134857967) SGTM

freitagbr:
I have tested this with something simple, like `cat foo.tar | tar -xO`, and the output is correct. Maybe we can get @vongohren to chime in here to see if this branch fixes the problem.

vongohren:
Hey we are not using shelljs anymore for this task, so I dont have anything laying around to verify.

freitagbr:
I'm going to go ahead and merge this, since it seems like a useful feature to add.

PR #775, opened 2017-09-21
feat(cat): number output lines (#750)

### LINKED ISSUE #750: cat doesn't support '-n' option
v7.9.0
0.7.8
### Operating system:
osx
### Description of the bug:
cat method don't support options, 
i need to cat a file with line number, but now it dosen't support it
```javascript

```

gcca:
Add `cat -n` support to show number lines. Fix #750 
Additionally this add two tests: 
- Empty file
- File without EOF


src/cat.js
```diff
@@ -37,6 +43,34 @@ function _cat(options, files) {
...
+
+  if (addEOF) {
+    lines.push('');
+  }
+
+  return lines.join('\n');
+}
+
+function number(n) {
+  return ('     ' + n).slice(-6);
```
nfischer: Add a comment? I think this does the right thing (prepends 5 spaces for 1-digit numbers, 4 spaces for 2-digit, 3 spaces for 3-digit, etc.), but it's not super obvious. A one-line comment explaining this would be great.

nfischer: Also, can we rename this function? "number" is vague. Maybe we can rename this to "prependPaddedNumber" and change its functionality to accept both a number and line (and it will insert all padding).

src/cat.js
```diff
@@ -37,6 +43,34 @@ function _cat(options, files) {
...
+  var addEOF = false;
+  var lines = cat.split('\n');
+
+  if (lines[lines.length - 1] === '') {
+    addEOF = true;
+    lines.splice(-1);
+  }
+
+  lines = lines.map(function (line, i) {
+    return number(i + 1) + ' ' + line;
```
nfischer: I think GNU cat uses a tab, not a single space, between the number and line. It also makes for slightly easier parsing. I'm not sure if this is specified by POSIX, however.

test/cat.js
```diff
@@ -64,3 +64,38 @@ test('glob', t => {
...
+});
+
+test('simple numbers without EOF', t => {
+  const result = shell.cat('-n', 'test/resources/cat/file3');
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  t.is(result.toString(), '     1 test3');
+});
+
+test('multiple numbers without EOF', t => {
```
nfischer: Can we add a test to check a file with 10+ lines (just to check that padding works as expected)?

test/cat.js
```diff
@@ -64,3 +64,38 @@ test('glob', t => {
...
+});
+
+test('multiple with numbers', t => {
+  const result = shell.cat('-n', 'test/resources/cat/file2', 'test/resources/cat/file1');
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  t.is(result.toString(), '     1 test2\n     2 test1\n');
+});
+
+test('simple without EOF', t => {
```
nfischer: Thanks for adding! Could you move this test above the numbers section? It would be nice to clump the `-n` tests together.

src/cat.js
```diff
@@ -37,6 +43,34 @@ function _cat(options, files) {
...
   return cat;
 }
 module.exports = _cat;
+
+function addNumbers(cat) {
+  var addEOF = false;
+  var lines = cat.split('\n');
+
+  if (lines[lines.length - 1] === '') {
+    addEOF = true;
```
nfischer: Hmmm can we do something better here? It's kind of silly to strip the last array element simply to add it back at the end.

I'm not certain what the best approach is here.

gcca: I think we can read by chunks from file stream. But even 'readline' package `split` the buffer in lines.
https://github.com/nodejs/node/blob/master/lib/readline.js#L980

src/cat.js
```diff
@@ -37,6 +43,33 @@ function _cat(options, files) {
...
+    cat = addNumbers(cat);
+  }
+
   return cat;
 }
 module.exports = _cat;
+
+function addNumbers(cat) {
+  var lines = cat.split('\n');
+  var lastLine = lines.splice(-1)[0];
```
nfischer: I think lines.pop() is a bit clearer?

test/cat.js
```diff
@@ -64,3 +64,56 @@ test('glob', t => {
...
 });
+
+test('without EOF', t => {
+  const result = shell.cat('test/resources/cat/file3');
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  t.is(result.toString(), 'test3');
+});
+
+test('empty', t => {
```
nfischer: Can you edit the PR description to mention that you're adding 2 additional tests for previous functionality? I assume this is code that worked before--please correct me if that's not so.

gcca: Yes, it has work fine. I add these tests as part of the `-n` implementation.

Issue #753, opened 2017-07-25
Feature request: pushd/popd -q option


nfischer:
Originally suggested in [this comment](https://github.com/shelljs/shelljs/issues/171#issuecomment-317327195) by @MattSturgeon

We might consider implementing the [`-q` option](http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html) (it's available in zsh, not in bash) for `pushd`/`popd`. this option does some zsh-specific things (i.e. the mention of `$chpwd_functions`), but it also suppresses console output.

In ShellJS, we would only want this to suppress console output (ignore the `$chwd_functions` stuff).

alexreg:
Any progress on this lately?

nfischer:
A PR would be welcome (and probably not very difficult). We only call `common.log()` [here](https://github.com/shelljs/shelljs/blob/master/src/dirs.js#L192) and [here](https://github.com/shelljs/shelljs/blob/master/src/dirs.js#L196). All that's needed is a little bit of plumbing to pass command options to `_dirs()`.

---

Workaround:

```js
shell.config.silent = true;
shell.pushd(...);
shell.config.silent = false;
```

nfischer:
Fixed in #777

PR #777, opened 2017-10-04
Added `-q` (quiet) option to `push`, `popd`, `dirs` functions.

alexreg:
Resolves issue #753.

No unit tests added yet; I'm not familiar with the framework, and not sure how best to do them.


src/dirs.js
```diff
@@ -95,7 +97,7 @@ function _pushd(options, dir) {
   }
 
   _dirStack = dirs;
-  return _dirs('');
+  return _dirs(options.quiet ? 'q' : '');
```
nfischer: I think you want `'-q'`, not `'q'`.

src/dirs.js
```diff
@@ -173,6 +178,7 @@ function _dirs(options, index) {
 
   options = common.parseOptions(options, {
     'c': 'clear',
+    'q': 'quiet',
```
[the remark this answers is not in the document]
alexreg: What's the alternative? I thought we already agreed on having a `-q` option, even though it isn't present on any shells (that I know of).

src/dirs.js
alexreg: Thanks!

alexreg:
@nfischer Okay, have a look now, I think things should be in order. :-)

src/dirs.js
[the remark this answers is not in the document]
nfischer: `zsh` has the `-q` flag for `pushd` and `popd`. `dirs` does not have this flag to my knowledge.

test/popd.js
```diff
@@ -114,3 +115,22 @@ test('Test that rootDir is not stored', t => {
   shell.popd(); // no more in the stack
   t.truthy(shell.error());
 });
+
+test('quiet mode', t => {
```
nfischer: Could we also test `stdout` for the case where `-q` is *not* passed? The test should have the same layout.

nfischer:
Oh, this needs docs generated. Please run `npm run gendocs`

alexreg:
> Mostly good.
> 
> I don't have a strong opposition to dirs -q. It makes logical sense compared to pushd/popd. If dirs ever does add a -q flag which differs in behavior, then we'll have to change. @freitagbr thoughts?

Yep, those were my thoughts too.

test/popd.js
alexreg: Almost did this before, but wasn't sure if you'd want them. Done now.

alexreg:
That's all done now. Word of warning: the existing code for `pushd` and `popd` writes to stderr, which differs in behaviour from the shell builtins for bash & zsh, from what I can tell. I've left it writing to stderr, but you may want to change that. (I don't know what function you'd want to call other that `common.log`, if so...)

test/popd.js
```diff
@@ -116,11 +116,30 @@ test('Test that rootDir is not stored', t => {
   t.truthy(shell.error());
 });
 
-test('quiet mode', t => {
+test('quiet mode off', t => {
   try {
     shell.config.silent = false;
+    shell.pushd('test/resources/pushd');
```
nfischer: I think this should be moved up one line, otherwise it will print during the test (which we try to avoid). Same thing for the `quite mode on` test.

nfischer:
> the existing code for pushd and popd writes to stderr, which differs in behaviour from the shell builtins

Yup, you're right. This is unfortunate legacy behavior. Feel free to file a bug and we'll get to it for a breaking release.

alexreg:
It prints before the mock is initialised though. Isn’t that aufficient?

nfischer:
> It prints before the mock is initialised though. Isn’t that aufficient?

You're correct: it doesn't affect correctness, but test results would be difficult to read if we let integration tests print to the console. That's why we use `shell.config.silent` in the first place during our tests.

alexreg:
No problem. That's done now.

PR #780, opened 2017-10-06
chore: set AVA options

nfischer:
This sets 2 AVA options:

 * `serial`: same behavior as the CLI flag, which this replaces
 * `powerAssert`: if an assert fails, it will inspect all objects involved in
   the failed assert. This causes readability issues if `t.falsy(shell.error())`
   breaks, since it inspects all of `shell` (which is way too large).

AVA options are a little easier to manage than CLI options (we only update one
place instead of 2 in `package.json`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
   ],
   "scripts": {
     "posttest": "npm run lint",
-    "test": "nyc --reporter=text --reporter=lcov ava --serial test/*.js",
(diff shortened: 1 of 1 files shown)


nfischer:
The output is still improved over previous AVA output (it's pretty good). It's just catastrophically verbose with the `powerAssert` option set, just because `shell` is a massive object...

PR #783, opened 2017-10-13
Wrap fs.statSync and fs.lstatSync

### LINKED ISSUE #745: Add wrappers around fs.statSync and fs.lstatSync
We should consider adding wrapper methods around these functions:

```js
// src/common.js

exports.statFollowLinks = fs.statSync.bind(fs);
exports.statNoFollowLinks = fs.lstatSync.bind(fs);
```

I'm usually not in favor of meaningless wrapper functions, but I think this could give us big wins for readability. We've had a few bugs (#733 is an example) that boil down to using the wrong `fs.*statSync()`. I personally have trouble remembering which one of those two follows links and which doesn't. Also, I don't give much thought when writing the code about whether or not to follow links, possibly introducing buggy behavior.

If we prefer our own aliases, we don't have to pay much performance penalty, but we remove these ambiguously named functions from our code. Also, since "FollowLinks" is in the name of the method, I'm more likely to double-check the behavior when writing the code. As a reviewer, this will remind me that we need to check behavior for both links and regular files.

We should also investigate writing a [custom eslint rule](http://eslint.org/docs/developer-guide/working-with-rules) to prohibit `fs.statSync` and `fs.lstatSync` in our code.

Finally, we need to scan the code base to make the replacements, possibly changing the link-behavior if we see an actual bug.

@freitagbr what do you think about this?

freitagbr:
Fixes #745 

Adds two new methods to `src/common.js`, `common.statFollowLinks` and `common.statNoFollowLinks`, which wrap `fs.statSync` and `fs.lstatSync`, respectively. This change is meant to improve readability and clarify intent.

### DIFF
--- a/src/cat.js
+++ b/src/cat.js
@@ -36,7 +36,7 @@ function _cat(options, files) {
   files.forEach(function (file) {
     if (!fs.existsSync(file)) {
       common.error('no such file or directory: ' + file);
-    } else if (fs.statSync(file).isDirectory()) {
+    } else if (common.statFollowLinks(file).isDirectory()) {
       common.error(file + ': Is a directory');
     }
 
--- a/src/cd.js
+++ b/src/cd.js
@@ -1,4 +1,3 @@
-var fs = require('fs');
 var os = require('os');
 var common = require('./common');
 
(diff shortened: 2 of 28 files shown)


freitagbr:
It's doubtful. I think V8 will just inline the functions anyway.

PR #784, opened 2017-10-13
Add a test for ls for a single file

### LINKED ISSUE #732: Add test for `ls regular-file.txt`
Originally pointed out in #467 

We should add a test case to `ls` of the sort `ls('resources/ls/file1')`, since I don't think we have a test case to check ls-ing a single file, no options, no wildcards.

freitagbr:
Fixes #732 

Adds a single test to `test/ls.js` that runs `ls` on a single file, which is not covered by the existing tests.

### DIFF
--- a/test/ls.js
+++ b/test/ls.js
@@ -87,6 +87,14 @@ test('simple arg, with a trailing slash', t => {
   t.is(result.length, 6);
 });
 
+test('simple arg, a file', t => {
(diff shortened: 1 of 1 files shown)


freitagbr:
All of the code paths were probably covered by the other tests, but not in the combination that the new test does. it's good having this test around because it tests a common use case for `ls`.

PR #785, opened 2017-10-13
Remove unnecessary shell.error checks from common tests

### LINKED ISSUE #714: Clean up common tests
Some of the tests in `test/common.js` make the check `t.falsy(shell.error())`. This is not necessary, because these methods will not actually set the shell error.

freitagbr:
Fixes #714 

Removes the `t.falsy(shell.error())` checks from methods that can't set the error state.


freitagbr:
The closure returned from `wrap` (i.e. any wrapped function) sets `state.error` to `null`, but those are covered in other tests obviously.

Issue #782, opened 2017-10-12
Harden shell.exec by writing the child process in a source file


nfischer:
Inspired by #524.

Exec has some security concerns, as pointed out in other issues. Due to the API contract, it's inherently vulnerable to shell-command-injection (same as `child_process.exec()`). Additionally, due to its architecture, it's also vulnerable to javascript-code-injection. This is because we write out a string (containing JS code) to a file and exec it.

I don't think we have any known security vulnerabilities here, but this is poor design and easy to mess up in future changes.

We should be able to improve the design while maintaining backwards compatibility by adopting the architecture in #524. We can write the child process code in a source file and pass parameters via IPC (more on that later).

This has 3 main benefits:

 1. Reduces attack surface (no JS injection)
 2. Child process code is more readable
 3. Meaningful line coverage for child process code

Care needs to be taken when passing child parameters. It's not valid to pass values on the shell commandline (shells do weird escape regarding `\n`, `\t`, etc.). I think our best option for IPC is the filesystem. I think IPC modules typically use sockets, but I'm not sure how well these perform for Windows. I think it's acceptable to use the filesystem for now, and come back later with a different IPC system as an enhancement.

nfischer:
Initial PR has landed. Additional improvements:

 - [ ] Swap `child_process.execSync` with `child_process.execFileSync` in `src/exec.js` (this is still backwards compatible, but continues to reduce attack surface)
 - [ ] Replace `codeFile` with actual exit status (reduce file IO)
 - [ ] Consider passing parameters via CLI arg (reduce file IO, this *should* be OK after switching to `execFileSync`)

freitagbr:
I'm not sure replacing `child_process.execSync` with `child_process.execFileSync` will work, at least with the same API. For example, I can do this with `child_process.execSync`:
```
> child_process.execSync('echo hi; echo yo').toString()
'hi\nyo\n'
```
But I cannot do that with `child_process.execFileSync`:
```
> child_process.execFileSync('echo', ['hi; echo yo']).toString()
'hi; echo yo\n'
```

nfischer:
I'm referring to the call within `src/exec.js`. Its responsibility is "launch a subprocess to run exec-child.js". This is an implementation detail, not exposing the API.

The API depends on the call to `child_process.exec()` from within `src/exec-child.js`. That's the part which cannot be changed.

nfischer:
[...]

PR #786, opened 2017-10-13
refactor(exec): move child process to source file

nfischer:
This PR refactors `shell.exec()` by putting its child process in a separate code
file. This also slightly cleans up dead code.

There's more potential to clean this up (e.g. exit status), but this is a good
enough start.

Issue #782

### DIFF
--- a/src/exec-child.js
+++ b/src/exec-child.js
@@ -1,3 +1,7 @@
+if (require.main !== module) {
+  throw new Error('This file should not be required');
+}
+
 var childProcess = require('child_process');
 var fs = require('fs');
 
--- a/src/exec-child.js
+++ b/src/exec-child.js
@@ -1,0 +1,42 @@
+var childProcess = require('child_process');
+var fs = require('fs');
+
+// Note: this will break if `paramFilePath` contains special characters ( '\n',
+// '\t', etc.). Such characters are possible if $TMP gets modified. We already
+// rely on tempdir() to work for other things, so this is an acceptable risk.
+var paramFilePath = process.argv[2];
+
+var serializedParams = fs.readFileSync(paramFilePath, 'utf8');
+var params = JSON.parse(serializedParams);
+
+var cmd = params.command;
+var execOptions = params.execOptions;
+var pipe = params.pipe;
+var stdoutFile = params.stdoutFile;
+var stderrFile = params.stderrFile;
+var codeFile = params.codeFile;
+
+var c = childProcess.exec(cmd, execOptions, function (err) {
+  if (!err) {
+    fs.writeFileSync(codeFile, '0');
+  } else if (err.code === undefined) {
+    fs.writeFileSync(codeFile, '1');
+  } else {
+    fs.writeFileSync(codeFile, err.code.toString());
+  }
+});
+
+var stdoutStream = fs.createWriteStream(stdoutFile);
+var stderrStream = fs.createWriteStream(stderrFile);
+
+c.stdout.pipe(stdoutStream, { end: false });
+c.stderr.pipe(stderrStream, { end: false });
+c.stdout.pipe(process.stdout);
+c.stderr.pipe(process.stderr);
+
+if (pipe) {
+  c.stdin.end(pipe);
+}
+
+c.stdout.on('end', stdoutStream.end);
+c.stderr.on('end', stderrStream.end);
--- a/src/exec.js
+++ b/src/exec.js
@@ -24,10 +24,10 @@ function execSync(cmd, opts, pipe) {
   }
 
   var tempDir = _tempDir();
-  var stdoutFile = path.resolve(tempDir + '/' + common.randomFileName());
-  var stderrFile = path.resolve(tempDir + '/' + common.randomFileName());
   var codeFile = path.resolve(tempDir + '/' + common.randomFileName());
-  var scriptFile = path.resolve(tempDir + '/' + common.randomFileName());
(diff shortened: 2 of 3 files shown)


nfischer:
@freitagbr I will also land separate CLs (against the same issue) for:

 - Remove `codeFile` and rely on exit status (`process.exit`) instead. Reduces file IO!
 - Remove one of the `execSync` calls (the one in `exec.js`) and replace with `execFileSync` for security. Also helps handle special characters (improves correctness on systems with unusual configurations).

nfischer:
@freitagbr ping

src/exec-child.js
```diff
@@ -0,0 +1,42 @@
...
+c.stderr.pipe(stderrStream, { end: false });
+c.stdout.pipe(process.stdout);
+c.stderr.pipe(process.stderr);
+
+if (pipe) {
+  c.stdin.end(pipe);
+}
+
+c.stdout.on('end', stdoutStream.end);
+c.stderr.on('end', stderrStream.end);
```
freitagbr: If this file is `require`'d, it will actually do something. To prevent this, wrap the contents like this:
```
if (require.main === module) {
    // everything goes here
}
```
This way, the code is only executed when the file is the entrypoint, like when `node exec-file.js` is run.

src/exec.js
```diff
@@ -37,47 +37,29 @@ function execSync(cmd, opts, pipe) {
...
+  var paramsToSerialize = {
+    command: cmd,
+    execOptions: opts,
+    pipe: pipe,
+    stdoutFile: stdoutFile,
+    stderrFile: stderrFile,
+    codeFile: codeFile,
+  };
+
+  fs.writeFileSync(paramsFile, JSON.stringify(paramsToSerialize), 'utf8');
```
freitagbr: What is the advantage of storing the params in a separate file, as opposed to simply passing them in as a argument to the command?

src/exec.js
```diff
@@ -91,10 +73,10 @@ function execSync(cmd, opts, pipe) {
     child.execSync(execCommand, opts);
   } catch (e) {
     // Clean up immediately if we have an exception
-    try { common.unlinkSync(scriptFile); } catch (e2) {}
-    try { common.unlinkSync(stdoutFile); } catch (e2) {}
-    try { common.unlinkSync(stderrFile); } catch (e2) {}
     try { common.unlinkSync(codeFile); } catch (e2) {}
+    try { common.unlinkSync(paramsFile); } catch (e2) {}
+    try { common.unlinkSync(stderrFile); } catch (e2) {}
+    try { common.unlinkSync(stdoutFile); } catch (e2) {}
```
freitagbr: We're not using `e2` anyway, might as well call it `e` (it won't overwrite the `e` from above).

src/exec-child.js
nfischer: Seems reasonable. Do you think it's worth adding this to similar files (e.g. `bin/shjs`)?

src/exec.js
nfischer: I think there's an eslint warning against that...

src/exec.js
nfischer: That was my original approach, but there's some issues with passing via CLI argument (at least in shells, via `execSync`). As an example, we could look at the multi-line string:

```
has"chars
.txt
```

`JSON.stringify` gives us `"has\"chars\n.txt"`. So far, so good.

When this gets passed to the shell, it looks something like:

```sh
node exec-child.js "has\"chars\n.txt"
```

The shell interprets the string and supplies the interpreted value to the node process, so `process.argv[2]` looks like `has"chars\n.txt` (that's a one-line string with a literal `\` followed by a literal `n`).

This is where we hit issues. Our string is partially deserialized, so we can't use `JSON.parse`, but it's still not equivalent to our original string. This is an unfortunate consequence of using both shell and JS--they have different opinions on how to escape strings.

---

We can consider passing these by CLI argument if we replace `execSync()` with `execFileSync()` within `exec.js`. This is one of my intended future optimizations, but I left it out to limit complexity. Such a change is backwards compatible (we just can't change `exec()` within `exec-child.js`).

src/exec-child.js
freitagbr: Yes, I think this would be a good idea for other executable-only files.

nfischer:
@freitagbr take a look at my comments. I'll upload a change tonight for your first comment (I don't think there's any work needed for the other two).

src/exec.js
freitagbr: Oh, right.

src/exec.js
freitagbr: Ok, that makes sense. We eliminate the need to write the JS file every call, but in return we need to write the arguments to a file. Not the best, but it's an improvement overall.

src/exec-child.js
nfischer: Filed #789. I think it would be better to fail loudly in this case (with an exception). Take a look at my alternate suggestion, let me know if there's some issue with this.

src/exec.js
nfischer: This is still strictly better. The JS file used to contain each argument written inside it (as literals), and `sizeof(code) + sizeof(arguments) > sizeof(arguments)`.

I'll also send some PRs to trim down the number of arguments, which will be a bigger win.

src/exec-child.js
freitagbr: Failing loudly sounds like a good idea.

nfischer:
PTAL @freitagbr

freitagbr:
LGTM

PR #790, opened 2017-10-20
Use execFileSync to launch child process

nfischer:
This uses `child_process.execFileSync` instead of `execSync` to launch the child
process. This further reduces the attack surface, removing a possible point for
command injection in the ShellJS implementation.

This does not affect backwards compatibility for the `shell.exec` API (the
behavior is determined by the call to `child_process.exec` within
`src/exec-child.js`).

Issue #782


freitagbr:
@nfischer LGTM

PR #791, opened 2017-10-20
Remove codeFile parameter

nfischer:
This parameter isn't needed, we can easily rely on exit code status for this.

Eliminating the parameter reduces file IO, code complexity, and removes a busy
loop.

Issue #782

### DIFF
--- a/src/exec-child.js
+++ b/src/exec-child.js
@@ -19,8 +19,9 @@ var pipe = params.pipe;
 var stdoutFile = params.stdoutFile;
 var stderrFile = params.stderrFile;
 
+var c;
 try {
-  var c = childProcess.exec(cmd, execOptions, function (err) {
+  c = childProcess.exec(cmd, execOptions, function (err) {
     if (!err) {
       process.exit(0);
     } else if (err.code === undefined) {
--- a/src/exec-child.js
+++ b/src/exec-child.js
@@ -32,6 +32,7 @@ try {
   });
 } catch (e) {
   // child_process could not run the command.
+  /* istanbul ignore next */
   process.exitCode = 127;
 }
 
--- a/src/exec-child.js
+++ b/src/exec-child.js
@@ -23,29 +23,26 @@ var c;
 try {
   c = childProcess.exec(cmd, execOptions, function (err) {
     if (!err) {
-      process.exit(0);
+      process.exitCode = 0;
     } else if (err.code === undefined) {
-      process.exit(1);
+      process.exitCode = 1;
     } else {
-      process.exit(err.code);
+      process.exitCode = err.code;
     }
   });
 } catch (e) {
   // child_process could not run the command.
-  process.exit(127);
+  process.exitCode = 127;
 }
 
 var stdoutStream = fs.createWriteStream(stdoutFile);
 var stderrStream = fs.createWriteStream(stderrFile);
 
-c.stdout.pipe(stdoutStream, { end: false });
-c.stderr.pipe(stderrStream, { end: false });
+c.stdout.pipe(stdoutStream);
+c.stderr.pipe(stderrStream);
 c.stdout.pipe(process.stdout);
 c.stderr.pipe(process.stderr);
 
 if (pipe) {
   c.stdin.end(pipe);
 }
-
-c.stdout.on('end', stdoutStream.end);
-c.stderr.on('end', stderrStream.end);
--- a/src/exec-child.js
+++ b/src/exec-child.js
@@ -18,17 +18,21 @@ var execOptions = params.execOptions;
 var pipe = params.pipe;
 var stdoutFile = params.stdoutFile;
 var stderrFile = params.stderrFile;
-var codeFile = params.codeFile;
 
-var c = childProcess.exec(cmd, execOptions, function (err) {
-  if (!err) {
-    fs.writeFileSync(codeFile, '0');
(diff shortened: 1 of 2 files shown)


src/exec-child.js
```diff
@@ -18,17 +18,22 @@ var execOptions = params.execOptions;
...
+      process.exit(0);
+    } else if (err.code === undefined) {
+      process.exit(1);
+    } else {
+      process.exit(err.code);
+    }
+  });
+} catch (e) {
+  // child_process could not run the command.
+  process.exit(127);
```
freitagbr: `process.exit` is potentially a bad idea if there are still callbacks in the queue, or if there are async tasks running, because they will not finish. Of course, at this point, neither of these should be the case, but I still think setting `process.exitCode` is a safer way to do this.

freitagbr: Also, below this, we are manually ending `stdoutStream` and `stderrStream` when `c.stdout` and `c.stderr` end. Why do they need to be ended manually?

nfischer: Ack on `process.exitCode`, sounds like it will be safer.

I'm not sure if we need to manually end the stream. This is legacy code. What would end the streams otherwise, though?

freitagbr: I would need to test this to make sure, but just a normal `pipe` should work:
```
c.stdout.pipe(stdoutStream);
c.stderr.pipe(stderrStream);
```

nfischer: I don't think we can do that. We need things as-they-are in order to:

1. Send  output to stdio in real-time
2. Save that output in a file (for when `exec-child.js` finishes)

If we pipe the output, then we can't achieve bullet 2.

freitagbr: You can just pipe twice:
```
c.stdout.pipe(fs.createWriteStream(stdoutFile));
c.stderr.pipe(fs.createWriteStream(stderrFile));
c.stdout.pipe(process.stdout);
c.stderr.pipe(process.stderr);
```

nfischer: Nevermind, I was confused. You're right 👍

nfischer:
Comments are addressed, I'll look into the CI failures

nfischer:
@freitagbr this is ready for review (tests should pass, the last commit is just a comment)

freitagbr:
For reference, I found a few commits related to the `{ end: false }` option passed for the pipes:

- [86e2d57c](https://github.com/shelljs/shelljs/commit/86e2d57cbc6df94033e603c8d7549a894cd72216#diff-c3bfabb5e6987aa21bc75ffd95a162d6)
- [ab142f52](https://github.com/shelljs/shelljs/commit/ab142f522f07965f352bf1f7e56a486696025b59#diff-c3bfabb5e6987aa21bc75ffd95a162d6)
- [3d01d643](https://github.com/shelljs/shelljs/commit/3d01d643852e812fff13088333f61c91b900c263#diff-c3bfabb5e6987aa21bc75ffd95a162d6)

The way we have it currently is basically identical to piping without, `{ end: false }`, so I'm not sure if the original issues will crop up again. Perhaps it was an issue with node itself.

src/exec-child.js
```diff
@@ -18,29 +18,32 @@ var execOptions = params.execOptions;
...
+      process.exitCode = 1;
+    } else {
+      process.exitCode = err.code;
+    }
+  });
+} catch (e) {
+  // child_process could not run the command.
+  /* istanbul ignore next */
+  process.exitCode = 127;
+}
```
freitagbr: If `childProcess.exec` throws, then `c` will be `undefined`, right? That means the code below this try/catch will fail.

nfischer:
> The way we have it currently is basically identical to piping without { end: false }, so I'm not sure if the original issues will crop up again. Perhaps it was an issue with node itself.

Thanks for researching. Looking back, we used to pipe stdout & stderr into the same file (externally visible as `shell.exec().output`). I changed this [a while back](https://github.com/shelljs/shelljs/pull/268#issuecomment-167981329) to separate the two streams (so we have `.stdout` and `.stderr`). Now that we have 2 readable streams piping into 2 writable streams, we can use default stream-closing behavior.

src/exec-child.js
nfischer: I think you're right. We should actually think of valid cases which might trigger this and document them. I think `maxBuffer` normally triggers this, but we're hitting the `maxBuffer` on the `execFileSync()` call in `exec.js` first.

freitagbr: Looking at the current code for [`childProcess.exec`](https://github.com/nodejs/node/blob/master/lib/child_process.js#L137), I'm not sure it can throw synchronous exceptions. Not sure about past versions of node, however.

nfischer: Hmmm you may be right.

I tried adding all sorts of broken/invalid cases and couldn't actually trigger this.

nfischer: Removed it.

nfischer:
@freitagbr I think everything is resolved.

src/exec-child.js
freitagbr: Just for the sake of verifying, I looked through the history of `childProcess.exec` and it hasn't changed much in 2 years, basically since `v0.12`. So, it is safe to remove the `try/catch`

freitagbr:
LGTM, I restarted the Travis builds for good measure.

nfischer:
Travis had problems last night 😞

PR #797, opened 2017-10-30
chore(eslint): use words instead of numbers

nfischer:
Eslint rules can be configured either using words or number values:

 * "off" or 0
 * "warn" or 1
 * "error" or 2

This switches our config to use the string values instead of the number
values, since the number values are too cryptic.

No change to our actual settings.


freitagbr:
LGTM

PR #801, opened 2017-11-10
docs: add a section on the plugin API

### LINKED ISSUE #391: Plugin System.

nfischer:
No change to logic, docs change only.

Now that the plugin API is finalized, we should link to the wiki page
documenting the plugin API.

Fixes #391

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,15 @@ var shell = require('shelljs');
 shell.echo('hello world');
 ```
 
+## Plugin API
+
+If you'd like to add your own ShellJS commands, we now support plugins! Check
+out the wiki pages to see:
+
+ - [How to import existing plugins in your project](https://github.com/shelljs/shelljs/wiki/Using-ShellJS-Plugins#using-an-existing-plugin)
+ - [How to write your own plugin](https://github.com/shelljs/shelljs/wiki/Using-ShellJS-Plugins#writing-a-plugin)
+ - [A list of known-plugins](https://github.com/shelljs/shelljs/wiki/Derivative-works-and-related-projects#plugins)
+
 <!-- DO NOT MODIFY BEYOND THIS POINT - IT'S AUTOMATICALLY GENERATED -->
 
 


nfischer:
@freitagbr can you also give thoughts on [this section](https://github.com/shelljs/shelljs/wiki/Using-ShellJS-Plugins#pluginregister-options)?

My concerns:

 - it's weird that `cmdOptions` is 2 different types. Maybe we can use `{}` to signal "no options"
 - `unix` is an awful property name. I'm not even sure this functionality should be exposed publicly.
 - We should fail loudly during plugin registration an unknown plugin-option is used.
   - Also fail loudly for bad values (e.g. `globStart: true`)
   - Also fail loudly for invalid combos (e.g. `globStart: 2` with `allowGlobbing: false`)
 - We should consider auto-generating this table from in-line documentation (this would require moving it to the README, but I don't think this needs to block release).

freitagbr:
Nice, I like this change.

Regarding the wiki section:
- For `cmdOptions`, using a boolean usually indicates that `true` and `false` are both accepted values. In the case of `cmdOptions`, passing `true` doesn't really make sense. Perhaps `null` would be a better fit here. Or, as you mentioned, and empty object. Also, are we trying to distinguish between "ignoring any flags passed" and "not allowing any flags to be passed"? `null` and `{}` might indicate the latter, but not passing `cmdOptions` at all might indicate the former.
- Yeah, `unix` isn't super descriptive, and it sounds like we don't remember it's purpose. I would err on the side of removing it, we can always add it later in a minor update.
-  Yeah, failing loudly is a good thing to do for plugin instantiation.
- I agree, the documentation should be auto-generated, just like the other APIs.

nfischer:
Closed in favor of #812

PR #802, opened 2017-11-10
Remove separate "internal error" from exec

### LINKED ISSUE #734: Merge 'exec: internal error' with ShellJSInternalError
I can't actually remember why we have both `ShellJSInternalError` and `exec: internal error`, but they mean the same thing: there's a bug in ShellJS code somewhere. These should probably get merged together, so that we can actually keep the exception stack for bugs in `exec`. This would help immensely in debugging.

While it's true that `exec()` sometimes crashes unavoidably (like if we overrun the max buffer), it's definitely still helpful to have the stack trace, and it still makes sense to report it as an exception, instead of with `shell.error()` (that should be reserved for the external command's stderr).

freitagbr:
Fixes #734 

Removes `common.error('internal error')` from `exec`, allowing any internal errors in `exec` to be wrapped by `ShellJSInternalError` instead.

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -77,14 +77,14 @@ function execSync(cmd, opts, pipe) {
 
   // fs.readFileSync uses buffer encoding by default, so call
   // it without the encoding option if the encoding is 'buffer'
-  var stdout;
-  var stderr;
+  var stdout = '';
+  var stderr = '';
   if (opts.encoding === 'buffer') {
-    stdout = fs.readFileSync(stdoutFile);
-    stderr = fs.readFileSync(stderrFile);
+    try { stdout = fs.readFileSync(stdoutFile); } catch (e) {}
+    try { stderr = fs.readFileSync(stderrFile); } catch (e) {}
   } else {
-    stdout = fs.readFileSync(stdoutFile, opts.encoding);
-    stderr = fs.readFileSync(stderrFile, opts.encoding);
+    try { stdout = fs.readFileSync(stdoutFile, opts.encoding); } catch (e) {}
+    try { stderr = fs.readFileSync(stderrFile, opts.encoding); } catch (e) {}
   }
 
   // No biggie if we can't erase the files now -- they're in a temp dir anyway
--- a/src/exec.js
+++ b/src/exec.js
@@ -6,6 +6,7 @@ var fs = require('fs');
 var child = require('child_process');
 
 var DEFAULT_MAXBUFFER_SIZE = 20 * 1024 * 1024;
+var DEFAULT_ERROR_CODE = 1;
 
 common.register('exec', _exec, {
   unix: false,
@@ -72,7 +73,7 @@ function execSync(cmd, opts, pipe) {
     child.execFileSync(common.config.execPath, execArgs, opts);
   } catch (e) {
     // Commands with non-zero exit code raise an exception.
-    code = e.status;
+    code = e.status || DEFAULT_ERROR_CODE;
   }
 
   // fs.readFileSync uses buffer encoding by default, so call
--- a/src/exec.js
+++ b/src/exec.js
@@ -76,15 +76,17 @@ function execSync(cmd, opts, pipe) {
   }
 
   // fs.readFileSync uses buffer encoding by default, so call
-  // it without the encoding option if the encoding is 'buffer'
+  // it without the encoding option if the encoding is 'buffer'.
+  // Also, if the exec timeout is too short for node to start up,
+  // the files will not be created, so these calls will throw.
   var stdout = '';
   var stderr = '';
   if (opts.encoding === 'buffer') {
-    try { stdout = fs.readFileSync(stdoutFile); } catch (e) {}
-    try { stderr = fs.readFileSync(stderrFile); } catch (e) {}
+    stdout = fs.readFileSync(stdoutFile);
+    stderr = fs.readFileSync(stderrFile);
   } else {
-    try { stdout = fs.readFileSync(stdoutFile, opts.encoding); } catch (e) {}
-    try { stderr = fs.readFileSync(stderrFile, opts.encoding); } catch (e) {}
+    stdout = fs.readFileSync(stdoutFile, opts.encoding);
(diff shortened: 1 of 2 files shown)


test/exec.js
```diff
@@ -35,7 +35,7 @@ test('config.fatal and unknown command', t => {
   shell.config.fatal = true;
   t.throws(() => {
     shell.exec('asdfasdf'); // could not find command
-  }, /exec: internal error/);
+  }, /((asdfasdf: not found)|(asdfasdf: command not found)|('asdfasdf' is not recognized))/);
```
nfischer: Why 3 regexes? If these are different shells, let's put a comment.

Should we check it's a shelljsinternalerror?

test/exec.js
```diff
@@ -127,8 +127,10 @@ test('set timeout option', t => {
   const result = shell.exec(`${JSON.stringify(shell.config.execPath)} test/resources/exec/slow.js 100`); // default timeout is ok
   t.falsy(shell.error());
   t.is(result.code, 0);
-  shell.exec(`${JSON.stringify(shell.config.execPath)} test/resources/exec/slow.js 100`, { timeout: 10 }); // times out
-  t.truthy(shell.error());
+  const err = t.throws(() => {
+    shell.exec(`${JSON.stringify(shell.config.execPath)} test/resources/exec/slow.js 100`, { timeout: 10 }); // times out
+  }, /ENOENT: no such file or directory/);
```
nfischer: Oh, let's actually debug this. This looks like a real bug

nfischer:
Wow, this is a huge improvement. For curiosity, what do the stack traces look like for these errors?

src/exec.js
```diff
@@ -77,14 +77,14 @@ function execSync(cmd, opts, pipe) {
...
   // fs.readFileSync uses buffer encoding by default, so call
   // it without the encoding option if the encoding is 'buffer'
-  var stdout;
-  var stderr;
+  var stdout = '';
+  var stderr = '';
   if (opts.encoding === 'buffer') {
-    stdout = fs.readFileSync(stdoutFile);
-    stderr = fs.readFileSync(stderrFile);
+    try { stdout = fs.readFileSync(stdoutFile); } catch (e) {}
```
nfischer: Is this really the right fix? Why aren't the files created by the child process? I think that's where the bug really is.

freitagbr: In the unit test specifically, the timeout is set to 10 ms, and node doesn't startup in that time. The code in `exec-child.js` is what is creating the stdout and stderr files, and so if node hasn't started, then the files will not exist after the `execSync` call exits.

Perhaps instead the files should be created before the `execSync` call, and the child process just writes to them?

nfischer: This sort of bug usually means that the `exec()` call inside `exec-child.js` throws an exception:

https://github.com/shelljs/shelljs/blob/8ab0a3a3931b59215553730ad86adef8b21a0fa0/src/exec-child.js#L19

nfischer: Oh, didn't see your last message. Huh, if `node` doesn't startup, then I think we have a few options:

 1. Don't try to handle this, let it be an internal error
 2. Modify the test's timeout so that we're more confident `node` can start up
 3. Create files beforehand (but this isn't really necessary in the common case, seems like a waste)

nfischer: This test case is all about checking the `timeout` option works. So we can also try something like:

```
echo start
sleep 5
echo stop # Check that this doesn't print, because we timeout before this
```

freitagbr: Sorry for the delay, I'll work on this soon.

nfischer: No rush. I think it's fine to wrap this into v0.9 anyway

freitagbr: The test passes when the timeout is set to 500 ms (minus the `try/catch` statements).If you think it's alright, I will just increase the timeout on the test.

test/exec.js
```diff
@@ -35,7 +35,11 @@ test('config.fatal and unknown command', t => {
   shell.config.fatal = true;
   t.throws(() => {
     shell.exec('asdfasdf'); // could not find command
-  }, /exec: internal error/);
+    // the expected message depends on where tests are run:
+    //   /bin/sh says "foo: not found"
+    //   /bin/bash says "foo: command not found"
```
nfischer: We only need to support `/bin/sh` and `cmd.exe`, since these are the default shells. I think error messages are implementation specific, so let's just limit it to those 2 if we can get away with it (so long as mac, linux, and windows bots are happy).

src/exec.js
nfischer: You mean to revert the try-catch statements here and to change the test timeout instead? I think that's fine.

freitagbr: Ok, I'll go ahead with this.

test/exec.js
freitagbr: On my machine, at least, `/bin/bash` is the default shell.

nfischer: How about we regex for `asdfasdf` instead? I think that's good enough.

nfischer:
@freitagbr any idea why we have travis failures?

nfischer:
Ah, I tried it locally. It looks like `e.status` is 0. Must be a behavior change between node 5/6. We probably shouldn't let `code` be 0 if an exception was thrown. Maybe we can try `code = e.status || DEFAULT_ERROR_CODE`.

nfischer:
LGTM. Could you add a comment explaining that `.status` is sometimes 0, even when this is in fact an error, and that `DEFAULT_ERROR_CODE` was chosen arbitrarily?

nfischer:
I don't think this is really breaking anything. The behavior changes I see are:

 - `exec()` used to abort in an unhelpful way when it had bugs. The bugs are still there, but now it aborts in a helpful way
   - If someone depended on aborting-without-exception and are now broken, it means they were doing something which caused an `exec()` bug to begin with. We don't support bugs
 - `exec()` used to return a shellstring with an empty `.stderr` attribute. Now its `.stderr` is useful

If you agree, feel free to merge

joelchoo:
@nfischer I'm really interested in getting this fix. Any indication of when 0.8.2 will be coming out?

PR #804, opened 2017-11-12
refactor: harden plugins against unknown options

nfischer:
This reworks the plugin API such that:

 - Unable to register a command with unknown wrap-options
 - `TypeError` raised for wrap-option type mistakes
 - Remove the `overWrite` option (it's unused, probably safest to not
   expose for now)
 - `cmdOptions` defaults to `null` instead of `false` for type
   consistency (no change to default behavior)
 - Move `pipeMethods` logic into `_register`, since it makes more sense
   there

This is not expected to have any effect on existing plugins.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -52,9 +52,6 @@ exports.state = state;
 
 delete process.env.OLDPWD; // initially, there's no previous directory
 
-// This is populated by calls to commonl.wrap()
-var pipeMethods = [];
-
 // Reliably test if something is any sort of javascript object
 function isObject(a) {
   return typeof a === 'object' && a !== null;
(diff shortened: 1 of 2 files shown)


src/common.js
```diff
@@ -428,22 +422,36 @@ exports.readFromPipe = _readFromPipe;
...
   allowGlobbing: true,
   canReceivePipe: false,
-  cmdOptions: false,
+  cmdOptions: null,
   globStart: 1,
   pipeOnly: false,
-  unix: true,
   wrapOutput: true,
-  overWrite: false,
+  unix: true,
```
freitagbr: Looking at the source, `unix` is only used by `exec`, because `exec` args differ from the other commands. I think now would be a good chance to improve this.

It makes sense to expose the ability to skip manipulating the args, so let's rename the option to make that clearer. Something like `preserveArgs`?

nfischer: I was planning to hide the `unix` option in a later PR. Renaming is ok too, but I don't see a need to expose it (we already offer ways to disable globbing and option parsing, which are the most expensive argument manipulations).

freitagbr: Ok, that's fine.

freitagbr:
LGTM

PR #807, opened 2017-11-16
refactor(exec): remove paramsFile

nfischer:
The `paramsFile` is obsolete now that we use `execFileSync()` for our
internal implementation. Instead, we pass parameters to the child
process directly as a single commandline parameter to reduce file I/O.

Issue #782


nfischer:
Reduces code 🎉  Decreases overall coverage because those lines were covered 😞

PR #809, opened 2018-01-06
adding wc unix command

jumson:
I added partial functionality of the Unix 'wc' command.
from the man page:
```
wc - Print newline, word, and byte counts for each FILE, 
and a total line if more than one FILE is specified.  A word 
is a non-zero-length sequence of characters delimited by 
white space.
```
and I implemented these options:
```
        -m, print the character counts
        -l, print the newline counts
        -w, print the word counts
```
I made a test file based on tail.js test, but I'm not sure how to use it.

I tested it alongside wc in bash and it gave similar results for single and multiple files.


jumson:
This just occurred to me...why create any new command implementations in javascript when the  shelljs/exec command exists?
regarding this pull request, does not:
```
shell.wc('file.txt')
```
and
```
shell.exec('wc file.txt')
```
accomplish the same thing?

jumson:
In response to myself -- the time to execute these commands on windows nodejs are below. There is a clear lag when using shelljs\exec function, compared to both the shelljs\wc function I just made, and simply using nodejs' child process, as well as the regule wc executable provided in indows and ubuntu (not WSL)

```

PS C:\Users\xx> Measure-Command {wc .\notes-uname.txt}
TotalSeconds      : 0.0386133
TotalMilliseconds : 38.6133

PS C:\Users\xx> Measure-Command {node .\child.js}
TotalSeconds      : 1.1918551
TotalMilliseconds : 1191.8551

PS C:\Users\xx> Measure-Command {node .\e.js}
TotalSeconds      : 1.2602646
TotalMilliseconds : 1260.2646

PS C:\Users\xx> Measure-Command {node .\wc.js}
TotalSeconds      : 0.2618337
TotalMilliseconds : 261.8337
```
and ubuntu:
```
root@xx# time wc notes-uname.txt
  39  131 1244 notes-uname.txt
real    0m0.020s
user    0m0.000s
sys     0m0.004s

root@xx# time node child.js
real    0m0.266s
user    0m0.116s
sys     0m0.020s

root@xx# time node e.js
real    0m1.216s
user    0m0.284s
sys     0m0.096s

root@xx# time node wc.js
real    0m1.020s
user    0m0.216s
sys     0m0.032s
```

nfischer:
> There is a clear lag when using shelljs\exec function

Among other reasons, `shell.exec()` needs to execute a system call (something like `execvp`), and that has associated overhead. JavaScript implementations don't need that particular syscall, which gives a slight advantage.

> why create any new command implementations in javascript

Another reason is to make cross-platform implementations 😄

.gitignore
```diff
@@ -19,3 +19,4 @@ coverage/
 # OS X
 .DS_Store
 
+file1.txt
```
nfischer: Don't change gitignore

commands.js
```diff
@@ -1,29 +1,30 @@
...
+    'tail',
+    'tempdir',
+    'test',
+    'to',
+    'toEnd',
+    'touch',
+    'uniq',
+    'wc',
+    'which',
+];
```
nfischer: Files should end in a newline

test/wc.js
```diff
@@ -0,0 +1,145 @@
...
+        'test/resources/head/file1.txt');
+    t.falsy(shell.error());
+    t.is(result.code, 0);
+    t.is(result.toString(),
+        bottomOfFile2
+        .slice(0, 4)
+        .reverse()
+        .concat(bottomOfFile1.slice(0, 4).reverse())
+        .join('\n') + '\n');
+});
```
nfischer: Files should end in a newline

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+
+        wc.push(temp);
+
+    });
+
+    if (moreThan1) wc.push(`${totalLines} ${totalWords} ${totalChars} total`);
+
+    return wc.join('\n');
+}
+module.exports = _wc;
```
nfischer: Files should end in a newline

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+//@ Read the end of a file.
+function _wc(options, files) {
+    var wc = [];
+    var pipe = common.readFromPipe();
+
+    if (!files && !pipe) common.error('no paths given');
+
+    var idx = 1;
+
+    /*  - from tail.js, which I'm using as a template. Not sure I need this section
```
nfischer: You can remove this section

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+//@ + `-w`: print the word counts (space delimited characters)
+//@ 
+//@ Examples:
+//@
+//@ ```javascript
+//@ var counts = wc('file.txt'); //returns charCount lineCount wordCount
+//@ var charCount = wc('-m','file.txt'); //returns number of characters
+//@ var lineCount = wc('-l','file.txt'); //returns number of newlines
+//@ var wordCount = wc('-w','file.txt'); //returns number of words 
+//@ var countFiles = wc('file1', 'file2'); // returns: lineCount wordCount charCount 'file1'\nlineCount wordCount charCount 'file2' \n totalLines totalWords totalChars 'total'
```
nfischer: Column limit. Maybe we can move this onto its own line? Or we could just remove this example.

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+//@ Available options:
+//@ + `-m`: print the char counts 
+//@ + `-l`: print the newline counts
+//@ + `-w`: print the word counts (space delimited characters)
+//@ 
+//@ Examples:
+//@
+//@ ```javascript
+//@ var counts = wc('file.txt'); //returns charCount lineCount wordCount
+//@ var charCount = wc('-m','file.txt'); //returns number of characters
```
nfischer: Put a space after the comment character, i.e.

```
// returns charCount...
```

instead of 

```
//returns charCount...
```

src/wc.js
```diff
@@ -0,0 +1,159 @@
+/*
```
nfischer: Remove this, since it describes unix `wc`, not shelljs `wc()`

README.md
```diff
@@ -656,6 +656,726 @@ uniq('-cd', 'foo.txt', 'bar.txt');
...
+Examples:
+
+```javascript
+var counts = wc('file.txt'); //returns charCount lineCount wordCount
+var charCount = wc('-m','file.txt'); //returns number of characters
+var lineCount = wc('-l','file.txt'); //returns number of newlines
+var wordCount = wc('-w','file.txt'); //returns number of words 
+var countFiles = wc('file1', 'file2'); // returns 'file1' charCount lineCount wordCount\n'file2' charCount lineCount wordCount\n'total' lineCount
+var countFiles = wc(['file1', 'file2']); //same as above
+```
```
nfischer: Whoa, what happened to the README? Did you run `npm run gendocs`? Maybe our script is broken...

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+function getLines(data) {
+    //data is the file contents
+    var lines = data.split('\n'); // array
+
+    if (lines[lines.length - 1] === '') {
+        //in case it wanted to count an empty line at the bottom
+        lines.pop();
+    }
+
+    return lines.length.toString();
```
nfischer: Return a number here instead (same thing for the other `get*()` functions), remove the `parseInt` calls below.

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function getWords(data) {
+    var regexSplit = /\s/;
+    var wordsInitial = data.split(regexSplit); //array
+
+    while (wordsInitial.indexOf('') !== -1) {
```
nfischer: It might be more elegant to write this as a `.filter()` expression:

```js
var wordsList = wordsInitial.filter(function (word) {
  if (word !== '') return word;
});
```

I haven't benchmarked, but the `.filter()` is probably more performant and memory efficient.

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+        //with words, cut out any blanks enumerated
+        wordsInitial.splice(wordsInitial.indexOf(''), 1);
+    }
+    return wordsInitial.length.toString();
+}
+
+function getChars(data) {
+    return data.length.toString();
```
nfischer: I think this is actually wrong (I'm looking at [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length)):

> uses a single 16-bit code unit to represent the most common characters, but needs to use two code units for less commonly-used characters, **so it's possible for the value returned by length to not match the actual number of characters in the string.**

Is there a way to guarantee that two-codepoint-long characters only count as a single character each?

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+        var thisWords = getWords(contents);
+        var thisChars = getChars(contents);
+        totalLines += parseInt(thisLines);
+        totalWords += parseInt(thisWords);
+        totalChars += parseInt(thisChars);
+
+        if (options.charCount || options.lineCount || options.wordCount) {
+            //modify wc accordingly
+
+            if (options.lineCount) temp += `${thisLines} `;
```
nfischer: Use regular strings (single quotes)

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        var contents = file === '-' ? pipe : fs.readFileSync(file, 'utf8');
+        /*
+        'm': 'charCount',
+        'l': 'lineCount',
+        'w': 'wordCount',
+        */
+
+        var temp = '';
```
nfischer: A better variable name might be `output` or `formattedInfo`

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+        }
+        temp += ` ${file}`
+
+        wc.push(temp);
+
+    });
+
+    if (moreThan1) wc.push(`${totalLines} ${totalWords} ${totalChars} total`);
+
+    return wc.join('\n');
```
nfischer: We probably want to end this with a trailing newline. I think most other commands follow this pattern.

src/wc.js
```diff
@@ -0,0 +1,159 @@
...
+
+    */
+
+    files = [].slice.call(arguments, idx);
+
+    if (pipe) {
+        files.unshift('-');
+    }
+
+    var moreThan1 = false;
```
nfischer: You can remove this variable, just inline `files.length > 1` down below.

jumson:
@nfischer thanks for the feedback,  i'll probably make adjustments this weekend.

src/wc.js
jumson: I went back and forth on this a lot. It's been a while (in grad school time) so my mind has been elsewhere. 
There is a complication in determining the actual encoding and how that is counted. I also wanted to implement the "bytes" option -- anyways, I'll relook this as I was assuming the .length property was sufficient.

I think i'll read the file into a buffer and work on the buffer (instead of reading it in as UTF-8, as it currently stand), possibly first determining it's encoding, then moving on from there. That might require another library, we shall see.

src/wc.js
jumson: i'll probably remove that example.

README.md
jumson: yeah..i'll fix it

src/wc.js
jumson: some more initial thoughts -- since i'm already reading this stream in as a utf-8, it should be in 'chunks' that made a 16-bit code unit into a character that .length property will count as one. Here is a demonstration, using the REPL:
```
> const buf5 = Buffer.from('tést');
undefined
> buf5
<Buffer 74 c3 a9 73 74>
> buf5.toString()
'tést'
> buf5.toString('utf-8')
'tést'
> buf5.toString('hex')
'74c3a97374'
> buf5.length
5
> buf5.toString().length
4
```

jumson: Reference: https://nodejs.org/api/buffer.html#buffer_buffer

jumson: also, based on this conversation, I think i'll implement the '--byte' option, which may be useful

nfischer:
@jumson no worries, and thanks for the pull request! It looks pretty good overall.

Also, I think `npm run lint` will flag some style issues.

src/wc.js
nfischer: I think you need to test with a different character. Try '🎉' instead. 'é' can be encoded in a single utf-16 code unit (2 bytes).

> I think i'll implement the '--byte' option, which may be useful

Yeah, that's fine, as long as we get the correct bytes of the file, not the bytes of the JS string.

Also, I'm not sure what the correct behavior is when the file is utf-16 encoded. Unix `wc` doesn't seem to give correct answers.

README.md
```diff
@@ -656,6 +662,29 @@ uniq('-cd', 'foo.txt', 'bar.txt');
...
+Available options:
++ `-m`: print the char counts 
++ `-l`: print the newline counts
++ `-w`: print the word counts (space delimited characters)
++ `-c`: print the byte counts 
+
+Examples:
+
+```javascript
+var counts = wc('file.txt'); //returns lineCount wordCount byteCount
```
nfischer: @jumson What I mean is to change "`//returns`" into "`// returns`"

nfischer:
@jumson it looks like you have test failures, can you check those?

test/wc.js
```diff
@@ -0,0 +1,146 @@
...
+
```
nfischer: Remove extra newline (just delete line 146)

src/wc.js
```diff
@@ -0,0 +1,169 @@
...
+        wc.push(formattedOutput);
+
+    });
+
+    if (files.length > 1) wc.push(`${totalLines} ${totalWords} ${totalBytes} total\n`);
+
+    return wc.join('\n');
+}
+
+module.exports = _wc;
```
nfischer: Still needs a newline at the end of line 169.

test/wc.js
nfischer: I opened #816 to catch this automatically. Once that PR lands, you can rebase off master and fix this with `npm run lint -- --fix`.

nfischer:
@jumson are you still working on this?

I think this is a better candidate for a [shelljs plugin](https://github.com/shelljs/shelljs/wiki/Using-ShellJS-Plugins) than a core method. Would you be interested in rewriting this in a separate repo as a plugin?

nfischer:
Closing due to lack of response (also, I think this is better implemented as a 3rd party plugin).

PR #816, opened 2018-01-18
chore(lint): don't allow excess trailing newlines

nfischer:
This enforces that files should not end with blank lines (i.e. more than
one trailing newline). The `eol-last` rule (inherited from airbnb's
config) already enforces that files end in at least one trailing
newline, so adding this rule enforces that it ends in exactly 1.

See https://eslint.org/docs/rules/no-multiple-empty-lines and
https://eslint.org/docs/rules/eol-last for more information.

Inspired by #809.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,7 +23,7 @@
     "consistent-return": "off",
     "no-mixed-operators": "off",
     "no-prototype-builtins": "off",
-    "no-multiple-empty-lines": ["error", { "max": 2, "maxEOF": 0 } ],
+    "no-multiple-empty-lines": ["error", { "max": 2, "maxBOF": 0, "maxEOF": 0 } ],
     "new-cap": ["error", {
       "capIsNewExceptions": [
         "ShellString"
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -23,6 +23,7 @@
     "consistent-return": "off",
     "no-mixed-operators": "off",
     "no-prototype-builtins": "off",
+    "no-multiple-empty-lines": ["error", { "max": 2, "maxEOF": 0 } ],
```
nfischer: @freitagbr We can also make `"max": 1`, but that involves changing:

https://github.com/shelljs/shelljs/blob/902e49c0f81834c4da61ace740cc285417b3c5d3/src/tempdir.js#L25-L29

and

https://github.com/shelljs/shelljs/blob/902e49c0f81834c4da61ace740cc285417b3c5d3/src/test.js#L17-L21

freitagbr: 2 lines max is fine. `"maxBOF": 0` should be added.

nfischer: Done

nfischer:
PTAL

freitagbr:
LGTM

PR #819, opened 2018-01-19
Revert "refactor(exec): remove paramsFile (#807)"

### LINKED ISSUE #818: Exec failing with internal error when piping large output
8

0.8.0

### Operating system:
Ubuntu 16.04.3

### Description of the bug:
Piping medium to large amounts of output between two exec's causes the second one to fail with 'Error: exec: internal error'.

```javascript
var shell = require('shelljs')
shell.config.fatal = true
shell.exec('cat small.txt', { silent: true }).exec('wc -l', { silent: true }).to('few-lines.txt')
shell.exec('cat big.txt', { silent: true }).exec('wc -l', { silent: true }).to('many-lines.txt')
```
Here, `small.txt` is a small file (under 1kb), while `big.txt` is a 2MB file.
The first command executes perfectly well, while the second one fails with `exec: internall error`.
Without `fatal = true`, the second exec on the second command simply returns `null` and the command fails due to `to` not being a property of null. (Likely related: https://github.com/shelljs/shelljs/issues/724)

### Note

I did some digging and it seems the problem is here:

https://github.com/shelljs/shelljs/blob/master/src/exec.js#L68

That call to `execFileSync` receives `execArgs`, which contains the serialized output from the first exec. So it's essentially calling a process with an argument which is over 2MB, which is breaking a limit on my OS. That call to `execFileSync` ends with an `Error: spawnSync /usr/bin/node E2BIG` being thrown by node.

nfischer:
This reverts commit 64d5899abc86dd7b7fa84455c0ce3551786c4b5b.

Reason for revert: If stdin is large, then the param object can become
an extremely long string, exceeding the maximum OS size limit on
commandline parameters.

Original change's description:
> refactor(exec): remove paramsFile (#807)
>
> The `paramsFile` is obsolete now that we use `execFileSync()` for our
> internal implementation. Instead, we pass parameters to the child
> process directly as a single commandline parameter to reduce file I/O.
>
> Issue #782

Fixes #818

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -21,6 +21,7 @@ function execSync(cmd, opts, pipe) {
   }
 
   var tempDir = _tempDir();
+  var paramsFile = path.resolve(tempDir + '/' + common.randomFileName());
   var stderrFile = path.resolve(tempDir + '/' + common.randomFileName());
   var stdoutFile = path.resolve(tempDir + '/' + common.randomFileName());
 
@@ -32,6 +33,7 @@ function execSync(cmd, opts, pipe) {
     encoding: 'utf8',
   }, opts);
 
+  if (fs.existsSync(paramsFile)) common.unlinkSync(paramsFile);
   if (fs.existsSync(stderrFile)) common.unlinkSync(stderrFile);
   if (fs.existsSync(stdoutFile)) common.unlinkSync(stdoutFile);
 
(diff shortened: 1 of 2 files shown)


nfischer:
Tested by creating a 6M lorem ipsum file:

```js
shell.cat('lorem.txt').exec('tr o a').to('out.txt');
// crashes without this revert
```

src/exec.js
```diff
@@ -21,6 +21,7 @@ function execSync(cmd, opts, pipe) {
   }
 
   var tempDir = _tempDir();
+  var paramsFile = path.resolve(tempDir + '/' + common.randomFileName());
```
freitagbr: This doesn't need to be fixed now, but I think the paths should be joined with `path.join`, instead of just concatenating the strings... future improvements.

nfischer:
Travis mac builders appear to be stuck. I'm just going to land this, because I'm pretty confident it's safe.

PR #824, opened 2018-02-08
Update README.md

Zearin:
Minor formatting tweaks

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,8 +137,8 @@ For less-commonly used commands and features, please check out our [wiki
 page](https://github.com/shelljs/shelljs/wiki).
 
 
-### cat([options,] file [, file ...])
-### cat([options,] file_array)
+### `cat([options,] file [, file ...])`
+### `cat([options,] file_array)`
 Available options:
 
 + `-n`: number all output lines
@@ -156,13 +156,13 @@ containing the files if more than one file is given (a new line character is
 introduced between each file).
 
 
-### cd([dir])
+### `cd([dir])`
 Changes to directory `dir` for the duration of the script. Changes to home
 directory if no argument is supplied.
 
 
-### chmod([options,] octal_mode || octal_string, file)
-### chmod([options,] symbolic_mode, file)
+### `chmod([options,] octal_mode || octal_string, file)`
+### `chmod([options,] symbolic_mode, file)`
 
 Available options:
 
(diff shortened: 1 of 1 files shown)


nfischer:
@Zearin Docs changes need to be done in `src/*.js` files first, then update the README using `npm run gendocs`

README.md
```diff
@@ -640,11 +640,11 @@ touch({ '-r': FILE }, '/path/to/some/dir/source.js');
 ```
 
 Update the access and modification times of each FILE to the current time.
-A FILE argument that does not exist is created empty, unless -c is supplied.
-This is a partial implementation of *[touch(1)](http://linux.die.net/man/1/touch)*.
+A FILE argument that does not exist is created empty, unless `-c` is supplied.
+This is a partial implementation of *[`touch(1)`](http://linux.die.net/man/1/touch)*.
```
nfischer: If we're using backtick formatting, let's remove the italics

README.md
```diff
@@ -273,10 +273,10 @@ Arguments:
 
 Display the list of currently remembered directories. Returns an array of paths in the stack, or a single path if +N or -N was specified.
 
-See also: pushd, popd
+See also: `pushd`, `popd`
 
 
-### echo([options,] string [, string ...])
+### `echo([options,] string [, string ...])`
```
nfischer: I think it looks a bit better with the headers as-is. Were there any formatting bugs I'm overlooking?

README.md
```diff
@@ -400,11 +400,11 @@ ln('file', 'newlink');
 ln('-sf', 'file', 'existing');
 ```
 
-Links source to dest. Use -f to force the link, should dest already exist.
+Links source to dest. Use `-f` to force the link, should `dest` already exist.
```
nfischer: Also wrap `source` and `dest` in backticks

Zearin:
Whoops!  That’s what I get for not looking more carefully.  I’ll update the docs in src as you said.

README.md
Zearin: No.  I just prefer the monospace font for any documentation that is referencing a command/option/option-value.  

But I can omit this when I redo the Pull Request.  :)

Zearin:
Reopening as a new Pull Request…

nfischer:
You can always just add more commits to the same branch 😄

PR #826, opened 2018-02-12
chore: update coffescript

jluria:
* update coffeescript to 2.2.1 from 1.10.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "ava": "^0.21.0",
     "chalk": "^1.1.3",
     "codecov": "^1.0.1",
-    "coffee-script": "^1.10.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
   "dependencies": {
     "glob": "^7.0.0",
     "interpret": "^1.0.0",
-    "rechoir": "^0.6.2"
+    "rechoir": "^0.6.0"
```
nfischer: Why did rechoir change?

nfischer:
@jluria have you had a chance to look at this again?

nfischer:
I don't think we should update until we drop support for node v4 and v5. Thanks for the contribution, we can revisit this in the future!

Issue #795, opened 2017-10-27
Add test for ls() on a symlink to a directory


nfischer:
This is one of our missing areas of coverage: https://codecov.io/gh/shelljs/shelljs/src/a187bd1b36ce28a5af214607257506ee28e1beb6/src/ls.js#L80

This shouldn't be too tricky to test, but I'm not sure if extra precautions need to be taken for Windows.

dwi2:
Hello.

I tried to come up a patch for this issue but I found out something interesting. In short, I believe these issue is already covered by https://github.com/shelljs/shelljs/blob/master/test/ls.js#L333. The report of codecov might be false negative.

I did the following things:
1. It's kind of silly but I injected console.log in lines between https://github.com/shelljs/shelljs/blob/master/src/ls.js#L82 and https://github.com/shelljs/shelljs/blob/master/src/ls.js#L86, then ran unit test locally. I did see these console.log got printed.


2. Ran test locally via `npm run test` and the coverage report is a little bit different from https://codecov.io/gh/shelljs/shelljs/src/master/src/ls.js .
My local coverage report looks like this: (on Mac OSX 10.11.6 with node v8.10.0)
```
----------------|----------|----------|----------|----------|-------------------|
File            |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------------|----------|----------|----------|----------|-------------------|
All files       |    95.81 |    89.21 |    99.24 |     96.6 |                   |
 shelljs        |      100 |      100 |      100 |      100 |                   |
  commands.js   |      100 |      100 |      100 |      100 |                   |
  global.js     |      100 |      100 |      100 |      100 |                   |
  plugin.js     |      100 |      100 |      100 |      100 |                   |
  shell.js      |      100 |      100 |      100 |      100 |                   |
 shelljs/src    |    95.74 |    89.21 |    99.22 |    96.54 |                   |
  cat.js        |      100 |      100 |      100 |      100 |                   |
  cd.js         |      100 |     87.5 |      100 |      100 |                35 |
  chmod.js      |    89.41 |    90.12 |      100 |    89.41 |... 81,183,187,192 |
  common.js     |    97.94 |    94.07 |    96.55 |     98.4 |       260,327,358 |
  cp.js         |    89.84 |    87.76 |      100 |    91.13 |... 76,177,178,179 |
  dirs.js       |    97.18 |    93.88 |      100 |    97.18 |             27,29 |
[...]

PR #837, opened 2018-04-19
Verify code coverage of ls() on Travis

dwi2:
### DON'T MERGE

This is to verify if #795 needs fix or not.
As described [here](https://github.com/shelljs/shelljs/issues/795#issuecomment-382564583), let's see if there are `console.log()` messages get printed out on Travis.

If our assumption is correct, we should see
```
[795] Follow symlink
[795] Follow existed symlink of directory
[795] Follow symlink
```
on travis log.


dwi2:
This is only for verification so close it since we got the logs we need.

PR #841, opened 2018-05-03
Add '-p' (preserve) option to 'cp', fixes #771

### LINKED ISSUE #771: Preserve timestamps on cp?
Basically, the title already says it all: Is it possible to preserve the timestamps of files and directories when using the `cp` command?

dwi2:
This fixes #771. 

### WIP
This is still work in progress. But I want to make sure I am on the right track so I just create pull request first.

I don't plan to implement the full `--preserve=ATTR_LIST` feature in this PR. I think that could be a follow-up work.

### TODO
- [ ] Add doc and run `npm run gendocs`
- [ ] Add test for `cp -p` on directory 
- [x] Figure out why tests fail on Travis CI

### DIFF
--- a/src/cp.js
+++ b/src/cp.js
@@ -11,6 +11,7 @@ common.register('cp', _cp, {
     'r': 'recursive',
     'L': 'followsymlink',
     'P': 'noFollowsymlink',
+    'p': 'preserve',
   },
   wrapOutput: false,
 });
@@ -51,6 +52,7 @@ function copyFileSync(srcFile, destFile, options) {
     var pos = 0;
     var fdr = null;
     var fdw = null;
+    var srcStat = common.statFollowLinks(srcFile);
 
     try {
       fdr = fs.openSync(srcFile, 'r');
@@ -75,7 +77,11 @@ function copyFileSync(srcFile, destFile, options) {
     fs.closeSync(fdr);
     fs.closeSync(fdw);
 
-    fs.chmodSync(destFile, common.statFollowLinks(srcFile).mode);
+    if (options.preserve) {
+      fs.chownSync(destFile, srcStat.uid, srcStat.gid);
+      fs.utimesSync(destFile, srcStat.atime, srcStat.mtime);
+    }
+    fs.chmodSync(destFile, srcStat.mode);
   }
 }
 
--- a/src/cp.js
+++ b/src/cp.js
@@ -74,13 +74,13 @@ function copyFileSync(srcFile, destFile, options) {
       pos += bytesRead;
     }
 
+    if (options.preserve) {
+      fs.fchownSync(fdw, srcStat.uid, srcStat.gid);
+      fs.futimesSync(fdw, srcStat.atime, srcStat.mtime);
+    }
+
     fs.closeSync(fdr);
     fs.closeSync(fdw);
-
-    if (options.preserve) {
-      fs.chownSync(destFile, srcStat.uid, srcStat.gid);
-      fs.utimesSync(destFile, srcStat.atime, srcStat.mtime);
-    }
   }
 }
 
--- a/src/cp.js
+++ b/src/cp.js
@@ -62,7 +62,7 @@ function copyFileSync(srcFile, destFile, options) {
     }
 
     try {
-      fdw = fs.openSync(destFile, 'w');
+      fdw = fs.openSync(destFile, 'w', srcStat.mode);
     } catch (e) {
       /* istanbul ignore next */
       common.error('copyFileSync: could not write to dest file (code=' + e.code + '):' + destFile);
@@ -81,7 +81,6 @@ function copyFileSync(srcFile, destFile, options) {
       fs.chownSync(destFile, srcStat.uid, srcStat.gid);
       fs.utimesSync(destFile, srcStat.atime, srcStat.mtime);
     }
-    fs.chmodSync(destFile, srcStat.mode);
   }
 }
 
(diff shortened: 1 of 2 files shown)


nfischer:
I kicked off a new appveyor build because the failure doesn't look related to your PR. Hopefully that's just a fluke.

src/cp.js
```diff
@@ -75,7 +77,11 @@ function copyFileSync(srcFile, destFile, options) {
     fs.closeSync(fdr);
     fs.closeSync(fdw);
 
-    fs.chmodSync(destFile, common.statFollowLinks(srcFile).mode);
+    if (options.preserve) {
+      fs.chownSync(destFile, srcStat.uid, srcStat.gid);
+      fs.utimesSync(destFile, srcStat.atime, srcStat.mtime);
+    }
+    fs.chmodSync(destFile, srcStat.mode);
```
freitagbr: Instead of changing the mode after writing the file, we can set the mode on the initial call to `fs.openSync` on line 65:
```
fdw = fs.openSync(destFile, 'w', srcStat.mode);
```

test/cp.js
```diff
@@ -756,3 +756,29 @@ test('should not overwrite recently created files (not give error no-force mode)
...
 });
+
+test('cp -p should preserve mode, ownership, and timestamp', t => {
+  const result = shell.cp('-p', 'test/resources/file1', `${t.context.tmp}/preservedFile1`);
+  const stat = common.statFollowLinks('test/resources/file1');
+  const statOfResult = common.statFollowLinks(`${t.context.tmp}/preservedFile1`);
+  t.is(result.code, 0);
+
+  t.is(stat.mtime.getTime(), statOfResult.mtime.getTime());
+  t.is(stat.atime.getTime(), statOfResult.atime.getTime());
```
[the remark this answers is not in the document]
nfischer: Probably a race. Try adding a sleep to the test (see `touch` tests for example), which will hopefully provide a reliable repro.

src/cp.js
```diff
@@ -72,10 +74,13 @@ function copyFileSync(srcFile, destFile, options) {
       pos += bytesRead;
     }
 
+    if (options.preserve) {
+      fs.fchownSync(fdw, srcStat.uid, srcStat.gid);
+      fs.futimesSync(fdw, srcStat.atime, srcStat.mtime);
```
dwi2: On linux, `fs.utimesSync()` always produces timestamps rounded off to `seconds` but original files could have  timestamps of millisecond level. But I don't why is that so I probably need more time to figure it out.
However `fs.futimesSync()` could produce timestamps of millisecond level. So I put it this way for now.

Also I tried to add sleep in test but it didn't work.

dwi2:
Also about the [appveyor build failed](https://ci.appveyor.com/project/shelljs/shelljs/build/1610/job/dgdr1ws2ote8fu9v), I checked the difference between the build result between this PR and [master branch](https://ci.appveyor.com/project/shelljs/shelljs/build/1605).

In the [build of this PR]((https://ci.appveyor.com/project/shelljs/shelljs/build/1610/job/dgdr1ws2ote8fu9v), appveyor tried to [install npm v6.0](https://ci.appveyor.com/project/shelljs/shelljs/build/1610/job/dgdr1ws2ote8fu9v#L258) on node JS v5 while it [installed npm v5.8](https://ci.appveyor.com/project/shelljs/shelljs/build/1605/job/l04fo91l3mome979#L147) in master build.
I couldn't find the support version of npm v6 but I suspect this is the reason why build fails. I think I also need to update [appveyor.xml](https://github.com/shelljs/shelljs/blob/master/appveyor.yml#L15) in this PR too.

nfischer:
> I think I also need to update appveyor.xml in this PR too.

Not your responsibility. I filed #844, I'll handle this ASAP.

> Also I tried to add sleep in test but it didn't work.

Sleeping is supposed to *break* flaky tests. The suspicion is that there's a race condition, so you should add the sleep for all platforms (especially those which appear to pass). I didn't look at test output though, but examining that should make it obvious if this was because of a race.

nfischer:
> > I think I also need to update appveyor.xml in this PR too.
> 
> Not your responsibility. I filed #844, I'll handle this ASAP.

@dwi2 FYI I fixed #844. Rebase at your convenience.

dwi2:
@nfischer Thanks a lot.

src/cp.js
nfischer: Huh, ok. This implementation is fine. I will add a comment explaining that `utimesSync` doesn't work but `futimesSync` does.

test/cp.js
```diff
@@ -756,3 +756,29 @@ test('should not overwrite recently created files (not give error no-force mode)
   // Ensure First file is copied
   t.is(shell.cat(`${t.context.tmp}/file1`).toString(), 'test1');
 });
+
+test('cp -p should preserve mode, ownership, and timestamp', t => {
+  const result = shell.cp('-p', 'test/resources/file1', `${t.context.tmp}/preservedFile1`);
```
nfischer: I'll follow-up to this to `touch()` the original file. Otherwise, we might have a race where the test setup creates `file1` and the test nearly-immediately copies it--depending on the timing/precision, a regular `cp()` call might even pass this test.

We should probably also chmod the file too, to double-check the mode gets copied correctly (and it isn't the default mode).

nfischer:
Close in favor of #869. Thanks for your help!

dwi2:
Thanks for carrying on.

PR #847, opened 2018-05-09
chore(appveyor): do not use latest npm

### LINKED ISSUE #844: Appveyor installs the latest npm, instead of the latest compatible npm
Discovered in #841, [this comment](https://github.com/shelljs/shelljs/pull/841#issuecomment-387068159).

Basically, fix [this line](https://github.com/shelljs/shelljs/blob/master/appveyor.yml#L15). We shouldn't install the latest npm all the time, because it may not be supported on older node versions.

High priority, because this will fail the next master build.

nfischer:
This changes appveyor from using the latest npm version to only using
the preinstalled npm version. This is safer, because it's guaranteed to
be a compatible version.

Fixes #844

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,6 @@ version: '{build}'
 # Install scripts. (runs after repo cloning)
 install:
   - ps: Install-Product node $env:nodejs_version
-  - npm -g install npm@latest
   - set PATH=%APPDATA%\npm;%PATH%
   - node --version
   - npm --version


nfischer:
I'm landing this, because we need to fix master ASAP. However, there's a caveat to this change: [npm recommends npm@3 on Windows](https://github.com/npm/npm#windows-install-or-upgrade), however this change bumps the node4 bot down to npm@2.15.11.

It's not clear if this causes correctness issues. This may merit further investigation.

PR #848, opened 2018-05-10
Prevent require-ing bin/shjs

### LINKED ISSUE #789: Prevent require()ing executable-only files
This was a point raised in #786. We can prevent `require()`ing files with:

```js
if (require.main === module) {
    // everything goes here
}
```

This might be a good idea for files which we only intend to execute (e.g. `bin/shjs`).

I would also consider rewriting as:

```js
if (require.main !== module) {
  throw new Error('Module should not be required');
}
// everything goes here
```

freitagbr:
Fixes #789 

This makes it so that `bin/shjs` cannot be `require`d as a module, and adds a test for this.

### DIFF
--- a/bin/shjs
+++ b/bin/shjs
@@ -1,19 +1,20 @@
 #!/usr/bin/env node
 
-function exit(msg) {
+if (require.main !== module) {
+  throw new Error('Executable-only module should not be required');
+}
+
+function exitWithErrorMessage(msg) {
   console.log(msg);
   console.log();
   process.exit(1);
 }
 
-if (require.main !== module) {
-  throw new Error('Executable-only module should not be required');
-}
-
 if (process.argv.length < 3) {
-  exit('ShellJS: missing argument (script name)');
+  exitWithErrorMessage('ShellJS: missing argument (script name)');
 }
 
+// we only need ShellJS methods past this point, so require here
 require('../global');
 
 var args,
@@ -28,7 +29,7 @@ if (!scriptName.match(/\.js/) && !scriptName.match(/\.coffee/)) {
 }
 
 if (!test('-f', scriptName)) {
-  exit('ShellJS: script not found ('+scriptName+')');
+  exitWithErrorMessage('ShellJS: script not found ('+scriptName+')');
 }
 
 args = process.argv.slice(3);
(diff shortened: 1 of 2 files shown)


bin/shjs
```diff
@@ -1,12 +1,21 @@
...
 
-if (process.argv.length < 3) {
-  console.log('ShellJS: missing argument (script name)');
+function exit(msg) {
+  console.log(msg);
   console.log();
   process.exit(1);
 }
 
+if (require.main !== module) {
```
nfischer: nit: can you move this up? Also, I assume you moved `require('../global')` down so that it doesn't pollute the test--this is OK, but you need to add a comment explaining why.

bin/shjs
```diff
@@ -1,12 +1,21 @@
 #!/usr/bin/env node
-require('../global');
 
-if (process.argv.length < 3) {
-  console.log('ShellJS: missing argument (script name)');
+function exit(msg) {
```
nfischer: nit: rename this function (it's too easy to confuse this with `shell.exit()`). Suggestion: `exitWithErrorMessage()`

freitagbr:
@nfischer PTAL

bin/shjs
```diff
@@ -1,12 +1,22 @@
...
+  console.log(msg);
   console.log();
   process.exit(1);
 }
 
+if (process.argv.length < 3) {
+  exitWithErrorMessage('ShellJS: missing argument (script name)');
+}
+
+// we only need ShellJS methods past this point, so require here
```
nfischer: Our usual style is to import at the top, not to import immediately before use. If there's a real requirement that we import down here, please document that requirement.

I suspect the requirement is: "we must import this after the `require.main` check, otherwise we pollute the global namespace as a side effect of the `disallow require-ing` test case."

If that's the real requirement, then you should move this immediately below the `require.main` check (and indicate this requirement). Otherwise, please explain why this is the appropriate spot, because I can't figure that out.

freitagbr: Correct, the `require` at least needs to be after the `require.main` check, because the error it throws is catchable, and we don't want any side effects, as you said. I also moved it after the `process.argv` check, because we still don't need any globals at that point, but the fail condition is exiting. There is less of a benefit to `require`-ing after this check, so I will move it up.

PR #849, opened 2018-05-10
Add tests for rm removing symlinks

### LINKED ISSUE #796: Improve test coverage for rm
Looking at [coverage for rm](https://codecov.io/gh/shelljs/shelljs/src/a187bd1b36ce28a5af214607257506ee28e1beb6/src/rm.js) it looks like we're missing quite a bit of coverage surrounding symlinks to directories (see also #795). Let's see if we can improve that.

We should also look at lines for which we should ignore coverage (do we need to do this for FIFOs, or can we have a unix-only test?).

freitagbr:
Fixes #796 

Adds a test for removing a symbolic link to a file, and for failing to remove the contents of a directory through a symbolic link without the `-r` flag. Improves test coverage from 95.71% to 98.57%.

### DIFF
--- a/test/rm.js
+++ b/test/rm.js
@@ -57,6 +57,15 @@ test('invalid option', t => {
   t.is(result.stderr, 'rm: option not recognized: @');
 });
 
+test('remove symbolic link to a dir without -r fails', t => {
+  const result = shell.rm(`${t.context.tmp}/rm/link_to_a_dir/`);
+  t.truthy(shell.error());
+  t.is(result.code, 1);
+  t.is(result.stderr, 'rm: path is a directory');
+  t.truthy(fs.existsSync(`${t.context.tmp}/rm/link_to_a_dir`));
+  t.truthy(fs.existsSync(`${t.context.tmp}/rm/a_dir`));
+});
+
 //
 // Valids
 //
@@ -306,5 +315,16 @@ test('remove fifo', t => {
     const result = shell.rm(fifo);
     t.falsy(shell.error());
     t.is(result.code, 0);
+  });
+});
+
+test('remove symbolic link', t => {
+  utils.skipOnWin(t, () => {
+    t.truthy(shell.test('-L', `${t.context.tmp}/link`));
+    const result = shell.rm(`${t.context.tmp}/link`);
+    t.falsy(shell.error());
+    t.is(result.code, 0);
+    t.falsy(shell.test('-L', `${t.context.tmp}/link`));
+    t.falsy(fs.existsSync(`${t.context.tmp}/link`));
   });
 });


test/rm.js
```diff
@@ -57,6 +57,17 @@ test('invalid option', t => {
   t.is(result.stderr, 'rm: option not recognized: @');
 });
 
+test('remove symbolic link to a dir without -r fails', t => {
```
nfischer: This statement is only correct because of the trailing slash:

```sh
$ rm link_to_a_dir/ # fails
$ rm link_to_a_dir # succeeds (it removes the link itself, does not affect dir or contents
```

Please clarify this in the test name, and please include a more verbose comment explaining the significance of the trailing slash.

---

Also, we have an existing test which handles the case with no trailing slash and no `-r` (yay), however that test case needs an assertion. Currently, it asserts that the destination dir still exists, but it should instead assert that the *contents* of that destination dir still exist. The destination **always** survives an `rm`. If you can, please fix that test too.

test/rm.js
```diff
@@ -308,3 +319,14 @@ test('remove fifo', t => {
     t.is(result.code, 0);
   });
 });
+
+test('remove symbolic link', t => {
```
nfischer: nit: `remove symbolic link to file`

Does this actually require `skipOnWin`? I would be surprised if this is so, because `remove symbolic link to a dir` passes without that.

Also, it might be nice to move this up closer to that test case.

test/rm.js
nfischer: Other ways we might test link-handling:

 * `-r`, link to a dir, trailing slash (delete contents and the link)
 * `-rf`, link to a dir, trailing slash (this is already implemented, but I mention it here for completeness)
 * `-r`, link to a dir, no trailing slash (deletes only the link, not the destination, nor the destination's contents)
 * `-rf`, link to a dir, no trailing slash (same as above, probably not necessary to copy this out)

freitagbr: I've tested the first one and I think the behavior is a little different (linux 4.4.0, bash 4.3.48):

- `-r`, link to a dir, trailing slash: `rm: cannot remove 'dir/': Not a directory`

Can you tell me if you get the same behavior in bash?

nfischer:
@freitagbr what state is this in?

test/rm.js
nfischer: Yes, I see the same behavior on my machine.

nfischer: Also, `rm -rf linkToDir/` will exit without error, but will remove neither the link nor the target.

nfischer:
@freitagbr 2 tests fail on Windows, it looks like we should skip them.

freitagbr:
My bad, I forgot to add `skipOnWin` to the new tests.

test/rm.js
```diff
@@ -57,16 +60,23 @@ test('invalid option', t => {
...
-    t.truthy(fs.existsSync(`${t.context.tmp}/rm/link_to_a_dir`));
-    t.truthy(fs.existsSync(`${t.context.tmp}/rm/a_dir`));
-  });
-});
+test(
+  'cannot remove a symbolic link to a directory with trailing slash without -r flag',
+  t => {
+    utils.skipOnWin(t, () => {
+      // the trailing slash signifies that we want to delete the source
+      // directory and its contents, which can only be done with the -r flag
```
nfischer: > to delete the source directory and its contents

Can you rephrase this as "delete the contents of the source directory, without removing the source directory itself or the link"?

test/rm.js
```diff
@@ -259,27 +280,75 @@ test(
...
+  t.falsy(fs.existsSync(`${t.context.tmp}/link`));
+  t.truthy(fs.existsSync(`${t.context.tmp}/file1`));
+});
+
 test('remove symbolic link to a dir', t => {
+  t.truthy(shell.test('-L', `${t.context.tmp}/rm/link_to_a_dir`));
   const result = shell.rm(`${t.context.tmp}/rm/link_to_a_dir`);
   t.falsy(shell.error());
   t.is(result.code, 0);
+  t.falsy(shell.test('-L', `${t.context.tmp}/rm/link_to_a_dir`));
```
nfischer: I think this is already covered by the next line? Or, does `existsSync` follow links? If there's a reason, can you add a comment?

test/rm.js
```diff
@@ -259,27 +280,75 @@ test(
...
     // Both the link and original dir should remain, but contents are deleted
     t.truthy(fs.existsSync(`${t.context.tmp}/rm/link_to_a_dir`));
     t.truthy(fs.existsSync(`${t.context.tmp}/rm/a_dir`));
     t.falsy(fs.existsSync(`${t.context.tmp}/rm/a_dir/a_file`));
   });
 });
 
+test('rm -r, symlink to a dir, no trailing slash', t => {
+  utils.skipOnWin(t, () => {
+    const result = shell.rm('-rf', `${t.context.tmp}/rm/link_to_a_dir`);
```
nfischer: Should be `-r` not `-rf`

src/rm.js
```diff
@@ -45,7 +45,12 @@ function rmdirSyncRecursive(dir, force, fromSymlink) {
 
   // if was directory was referenced through a symbolic link,
   // the contents should be removed, but not the directory itself
-  if (fromSymlink) return;
+  if (fromSymlink) {
```
nfischer: As I understand, the issue is that `link_to_a_dir` refers to something, but `link_to_a_dir/` does not (because only directories may be optionally referred to with a trailing slash). But in both cases, `link_to_a_dir` references a directory, and so I would expect `fromSymlink` to be in `true` in both cases. 

It sounds like we should rename the variable, or we should add a comment explaining what it actually represents.

PR #850, opened 2018-05-10
chore: output npm version in travis

nfischer:
Knowing the `npm` version was a huge help to debugging issue #844. For
some reason, we output this on appveyor but not on Travis. This PR fixes
that.


nfischer:
I think it's because this branch isn't rebased off the fix.

PR #854, opened 2018-06-03
refactor(mkdir): consistent error handling

### LINKED ISSUE #722: Refactor mkdir
It's probably a good idea to refactor `mkdir()` and make sure we accommodate all the error codes listed in [the man page](http://man7.org/linux/man-pages/man2/mkdir.2.html).

I think a much better approach (after checking for the `-p` flag and handling that as appropriate) would be to call `fs.mkdir()` and catch any exception. Then we can have an if-else ladder handling each case as appropriate. This is more in-line with how GNU mkdir behaves.

This has the added bonus that we'll miss far fewer edge cases, since we're explicitly accounting for each type of exception.

Related to #721

nfischer:
This refactors mkdir() to (for the most part) invoke mkdirSync() and
catch any exceptions, instead of the error-prone approach of trying to
predict when specific errors would happen.

This adds two new tests for "name too long" and that it's OK to `mkdir
-p` a directory which already exists. The latter was implicitly covered
by another test case, but this should be explicit.

Lastly, this changes error messages to be more consistent with Linux
mkdir, and adds the "avoid-escape" rule in the test directory to allow
using double-quoted strings when it aids readability.

Fixes #722

### DIFF
--- a/src/mkdir.js
+++ b/src/mkdir.js
@@ -51,29 +51,13 @@ function mkdirSyncRecursive(dir) {
 function _mkdir(options, dirs) {
   if (!dirs) common.error('no paths given');
 
-  if (typeof dirs === 'string') {
-    dirs = [].slice.call(arguments, 1);
-  }
-  // if it's array leave it as it is
+  dirs = [].slice.call(arguments, 1);
 
   dirs.forEach(function (dir) {
-    try {
-      var stat = common.statNoFollowLinks(dir);
-      if (!options.fullpath) {
-        common.error('path already exists: ' + dir, { continue: true });
-      } else if (stat.isFile()) {
-        common.error('cannot create directory ' + dir + ': File exists', { continue: true });
-      }
-      return; // skip dir
-    } catch (e) {
-      // do nothing
-    }
-
-    // Base dir does not exist, and no -p option given
-    var baseDir = path.dirname(dir);
-    if (!fs.existsSync(baseDir) && !options.fullpath) {
-      common.error('no such file or directory: ' + baseDir, { continue: true });
-      return; // skip dir
+    // Skip mkdir if -p option is given and directory already exists
+    if (fs.existsSync(dir) && common.statNoFollowLinks(dir).isDirectory() &&
+      options.fullpath) {
+      return;
     }
 
     try {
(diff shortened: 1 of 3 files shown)


nfischer:
@freitagbr take a look, but this should probably be considered as a breaking change because it changes error messages.

src/mkdir.js
```diff
@@ -51,29 +51,13 @@ function mkdirSyncRecursive(dir) {
 function _mkdir(options, dirs) {
   if (!dirs) common.error('no paths given');
 
-  if (typeof dirs === 'string') {
```
nfischer: I think the type is always string. We introduced some code to flatten out arrays. Even in the case of ShellStrings, I think we also convert those into bare strings.

src/mkdir.js
```diff
@@ -83,16 +67,23 @@ function _mkdir(options, dirs) {
         fs.mkdirSync(dir, parseInt('0777', 8));
       }
     } catch (e) {
-      var reason;
-      if (e.code === 'EACCES') {
-        reason = 'Permission denied';
-      } else if (e.code === 'ENOTDIR' || e.code === 'ENOENT') {
```
nfischer: I think `ENOENT` should be handled distinctly from `ENOTDIR`, not sure why they're in the same branch here.

src/mkdir.js
```diff
@@ -83,16 +67,23 @@ function _mkdir(options, dirs) {
...
-      }
-      common.error('cannot create directory ' + dir + ': ' + reason, { continue: true });
+      // See mkdir(2) for details on each error case
+      // http://man7.org/linux/man-pages/man2/mkdir.2.html
+      var codeToMessage = {
+        EACCES: 'Permission denied',
+        EEXIST: 'File exists',
+        ENAMETOOLONG: 'File name too long',
+        ENOENT: 'No such file or directory',
+        ENOTDIR: 'Not a directory',
```
nfischer: I limited this to the list of error codes I could repro locally. I think that's fine for now, and we can add more later.

test/.eslintrc.json
```diff
@@ -11,6 +11,7 @@
     "no-param-reassign": "off",
     "no-console": "off",
     "curly": "off",
+    "quotes": ["error", "single", "avoid-escape"],
```
nfischer: Let me know if you'd like to promote this to the top-level config.

test/mkdir.js
```diff
@@ -145,6 +159,21 @@ test('-p flag', t => {
   shell.rm('-Rf', `${t.context.tmp}/a`); // revert
 });
 
+test('create same dir twice with -p flag', t => {
```
nfischer: This is implicitly covered in the "globbed dir" test, but this merits explicit coverage.

src/mkdir.js
nfischer: Hmmm this is weird. On unix, ENNOTDIR and ENOENT are for different reasons. On appveyor, I see ENOENT when part of the path is a regular file. Not sure yet how we should handle this.

src/mkdir.js
freitagbr: Is this enum something that should be declared globally? Perhaps in `src/common.js`?

test/.eslintrc.json
freitagbr: Are there any escaped quotes outside of `test/`? I can't find any in `src/`.

nfischer: Guess there's not. Let's keep it here then.

src/mkdir.js
nfischer: Some of the error codes can be shared (e.g., `EEXIST`). However, there's no real advantage to moving these to common.js. It's the difference between `common.codes.EEXIST` and `'EEXIST'`.

When it comes to user-facing error messages, I think each command wants a different message as it sees fit.

freitagbr: Ok, makes sense. This enum could at least be declared globally in this file.

freitagbr (changes requested):

src/mkdir.js
nfischer: Done

src/mkdir.js
freitagbr: Is this still causing the appveyor CI to fail?

nfischer: Yes. I haven't fixed this yet.

nfischer:
Closing out old pull requests.

PR #863, opened 2018-06-25
fix(mocks): fix conflict between mocks and skip

### LINKED ISSUE #862: test infra: mocks and skipOnWin conflict
I observed this while reviewing PR #861. If a test suite uses `mocks.init()` and also uses `skipOnWin`, `skipOnWin` *should* output stderr indicating the test is skipped, but the message is actually suppressed by `mocks`.

We can fix this by capturing `process.stderr.write` on startup, within `utils.js`, instead of using `console.warn()` directly.

nfischer:
This fixes a conflict between mocks.init() and
utils.skipOnWin/skipOnUnix. mocks.init() mocks out process.stderr.write,
which utils.js implicitly depends on.

Instead, preserve stderr.write in a local variable to avoid polluting
mocked stdio and to correctly output warning messages.

Fixes #862
Test: locally apply mocks.init() inside test/which.js

### DIFF
--- a/test/utils/utils.js
+++ b/test/utils/utils.js
@@ -5,6 +5,9 @@ const chalk = require('chalk');
 
 const common = require('../../src/common');
 
+// Capture process.stderr.write, otherwise we have a conflict with mocks.js
+const _processStderrWrite = process.stderr.write.bind(process.stderr);
+
 function numLines(str) {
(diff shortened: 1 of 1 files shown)


freitagbr:
LGTM

PR #864, opened 2018-06-27
chore: bump dev dependencies and add package-lock

nfischer:
No change to logic.

This reduces install spam by bumping codecov to avoid depending on a
deprecated graceful-fs version, as well as adding a package-lock.json.


freitagbr:
LGTM

PR #869, opened 2018-07-04
Add preserve option to cp

### LINKED ISSUE #771: Preserve timestamps on cp?

nfischer:
This is a redo of #841.

This adds the `-p` flag to `cp()` to preserve certain attributes, such as timestamps.

Fixes #771

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ Available options:
 + `-r`, `-R`: recursive
 + `-L`: follow symlinks
 + `-P`: don't follow symlinks
++ `-p`: preserve file mode, ownership, and timestamps
 
 Examples:
 
--- a/src/cp.js
+++ b/src/cp.js
@@ -76,6 +76,8 @@ function copyFileSync(srcFile, destFile, options) {
 
     if (options.preserve) {
       fs.fchownSync(fdw, srcStat.uid, srcStat.gid);
+      // Note: utimesSync does not work (rounds to seconds), but futimesSync has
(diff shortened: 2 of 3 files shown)


nfischer:
Work left:

 - [ ] the sleeps are misplaced. Remove them. Replace them with correct solution (use touch to modify source timestamps to something other than "now")
 - [ ] fix this for directories (that was never implemented)
 - [ ] add comments I requested in the original PR

pranshu667:
Hey, I am new to this.
This is still open right ?

monstrito (commented):
@maldito-dot
GET / HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Cookie: sessionid=foo;another-cookie=bar
Host: pie.dev
User-Agent: HTTPie/0.9.9

nfischer:
If anyone wants to take this over, help would be welcome! 🙏🏻

nfischer:
Hmm. There appears to be a node v9 bug which limits the resolution of atime/mtime to seconds (instead of milliseconds). I wonder if changing the test values from `12345` to `12345000` will work.

nfischer:
Blocked on #1074

nfischer:
Actually... maybe `cp -p` **is** supposed to update the source file's atime. This behavior seems weird, but it seems to be how bash behaves.

Issue #834, opened 2018-04-06
Robustly clean up and restore state after tests


nfischer:
Inspired by #833. We should create a robust solution to ensure each test runs with fresh state, and we successfully cleanup after each test.

Right now, some tests can `cd()` to change the working directory, and we won't properly cleanup after them. As an example, here's a bad `afterEach()` which assumes the test doesn't change working directory:

https://github.com/shelljs/shelljs/blob/9035b27403c7e3c3dc801939d20dec8fe6143c06/test/ln.js#L18-L20

Another issue is that `t.context.tmp` usually refers to a **relative path** returned by `utils.getTempDir()`. I think we can be more robust by returning an absolute path.

---

We should create utility functions to do robust setup and teardown, to be used by every test suite (and the test suites can extend this to provide additional setup).

@freitagbr any additional thoughts?

freitagbr:
I agree that `t.context.tmp` should contain the full path of the temporary directory, not the relative. But, the setup and teardown functions should take care of setting all of this up. The `setup` and `teardown` functions might look like the `beforeEach` and `afterEach` in `test/ln.js`, for example. Then, in every test, we could just put `test.beforeEach(setup)` and `test.afterEach(teardown)`. I don't know if ava has a way to do this globally for every test.

Another improvement might be using a dedicated temp directory module like [`tmp`](https://www.npmjs.com/package/tmp) or [`tempy`](https://www.npmjs.com/package/tempy), though our solution should continue to work fine.

nfischer:
> I don't know if ava has a way to do this globally for every test.

@freitagbr Huh? `beforeEach()` does run before every test, as-is. What did you mean?

nfischer:
I'll write up a PR in the next couple days.

freitagbr:
> Huh? beforeEach() does run before every test, as-is.

`beforeEach()` runs before every test in that file, but not for every file. Some test frameworks have a way to specify a `beforeEach` for every test in every file, but from reading the documentation, I don't think ava has a way to do this. But that's fine, it's not a big deal if each test file has `beforeEach` in it.

nfischer:
Another thing (caught this in #874), we should robustly set `process.env` to its original value (resetting each property). This should happen in `afterEach.always`.

PR #874, opened 2018-07-12
Fix(which): match only executable files (#657)

### LINKED ISSUE #657: which() should check executability before returning a value
`which()` should only return a value if it finds a matching file in the PATH **and** that file is executable.

I'm not sure if it's possible to reliably check executability on Windows, but this should at least be done for unix. [This section](https://nodejs.org/api/fs.html#fs_file_mode_constants) of the docs suggests a way of using `fs.statSync()` to check this info.

The right way would be to also check if the current user matches the `user`, `group`, or `all` fields as well, but I think this might be tricky. It should be sufficient to just assume that we do have a match, and just do an OR of all three (ex. `userCanExecute() || groupCanExecute() || allCanExecute()`).

termosa:
Fixes #657.

### DIFF
--- a/src/which.js
+++ b/src/which.js
@@ -18,8 +18,17 @@ function splitPath(p) {
   return p ? p.split(path.delimiter) : [];
 }
 
+function isExecutable(pathName) {
+  try {
+    fs.accessSync(pathName, fs.constants.X_OK);
+  } catch (err) {
+    return false;
+  }
+  return true;
+}
+
 function checkPath(pathName) {
-  return fs.existsSync(pathName) && !common.statFollowLinks(pathName).isDirectory();
+  return fs.existsSync(pathName) && !common.statFollowLinks(pathName).isDirectory() && isExecutable(pathName);
 }
 
 //@
--- a/src/which.js
+++ b/src/which.js
@@ -13,6 +13,9 @@ common.register('which', _which, {
 // set on Windows.
 var XP_DEFAULT_PATHEXT = '.com;.exe;.bat;.cmd;.vbs;.vbe;.js;.jse;.wsf;.wsh';
 
+// For earlier versions of NodeJS that doesn't have a list of constants (< v6)
+var FILE_EXECUTABLE_MODE = 1;
+
 // Cross-platform method for splitting environment `PATH` variables
 function splitPath(p) {
   return p ? p.split(path.delimiter) : [];
@@ -20,7 +23,7 @@ function splitPath(p) {
 
 function isExecutable(pathName) {
   try {
-    fs.accessSync(pathName, fs.constants.X_OK);
+    fs.accessSync(pathName, FILE_EXECUTABLE_MODE);
   } catch (err) {
     return false;
   }
--- a/src/which.js
+++ b/src/which.js
@@ -16,6 +16,10 @@ var XP_DEFAULT_PATHEXT = '.com;.exe;.bat;.cmd;.vbs;.vbe;.js;.jse;.wsf;.wsh';
 // For earlier versions of NodeJS that doesn't have a list of constants (< v6)
 var FILE_EXECUTABLE_MODE = 1;
 
+function isWindowsPlatform() {
+  return process.platform === 'win32';
+}
+
 // Cross-platform method for splitting environment `PATH` variables
 function splitPath(p) {
   return p ? p.split(path.delimiter) : [];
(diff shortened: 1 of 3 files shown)


test/which.js
```diff
@@ -69,5 +70,25 @@ test('Searching with -a flag returns an array with first item equals to the regu
   t.falsy(shell.error());
   t.truthy(resultForWhich);
   t.truthy(resultForWhichA);
-  t.is(resultForWhich.toString(), resultForWhichA[0].toString());
+  t.is(resultForWhich.toString(), resultForWhichA[0]);
+});
+
+test('None executable files does not appear in the result list', t => {
+  const commandName = 'node'; // Should be an existing command
+  const pathEnv = process.env.path || process.env.Path || process.env.PATH;
```
nfischer: This doesn't look right. This should be `process.env.PATH`.

Only all-caps `$PATH` works on unix. On Windows, `process.env.path` is totally the same as `process.env.PATH` (and all variations of capitalization of "path").

test/which.js
```diff
@@ -69,5 +70,25 @@ test('Searching with -a flag returns an array with first item equals to the regu
...
+  const extraPath = path.resolve(__dirname, 'resources', 'which');
+  const matchingFile = path.resolve(extraPath, commandName);
+
+  // make sure that file is exists (will throw error otherwise)
+  fs.statSync(matchingFile);
```
nfischer: nit: can you make this `t.truthy(fs.existsSync())`? We generally prefer assertion failures over exceptions.

src/which.js
```diff
@@ -13,13 +13,25 @@ common.register('which', _which, {
 // set on Windows.
 var XP_DEFAULT_PATHEXT = '.com;.exe;.bat;.cmd;.vbs;.vbe;.js;.jse;.wsf;.wsh';
 
+// For earlier versions of NodeJS that doesn't have a list of constants (< v6)
```
nfischer: Could you add a TODO to use the node-provided constant on v6+? We'll eventually drop v4/5.

nfischer: Also, I'm trusting that (1) this value is respected on v4, even though the constant does not exist, and (2) the value is treated the same as `F_OK` on Windows, just as `X_OK` is documented:

> Flag indicating that the file can be executed by the calling process. This has no effect on Windows (will behave like fs.constants.F_OK).

test/which.js
```diff
@@ -69,5 +70,25 @@ test('Searching with -a flag returns an array with first item equals to the regu
...
+
+  process.env.path = process.env.Path = process.env.PATH = extraPath + path.delimiter + pathEnv;
+  const resultForWhich = shell.which(commandName);
+  const resultForWhichA = shell.which('-a', commandName);
+  t.falsy(shell.error());
+  t.truthy(resultForWhich);
+  t.truthy(resultForWhichA);
+  t.truthy(resultForWhichA.length);
+  t.not(resultForWhich.toString(), matchingFile);
+  t.not(resultForWhichA[0], matchingFile);
```
nfischer: Could you rewrite this as:

```js
t.falsy(resultForWhichA.includes(matchingFile));
```

Also, I think it makes sense to skip this test on Windows, or to assert the array *does* contain the value on Windows. WDYT?

src/which.js
termosa: True, on Windows `X_OK` and `F_OK` are equivalent. But on Windows platform file extensions are used to determine if file is executable or not.
I will not do this check for Windows platform.

test/which.js
termosa: `includes` is not handled by Ava and it is not supported in node@5. I would also save `[0]` index to make sure that order is proper.

src/which.js
nfischer: > I will not do this check for Windows platform.

Sounds good

termosa:
@nfischer done

test/which.js
```diff
@@ -69,5 +70,27 @@ test('Searching with -a flag returns an array with first item equals to the regu
...
+  const resultForWhich = shell.which(commandName);
+  const resultForWhichA = shell.which('-a', commandName);
+  t.falsy(shell.error());
+  t.truthy(resultForWhich);
+  t.truthy(resultForWhichA);
+  t.truthy(resultForWhichA.length);
+  t.not(resultForWhich.toString(), matchingFile);
+  t.is(resultForWhichA.indexOf(matchingFile), -1);
+
+  process.env.PATH = pathEnv;
```
nfischer: Oh sorry, I must have missed this. This should go in an `test.afterEach.always` block. Example:

https://github.com/shelljs/shelljs/blob/6d66a1af98430bc0275e2125d4cae41ccb0de645/test/exec.js#L14-L17

This shouldn't block this change. I added a comment to #834 to track this.

nfischer:
Thanks @termosa!

PR #876, opened 2018-07-19
grep includes the i flag

ppsleep:
grep includes the i flag to ignored upper/lower case differences


src/grep.js
```diff
@@ -17,7 +18,8 @@ common.register('grep', _grep, {
 //@ Available options:
 //@
 //@ + `-v`: Invert `regex_filter` (only print non-matching lines).
-//@ + `-l`: Print only filenames of matching files
+//@ + `-l`: Print only filenames of matching files.
+//@ + `-i`: Ignored upper/lower case differences.
```
nfischer: nit: can you change this sentence to simply "Ignore case."?

test/resources/case1
```diff
@@ -0,0 +1 @@
+Test3
```
nfischer: This is OK, but could you move these files under `test/resources/grep/case1*`? That should fix the other failures.

ppsleep: OK, I moved it, and sorry for my thoughtless actions.

src/grep.js
```diff
@@ -48,6 +50,9 @@ function _grep(options, regex, files) {
     }
 
     var contents = file === '-' ? pipe : fs.readFileSync(file, 'utf8');
+    if (options.ignoreCase) {
+      regex = new RegExp(regex, 'i');
```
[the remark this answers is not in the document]
ppsleep: Oh, it's a low-level error, i fixed it, sorry.

nfischer:
@ppsleep thanks for the great contribution!

Issue #878, opened 2018-07-23
mv silently swallows errors


vgalka-sl:
### Node version
v8.9.1

### ShellJS version
0.8.2

### Operating system
Windows

### Description of the bug
Some errors are silently ignored.

Example:
```sh.mv('foo/a/file.js', 'bar/a/file.js');```
Fails when destination directory `bar` does not exist. But no error is returned.

nfischer:
Thanks for reporting this. Confirmed this happens on unix too (via `shx v0.3.1 (using ShellJS v0.8.2)`)

iobuhov:
Had same issue with `0.8.5` on mac

PR #879, opened 2018-07-23
Report forgotten errors from 'mv'

vgalka-sl:
Fix for issue #878.

### DIFF
--- a/src/mv.js
+++ b/src/mv.js
@@ -110,6 +110,8 @@ function _mv(options, sources, dest) {
         // exception to pass up to the top level.
         cp('-r', src, thisDest);
         rm('-rf', src);
+      } else {
+          common.error(e.message, { continue: true });
       }
     }
   }); // forEach(src)


src/mv.js
```diff
@@ -110,6 +110,8 @@ function _mv(options, sources, dest) {
         // exception to pass up to the top level.
         cp('-r', src, thisDest);
         rm('-rf', src);
+      } else {
+        common.error(e.message, { continue: true });
```
nfischer: Could you catch a specific `e.code` and output a message like `mv: cannot move 'foo/a/file.js' to 'bar/a/file.js': No such file or directory`?

Also, could you add an else branch with `throw e`? The distinction is that we only want to use `common.error()` for expected error conditions, and would prefer to fail loudly with an exception for unexpected conditions (so that we can fix these as they're reported).

nfischer:
@vgalka-sl any update on this issue? Let me know if I can clarify anything.

PR #880, opened 2018-07-23
docs: remove gitter badge

nfischer:
No change to logic.

We don't really use the gitter chatroom right now, so we shouldn't
direct users to chat there. It's best to have issues reported directly
on Github.

This PR removes the gitter badge from the README.


nfischer:
Ping @freitagbr

Issue #788, opened 2017-10-18
Reduce quote escaping by using template literals


nfischer:
I noticed that we're escaping quote characters in some of our test descriptions ([example](https://github.com/shelljs/shelljs/blob/master/test/ls.js#L38)). We should be able to avoid this by preferring [template literals](https://ponyfoo.com/articles/template-literals-strictly-better-strings) for such scenarios. The motivation is to improve readability.

We should also consider:

 1. Prefer template literals over string concatenation ([eslint rule](https://eslint.org/docs/rules/prefer-template))
 2. Prefer template literals everywhere (I don't think there's any real downside) ([eslint rule](https://eslint.org/docs/rules/quotes))

nfischer:
To clarify, this should be limited to only our tests (because those use ES6). Once we drop support for node 4/5, we can expand this to cover the whole source tree.

Preferring template literals for all of the tests might be a bit harsh, but preferring them over string concatenation sound reasonable.

blubrick:
I would be happy to have a crack at this gruntwork.  It seems pretty simple, if a little tedious(hence the 'chore' label, I guess).

So I cloned the repo and ran `npm install` 

Is that right - 51 package vulnerabilities?  I figure I gotta be doing something wrong.  (Ubuntu 18.04, node 12.7.0, npm 6.11.3)  

Fortunately, it seems that only one of them is considered to be a `SEMVER WARNING`.  The remaining 50 (lots of dupes) can be resolved via `npm audit fix` upgrading 6 packages, but then my build env will not not be the same as the repo. 

Do you think I should raise this as a separate issue?

nfischer:
@blubrick yes, please file a separate issue. I'm happy to take a look at the report (and would also appreciate PRs to update flagged packages if doing so does not break the project).

nfischer:
> Do you think I should raise this as a separate issue?

Filed as #969

PR #889, opened 2018-09-18
feat(cp): support update flag when recursing

### LINKED ISSUE #808: cp -Ru respects the -R but not the -u
Love the lib - was looking to use this to reduce copy times on builds, but ran into the following:

8.6.0
shelljs 0.7.8, tested both directly and via shx 0.2.2
### Operating system:
Arch Linux (kernel 4.13.12-1)
### Description of the bug:
When combining `recursive` `-R` and `update` `-u` options in both `shelljs` and `shx`, "recursive" overrides and ignores "update" - files are recursively copied regardless of whether the src files are newer than dist.

Test case (assuming src/static and dist/static have contents)
```javascript

// overwrite dist/static, verify that all timestamps in dist are now newer than those in src
shell.cp('-R', 'src/static', 'dist/');
console.log(shell.ls('-l', 'src/static').stdout, '\n' + shell.ls('-l', 'dist/static').stdout);
console.log('----------');

setTimeout(() => {
  shell.cp('-Ru', 'src/static/', 'dist/');
  console.log(shell.ls('-l', 'dist/static').stdout);

  // timestamps shouldn't change, but are ~1s newer
}, 1000);
```

### Expected result
Same as
```bash
cp -R src/static/ dist/ && ls -l src/static && ls -l dist/static
echo '\n-----\n'
sleep 1
cp -Ru src/static/ dist/ && ls -l dist/static
```

I think I see where the fix is - just wanted to file this before I forgot in case I don't end up having time in the near future to create a PR.

joshi-sh:
Fixed `cp -Ru` ignoring `-u`, and added a test.

Fixes #808

### DIFF
--- a/test/cp.js
+++ b/test/cp.js
@@ -463,7 +463,6 @@ test('-Ru respects the -u flag recursively (don\'t update newer file)', t => {
   shell.ShellString('Source File Contents\n').to(sourceFile);
   shell.ShellString('Destination File Contents\n').to(destFile);
   // End setup
-  
   // Get the old mtime for dest
   const oldTime = fs.statSync(destFile).mtimeMs;
   // Set the source file to be older than the destination file
@@ -486,7 +485,6 @@ test('-Ru respects the -u flag recursively (update older file)', t => {
   shell.ShellString('Source File Contents\n').to(sourceFile);
   shell.ShellString('Destination File Contents\n').to(destFile);
   // End setup
-  
   // Get the old mtime for dest
   const oldTime = fs.statSync(destFile).mtimeMs;
   // Set the destination file to be older than the source file
--- a/test/cp.js
+++ b/test/cp.js
@@ -458,8 +458,8 @@ test('-Ru respects the -u flag recursively (don\'t update newer file)', t => {
   const destDir    = `${dir}/new`;
   const destFile   = `${destDir}/file`;
   [sourceDir, destDir].forEach(d => shell.mkdir('-p', d));
-  ShellString('foo\n').to(sourceFile);
-  ShellString('bar\n').to(destFile);
+  shell.ShellString('foo\n').to(sourceFile);
+  shell.ShellString('bar\n').to(destFile);
   // Get the old mtime for dest
   const oldTime = fs.statSync(destFile).mtimeMs;
   // Send the old file to two days ago
@@ -477,8 +477,8 @@ test('-Ru respects the -u flag recursively (update older file)', t => {
   const destDir    = `${dir}/new`;
   const destFile   = `${destDir}/file`;
   [sourceDir, destDir].forEach(d => shell.mkdir('-p', d));
-  ShellString('foo\n').to(sourceFile);
-  ShellString('bar\n').to(destFile);
+  shell.ShellString('foo\n').to(sourceFile);
+  shell.ShellString('bar\n').to(destFile);
   // Get the old mtime for dest
   const oldTime = fs.statSync(destFile).mtimeMs;
   // Send the source file to two days ahead
--- a/test/cp.js
+++ b/test/cp.js
@@ -451,17 +451,43 @@ test('-R implies -P', t => {
   });
 });
 
-test('-Ru respects the -u flag recursively', t => {
-  const [source, dest] = ['foo', 'bar'].map(s => `test/resources/cp/cp-Ru/${s}/file`);
-  // First, update the dest file
-  shell.touch(dest);
+test('-Ru respects the -u flag recursively (don\'t update newer file)', t => {
+  const dir = `${t.context.tmp}/cp-Ru`;
+  const sourceDir  = `${dir}/old`;
+  const sourceFile = `${sourceDir}/file`;
+  const destDir    = `${dir}/new`;
+  const destFile   = `${destDir}/file`;
+  [sourceDir, destDir].forEach(d => shell.mkdir('-p', d));
+  ShellString('foo\n').to(sourceFile);
+  ShellString('bar\n').to(destFile);
   // Get the old mtime for dest
-  const oldTime = fs.statSync(dest).mtimeMs;
+  const oldTime = fs.statSync(destFile).mtimeMs;
+  // Send the old file to two days ago
+  shell.touch('-m', oldTime - (2 * 24 * 60 * 60 * 1000), sourceFile);
   // Now, copy the old dir to the new one
-  shell.cp('-Ru', source, dest);
+  shell.cp('-Ru', sourceDir, destDir);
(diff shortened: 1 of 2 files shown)


test/cp.js
```diff
@@ -451,6 +451,19 @@ test('-R implies -P', t => {
   });
 });
 
+test('-Ru no longer ignores -u', t => {
```
nfischer: nit: `-Ru respects the -u flag recursively`

test/cp.js
```diff
@@ -451,6 +451,19 @@ test('-R implies -P', t => {
...
+  const [source, dest] = ['foo', 'bar'].map(s => `test/resources/cp/cp-Ru/${s}/file`);
+  // First, update the dest file
+  shell.touch(dest);
```
nfischer: You also need to update the source directory to an older time stamp, right? Otherwise, we'll see race conditions in very-fresh checkouts.

test/cp.js
```diff
@@ -451,6 +451,19 @@ test('-R implies -P', t => {
   });
 });
 
+test('-Ru no longer ignores -u', t => {
+  const [source, dest] = ['foo', 'bar'].map(s => `test/resources/cp/cp-Ru/${s}/file`);
```
nfischer: You should copy to `t.context.tmp`. Treat `test/resources/` as read-only.

test/cp.js
```diff
@@ -451,6 +451,19 @@ test('-R implies -P', t => {
...
+  shell.touch(dest);
+  // Get the old mtime for dest
+  const oldTime = fs.statSync(dest).mtimeMs;
+  // Now, copy the old dir to the new one
+  shell.cp('-Ru', source, dest);
+  // Get the new mtime for dest
+  const newTime = fs.statSync(dest).mtimeMs;
+  t.is(oldTime, newTime);
+});
+
```
nfischer: This test case is mostly lgtm, but you need another test case where src is newer than dest. `-u` means the copy should not happen. You should compare file contents to verify the copy does not occur (a successful copy would change file contents, but `-u` prevents the copy).

nfischer:
@joshi-sh can you check the `npm run gendocs` warning?

test/cp.js
joshi-sh: If src is newer than dest, then the copy should (regardless of `-u`) occur. That said, checking file contents is a better way to test cp than checking timestamps, resolving an issue you raised in an earlier comment. I'll set it up to do that instead, and add the additional test.

joshi-sh:
Running on Ubuntu 16.04.3 on top of WSL with node v8.12.0 produces no errors

test/cp.js
joshi-sh: In a later comment, you've mentioned checking file contents. That would resolve this issue as well.

nfischer:
@joshi-sh it might not be a docs issue, but there might be modified or extra files leftover after running tests + lint + generate docs. Seems to be v8.x.y specific.

test/cp.js
nfischer: Yeah, file contents is fine to use here as well.

test/cp.js
```diff
@@ -451,6 +451,45 @@ test('-R implies -P', t => {
...
 
+test('-Ru respects the -u flag recursively (don\'t update newer file)', t => {
+  const dir = `${t.context.tmp}/cp-Ru`;
+  const sourceDir = `${dir}/old`;
+  const sourceFile = `${sourceDir}/file`;
+  const destDir = `${dir}/new`;
+  const destFile = `${destDir}/file`;
+  [sourceDir, destDir].forEach(d => shell.mkdir('-p', d));
+  shell.ShellString('foo\n').to(sourceFile);
+  shell.ShellString('bar\n').to(destFile);
```
nfischer: I would like to separate setup from test logic. Two (alternative) suggestions:

 1. Instead of setting up at runtime, commit this directory structure under `test/resources/`, or
 2. Add an empty line between setup code and test logic (possibly add a comment above such as `// Setup code`)

---

You'll need to continue modifying the time stamps at runtime, as you have here.

test/cp.js
```diff
@@ -451,6 +451,45 @@ test('-R implies -P', t => {
...
+  // Get the old mtime for dest
+  const oldTime = fs.statSync(destFile).mtimeMs;
+  // Send the old file to two days ago
```
nfischer: nit: you can better document this by creating a constant `TWO_DAYS_IN_MS`.

You could then make this comment more descriptive: `// Set the sourceFile to be older than destFile`.

test/cp.js
```diff
@@ -451,6 +451,45 @@ test('-R implies -P', t => {
...
+  [sourceDir, destDir].forEach(d => shell.mkdir('-p', d));
+  shell.ShellString('foo\n').to(sourceFile);
+  shell.ShellString('bar\n').to(destFile);
+  // Get the old mtime for dest
+  const oldTime = fs.statSync(destFile).mtimeMs;
+  // Send the source file to two days ahead
+  shell.touch('-m', oldTime + (2 * 24 * 60 * 60 * 1000), sourceFile);
+  // Now, copy the old dir to the new one
+  shell.cp('-Ru', sourceDir, destDir);
+  // Get the new mtime for dest
```
nfischer: nit: I think this comment is leftover from a previous commit (you're not checking mtimes anymore)

test/cp.js
```diff
@@ -451,6 +451,45 @@ test('-R implies -P', t => {
...
 });
 
+test('-Ru respects the -u flag recursively (don\'t update newer file)', t => {
+  const dir = `${t.context.tmp}/cp-Ru`;
+  const sourceDir = `${dir}/old`;
+  const sourceFile = `${sourceDir}/file`;
+  const destDir = `${dir}/new`;
+  const destFile = `${destDir}/file`;
+  [sourceDir, destDir].forEach(d => shell.mkdir('-p', d));
+  shell.ShellString('foo\n').to(sourceFile);
```
nfischer: nit: file contents could be more descriptive. How about `source file contents` and `destination file contents`?

test/cp.js
```diff
@@ -451,6 +451,45 @@ test('-R implies -P', t => {
   });
 });
 
+test('-Ru respects the -u flag recursively (don\'t update newer file)', t => {
```
nfischer: You **might** be able to replace this with a template literal to avoid escaping the apostrophe. I can't remember if our lint currently forbids this (see #788). If lint complains, just leave as-is.

nfischer:
> Running npm run gendocs on Ubuntu 16.04.3 on top of WSL with node v8.12.0 produces no errors

Fixed the issue in #896. Please rebase.

PR #892, opened 2018-10-10
Fix silent exec

### LINKED ISSUE #851: silent does not always work
node 8.9.1
shelljs 0.8.1
mac osx

I expect this code to not produce any output but it does. 

```

shell.exec('git rev-parse --abbrev-ref  master2@{upstream}', {silent:true});
```

Run it in a git repository that does not have a branch `master2` and you should see an error printed

```
exec: fatal: no such branch: 'master2'
```

### LINKED ISSUE #891: Redundant printing of stderr when using exec
v10.11.0

0.8.2

### Operating system:
High Sierra

### Description of the bug:
When using synchronous `exec` for some reason after running the command, `exec` itself prints `exec: [stderr]` to console again. I have searched the documentation, the issues, and the code but can't seem to figure out how to silence this. My usage is for `react-scripts test` via `create-react-app` and linting with `flow`. Is it possible to stop this post-`exec` reprinting of `stderr` and if not would it be reasonable to request removing this behavior? 

Using the silent option for `exec` does the opposite of what I want. All the output is silenced except this `exec: [stderr]` reprinting. Setting silent with `config` kills everything. This is especially troublesome with testing since test cases get printed live then after testing completes, the entire log gets printed again but prefixed with `exec: `.

To illustrate if I do something stupid like:
```javascript
shell.exec('cat nonexistentfile');
```
I get in output:
```
cat: nonexistentfile: No such file or directory
exec: cat: nonexistentfile: No such file or directory
```

nfischer:
Based on #861 

Fixes #851 
Fixes #891

### DIFF
--- a/src/exec.js
+++ b/src/exec.js
@@ -96,7 +96,7 @@ function execSync(cmd, opts, pipe) {
   try { common.unlinkSync(stdoutFile); } catch (e) {}
 
   if (code !== 0) {
-    common.error(stderr, code, { continue: true });
+    common.error(stderr, code, { continue: true, silent: opts.silent });
   }
   var obj = common.ShellString(stdout, stderr, code);
   return obj;
--- a/src/exec.js
+++ b/src/exec.js
@@ -96,7 +96,10 @@ function execSync(cmd, opts, pipe) {
   try { common.unlinkSync(stdoutFile); } catch (e) {}
 
   if (code !== 0) {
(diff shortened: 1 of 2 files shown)


nfischer:
CI is busted right now. I think an `npm` update is modifying `package-lock.json`. For some reason it only repros on node v8. Same thing blocks #889 (CC @joshi-sh).

nfischer:
This was released in v0.8.3.

PR #894, opened 2018-10-10
chore(travis): fix packagelock for node v8

### LINKED ISSUE #893: Travis CI currently broken
Due to a `npm` change, `npm install` modifies `package-lock.json`, which turns bots red (we check for modified files at the end of the script).

Blocking #892, #889, and any other PRs.

nfischer:
CI was broken because the npm which ships with node v8 (and only that
version of npm, for some reason), would modify package-lock.json when we
ran `npm install`. This triggered our post-test hook which enforces docs
changes: the modified file was detected as a false-positive for
uncommitted docs changes.

This fixes the issue by committing those docs changes. This seems to be
compatible with other npm versions.

Fixes #893


nfischer:
Unfortunately, this wasn't the righht fix.

PR #901, opened 2018-11-08
fix: Exit 1 with empty string if no match

### LINKED ISSUE #900: grep exit status and extra newlines
v8.11.3

0.8.2

### Operating system:
Mac OS X

### Description of the bug:
`grep()` returns 0, even if a string is not found, and the resulting string also contains a newline.

(given a file `testfile` without the string asdfasdf at the beginning of a line):

```javascript

console.log(shell.grep(/^asdfasdf/, 'testfile'))
```
Output of above:
```
{ [String: '\n']
  stdout: '\n',
  stderr: null,
  code: 0,
  cat: [Function: bound ],
  exec: [Function: bound ],
  grep: [Function: bound ],
  head: [Function: bound ],
  sed: [Function: bound ],
  sort: [Function: bound ],
  tail: [Function: bound ],
  to: [Function: bound ],
  toEnd: [Function: bound ],
  uniq: [Function: bound ] }
```
I don't believe you'll have even a newline if you grep for something that isn't found, and you'll definitely get a non-0 status code if the string isn't found.

```bash
$ grep -q ^asdfasdf testfile 
$ echo $?
1
$ grep ^asdfasdf testfile > /tmp/asdf
$ cat -v -e /tmp/asdf 
$ file /tmp/asdf 
/tmp/asdf: empty
```
As best I could tell, the only reliable way to detect the absence of a string to use something like:
```javascript
shell.grep(/^asdfasdf/, 'testfile').stdout.trim())
```

Of course, at this point, it might be easier to use regexp on the file directly, but this still seems inconsistent with the behavior of the `grep` command.

wyardley:
Resolve an issue where grep returned 0 (with a `\n` as the output), even when the regex doesn't match.

Fixes #900

### DIFF
--- a/src/grep.js
+++ b/src/grep.js
@@ -68,6 +68,10 @@ function _grep(options, regex, files) {
     }
   });
 
+  if (grep.length === 0 && common.state.errorCode !== 2) {
+    // We didn't hit the error above, but pattern didn't match
+    common.error('', { silent: true });
+  }
   return grep.join('\n') + '\n';
 }
 module.exports = _grep;
--- a/test/grep.js
+++ b/test/grep.js
@@ -49,6 +49,13 @@ test('if at least one file is missing, this should be an error', t => {
   t.is(result.code, 2);
 });
 
+test("multiple files, one doesn't exist, one doesn't match", t => {
+  const result = shell.grep(/oogabooga/, 'test/resources/file1.txt',
+    'test/resources/filedoesnotexist.txt');
+  t.truthy(shell.error());
+  t.is(result.code, 2);
+});
+
 //
 // Valids
 //
@@ -125,6 +132,17 @@ test('one file, * in string-regex, make sure * is not globbed', t => {
   const result = shell.grep('l*\\.js', 'test/resources/grep/file');
   t.falsy(shell.error());
   t.is(result.toString(), 'this line ends in.js\nlllllllllllllllll.js\n');
+});
+
+test("one file, pattern doesn't match", t => {
+  const result = shell.grep('notfoundstring', 'test/resources/grep/file');
+  t.truthy(shell.error());
+  t.is(result.toString(), '');
+  t.is(result.stdout, '');
+  // TODO(#900): "grep: " isn't really the correct stderr output, but we need a
+  // non-empty string so `shell.error()` is truthy.
+  t.is(result.stderr, 'grep: ');
+  t.is(result.code, 1);
 });
 
 test('-l option', t => {


wyardley:
@nfischer 
The failure is on those other two test cases on line 39 and 48; seems to be returning both the original error, and the new one, even with that explicit return.

```
     38:   t.truthy(shell.error());                                          
   39:   t.is(result.stderr, 'grep: no such file or directory: /asdfasdf');
   40:   t.is(result.code, 2);                                             

  Difference:

  - `grep: no such file or directory: /asdfasdf␊
  - grep: `
  + `grep: no such file or directory: /asdfasdf```

**Comment by nfischer** [comment]
@wyardley can you add "Fixes #900" to the PR description?

File: src/grep.js [hunk 1]
```diff
@@ -68,6 +68,10 @@ function _grep(options, regex, files) {
     }
   });
 
+  if (grep.length === 0) {
```
nfischer: Ah, so this is an actual bug. There's a few reasons the `grep` array may be empty:

 1. No files (but we exit early for this
 2. Files are specified, but we couldn't read any of them (perhaps files don't exist)
 3. We can read some files but not others. Of the files we can read, none match the regex
 4. We can read all files, but none contain the regex

Case 1 is [already handled](https://github.com/shelljs/shelljs/blob/4bd22e77423182219cd43f0c8d38621b1c957986/src/grep.js#L37). You filed a ticket about case 4.

For cases 2 and 3, the correct behavior is a non-zero exit code, and stderr should say something like `grep: no such file or directory: nameoffile.txt` (repeated for each problematic file). Error code should be 2, as this indicates a problematic file.

---

I think you can change this if-condition to also check `common.state.error` or `common.state.errorCode`. Don't modify this value.

wyardley: Ok, thanks. That’s what I thought was happening. Wasn’t sure what to add to the conditional, that info is very helpful

test/grep.js
```diff
@@ -127,6 +134,15 @@ test('one file, * in string-regex, make sure * is not globbed', t => {
   t.is(result.toString(), 'this line ends in.js\nlllllllllllllllll.js\n');
 });
 
+test("one file, pattern doesn't match", t => {
```
wyardley: Do you want this under 'valids' or 'invalids?
Even though it's a non-0 exit code,  I would personally consider testing for the absence of something to be a "valid" use case, but I'm happy to move it.

wyardley:
@nfischer Ok, how's that look. I added another test case for the specific case where there are two files, one which dne and the other of which doesn't match.

test/grep.js
```diff
@@ -49,6 +49,13 @@ test('if at least one file is missing, this should be an error', t => {
   t.is(result.code, 2);
 });
 
+test("multiple files, one doesn't exist, one doesn't match", t => {
+  const result = shell.grep(/oogabooga/, 'test/resources/file1.txt',
+    'test/resources/thisonedne');
```
nfischer: Generally prefer to expand abbreviations:

```suggestion
    'test/resources/filedoesnotexist.txt');
```

test/grep.js
nfischer: This spot LGTM

test/grep.js
```diff
@@ -127,6 +134,15 @@ test('one file, * in string-regex, make sure * is not globbed', t => {
   t.is(result.toString(), 'this line ends in.js\nlllllllllllllllll.js\n');
 });
 
+test("one file, pattern doesn't match", t => {
+  const result = shell.grep('notfoundstring', 'test/resources/grep/file');
+  t.truthy(shell.error());
+  t.is(result.toString(), '');
+  t.is(result.stdout, '');
+  t.is(result.stderr, 'grep: ');
```
nfischer:
```suggestion
  // TODO(#900): "grep: " isn't really the correct stderr output, but we need a
  // non-empty string so `shell.error()` is truthy.
  t.is(result.stderr, 'grep: ');
```

nfischer: Also, please double-check that the suggestion's line-length falls within our limit

wyardley:
@nfischer applied your suggestions and squashed

Nel75:
Hi. This fix is still not available on the latest release [v0.8.4](https://github.com/shelljs/shelljs/releases/tag/v0.8.4). What is its status? Was there an issue with its packaging? This fix is really needed, as the current grep implementation is broken - I would argue it's a fixing change rather than a breaking one, as anyone relying on exit code 1 to detect the absence of matches is currently broken. Thanks.

nfischer:
I think this landed after v0.8.3 went out. v0.8.4 was a small patch release for a specific bug, which is why this didn't make it in that release either.

I've updated the milestone field to clarify this is expected to be part of v0.9.0. I haven't had time to figure out what needs to make it in that release and what should be punted to the next release.

Nel75:
@nfischer Thanks for the quick reply. Any idea when v0.9.0 might be released? It's been over 2 years now since this fix, so I'll have to use another script in the mean time.

PR #903, opened 2018-11-08
refactor: don't expose tempdir in common.state

### LINKED ISSUE #902: refactor: remove common.state.tempDir
We should remove `common.state.tempDir`. This really doesn't need to be in `common`. It makes more sense to cache this in `tempdir.js` itself.

nfischer:
Previously, the cached `tempdir` value was stored in `common.state`.
Unlike the other `common.state` values, this isn't immediately useful to
other commands (they can just call the tempdir API). So, this moves the
cached value into `tempdir.js`.

This also adds a unit test for the caching behavior, and exposes
test-only helpers to verify this behavior.

Finally, this adds a note to `common.state` that values should generally
be considered read-only, since this can be important for customized
behavior. Although, I recognize our code base has one exception to this
rule (`echo()`), we should strive to maintain this.

Fixes #902
Test: Added a unit test.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -42,11 +42,11 @@ var config = {
 config.reset();
 exports.config = config;
 
+// Note: commands should generally consider these as read-only values.
 var state = {
   error: null,
   errorCode: 0,
   currentCmd: 'shell.js',
-  tempDir: null,
 };
 exports.state = state;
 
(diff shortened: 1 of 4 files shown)


nfischer:
@wyardley interested in doing a code review?

nfischer:
Thanks!

PR #904, opened 2018-11-10
Add test case for sed on empty file

wyardley:
As discussed as an aside in #900, add test case with an empty file.


wyardley:
Not sure about the Appveyor failure - saw it on a few other PRs. force-pushing w/ a comment change didn't seem to fix it tho. Seems unrelated

PR #906, opened 2018-11-13
feat(plugin): support deprecating plugins

### LINKED ISSUE #541: Idea: add a way of warning users about using deprecated functions
Once #524 is merged, it would be good to deprecate `shell.exec()` due to its vulnerabilities and bugs. While we can mark it as deprecated in the docs, since it's a security vulnerability, we may also want to provide deprecation warnings at runtime.

However, since `exec()` still has a valid use case, I don't think we should ever get rid of it (so it's soft deprecation). Also, if we're still allowing people to use it, we should provide an API to skip deprecation warnings if they know what they're doing.

One approach we could take:

 - add `shell.config.showDeprecationWarnings` (defaults to true)
 - then add a function in `common`) to log a deprecation warning to stderr, only when the above option is true
 - in the deprecated functions, we can call this log function explicitly, or perhaps add a plugin option for deprecated functions of the form:

 ```
 deprecationMessage: 'foo is deprecated, use bar instead',
 ```

This idea is still open for discussion, so don't implement it yet until the "help wanted" label is applied.

nfischer:
This adds a new wrap() option named `.deprecated`. This currently accepts
a string value, which will be the deprecation message for the plugin.
Deprecation works on standard commands as well as pipes. The call stack
points directly to the deprecated method, and not to any ShellJS
internals.

Fixes #541
Test: This adds unit tests.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -4,6 +4,7 @@
 
 var os = require('os');
 var fs = require('fs');
+var util = require('util');
 var glob = require('glob');
 var shell = require('..');
 
@@ -312,7 +313,7 @@ exports.randomFileName = randomFileName;
 // command-logging, and other nice things
 function wrap(cmd, fn, options) {
   options = options || {};
-  return function () {
+  var fun = function () {
     var retValue = null;
 
     state.currentCmd = cmd;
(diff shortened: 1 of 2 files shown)


nfischer:
Closing out old pull requests.

PR #907, opened 2018-11-15
chore: remove gitter integration

nfischer:
No change to logic.

This removes gitter integration with the project. I don't monitor gitter
right now, and I've never noticed the integration to be particularly
useful otherwise.

Test: N/A


nfischer:
Also, it's not clear if the gitter integration is even working right now, aside from this PR. Probably fine to remove.

Issue #873, opened 2018-07-12
Drop node v4 support


nfischer:
According to the [Node LTS schedule](https://github.com/nodejs/Release), v4 is no longer maintained, and v6 is in maintenance mode until April 2019.

We shouldn't bother supported unsupported node versions. shelljs@v0.9.0 (our next release) should be the last minor release to support v4.

- [ ] bump `package.engines`
- [ ] change README ("ShellJS is proudly tested on every node release since...")
- [ ] drop v4 & v5 from travis/appveyor
- [ ] do all the same for `shx`

nfischer:
I'll probably write a script to automate most of these steps (and to enforce all of these spots stay in-sync).

nfischer:
Unfortunately, I think we need to bump priority. It seems appveyor node@4 is too flaky (https://github.com/shelljs/shelljs/issues/893#issuecomment-442787807).

PR #908, opened 2018-11-26
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
Test only Node.js 6, 8, 10 (LTS) and stable (11).

https://github.com/nodejs/Release/blob/master/README.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,16 @@
-language: c++
+language: node_js
 sudo: false
-env:
-  - NODE_VERSION="4"
-  - NODE_VERSION="5"
-  - NODE_VERSION="6"
-  - NODE_VERSION="7"
-  - NODE_VERSION="8"
-  - NODE_VERSION="9"
+node_js:
+  - "6"
+  - "8"
+  - "10"
+  - "node"
 
 # keep this blank to make sure there are no before_install steps
 before_install:
 
 install:
-  - rm -rf ~/.nvm
-  - git clone https://github.com/creationix/nvm.git ~/.nvm
-  - source ~/.nvm/nvm.sh
-  - nvm install $NODE_VERSION
-  - node --version
-  - npm --version
-  - npm run ci-or-install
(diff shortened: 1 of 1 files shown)


nfischer:
- For dropping v4/v5, see #873
- v10 has a bug right now (#856) (help would be appreciated)
- We probably *won't* drop v7/v9
- We can't use `npm install`, we have our own script (see #896)
- We chose `c++` as a language a long time ago, and I can't remember why (so I can't tell you if the reason is still relevant). I think it was just to get finer-grained control over our CI steps

DanielRuf:
> We probably won't drop v7/v9

But why? They reached their EOL.

DanielRuf:
> We can't use npm install, we have our own script (see #896)

This produces different results in CI and production (lockfiles are not working there).

> We chose c++ as a language a long time ago, and I can't remember why (so I can't tell you if the reason is still relevant). I think it was just to get finer-grained control over our CI steps

But it does not seem to be used.
I'll check and update the CI script.

DanielRuf:
> Not sure I quite understand. Are you objecting to the use of lockfiles in general, or objecting to our install wrapper?

The first, this is why `npm ci` is a bit dangerous (works in our CI - but not in production).

DanielRuf:
Probably, or use `npm install` on CI.
At least CI often produces a different dependency tree than the installed package in production as it is often a child dependency of another dependency and below the root lockfile it is only locked to the SemVer range in the package.json files (or better: pinned there to a specific version).

Not yet sure what the best solution for everyone is. It's a bit inconsistent across platforms (dev, prod, CI), tools (yarn, npm version) and repository settings (a lockfile might be useful for contributors and CI to build reproducible builds but many big devs like Sindre and others do not use lockfiles for such reasons afaik).

Also see https://github.com/sindresorhus/ama/issues/479

nfischer:
Replaced by #910, #917, and #921

PR #910, opened 2018-11-29
ci: change language to node_js and remove obsolete scripts

DanielRuf:
We can safely use `node_js` on all three supported platforms on Travis CI and remove the obsolete scripts.

See https://github.com/shelljs/shelljs/pull/908#issuecomment-442732567

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,14 @@
-language: c++
+language: node_js
 sudo: false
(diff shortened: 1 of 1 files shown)


nfischer:
@DanielRuf I'll eventually do the same for shelljs/shx (but feel free to send a PR if you get to it first).

PR #921, opened 2018-12-03
chore(node): add v10 and v11 to CI

nfischer:
This adds support for v10 and v11, and removes a TODO.


.travis.yml
```diff
@@ -5,6 +5,8 @@ node_js:
   - 7
   - 8
   - 9
+  - 10
+  - 11
```
DanielRuf: I still recommend `node` (fail early and also auto update when 12 comes out - otherwise we might miss the release of `12` and there could be breaking changes so someone has to update the CI config again when a new major version is released).

nfischer: Thanks for taking a look. You raise a good point, but `node` isn't really the answer: in that case we would test 12 but miss 11, which is not what we want. Even if we didn't test unsupported node releases, we'd still hit this problem between 12 & 13 (at which point both would be supported).

IMO auto updating is not desirable, because a breaking change in 12 would cause CI to fail suddenly on master and CI would reject unrelated PRs.

Let me know if you see a better path forward.

DanielRuf: We do not want to test 11 when 12 is released as it will reach its EOL then and 13 will be released. See https://github.com/nodejs/Release/blob/master/schedule.png

But I guess we can currently keep it as is. Both ways / solutions have their advantages and disadvantages.

PR #925, opened 2019-01-04
test(touch): add coverage for -d option

nfischer:
No change to production logic.

We never had coverage for `touch({'-d': ...})`, so this adds test
coverage. This also updates documentation to clarify the parameter
should be an instance of the `Date` type.

Test: `handles date argument` case


nfischer:
This was covered by some `cp.js` tests, but it's probably clearer to put a test in `touch.js` anyway.

PR #961, opened 2019-08-12
Add boolean `fatal` option to `exec()` function

WesCossick:
This PR introduces a new boolean `fatal` option for the `exec()` function. Like the existing `silent` option, this new option allows you to override the global `fatal` configuration parameter on a per-command basis, like:

```js
exec('example command', {
    fatal: true|false,
});
```

This eliminates the need for the following pattern:

```js
const oldFatal = shell.config.fatal;
shell.config.fatal = false;
exec('some command that might fail');
shell.config.fatal = oldFatal;
```

The old pattern can introduce race conditions in `async` functions, an issue that was surfaced in our code with the new ESLint rule, [require-atomic-updates](https://eslint.org/docs/rules/require-atomic-updates).

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -98,6 +98,7 @@ function error(msg, _code, options) {
     code: 1,
     prefix: state.currentCmd + ': ',
     silent: false,
+    fatal: config.fatal,
   };
 
   if (typeof _code === 'number' && isObject(options)) {
@@ -118,7 +119,7 @@ function error(msg, _code, options) {
   state.error += logEntry;
 
   // Throw an error, or log the entry
-  if (config.fatal && options.fatal !== false) throw new Error(logEntry);
+  if (config.fatal) throw new Error(logEntry);
   if (msg.length > 0 && !options.silent) log(logEntry);
 
   if (!options.continue) {
--- a/src/common.js
+++ b/src/common.js
@@ -118,7 +118,7 @@ function error(msg, _code, options) {
   state.error += logEntry;
 
   // Throw an error, or log the entry
-  if (config.fatal) throw new Error(logEntry);
+  if (config.fatal && options.fatal !== false) throw new Error(logEntry);
   if (msg.length > 0 && !options.silent) log(logEntry);
 
   if (!options.continue) {
--- a/src/common.js
+++ b/src/common.js
@@ -119,7 +119,7 @@ function error(msg, _code, options) {
   state.error += logEntry;
 
   // Throw an error, or log the entry
-  if (config.fatal) throw new Error(logEntry);
+  if (options.fatal) throw new Error(logEntry);
   if (msg.length > 0 && !options.silent) log(logEntry);
 
   if (!options.continue) {
--- a/src/common.js
+++ b/src/common.js
@@ -425,7 +425,7 @@ function wrap(cmd, fn, options) {
         e.name = 'ShellJSInternalError';
         throw e;
       }
-      if (config.fatal) throw e;
+      if (config.fatal || options.handlesFatalDynamically) throw e;
     }
 
     if (options.wrapOutput &&
@@ -451,6 +451,7 @@ var DEFAULT_WRAP_OPTIONS = {
   canReceivePipe: false,
   cmdOptions: null,
   globStart: 1,
+  handlesFatalDynamically: false,
   pipeOnly: false,
   wrapOutput: true,
   unix: true,
--- a/src/common.js
+++ b/src/common.js
@@ -88,7 +88,7 @@ CommandError.prototype = Object.create(Error.prototype);
 CommandError.prototype.constructor = CommandError;
 exports.CommandError = CommandError; // visible for testing
 
-// Shows error message. Throws if config.fatal is true
+// Shows error message. Throws if fatal is true (defaults to config.fatal, overridable with options.fatal)
 function error(msg, _code, options) {
   // Validate input
   if (typeof msg !== 'string') throw new Error('msg must be a string');
--- a/src/exec.js
+++ b/src/exec.js
@@ -12,13 +12,14 @@ common.register('exec', _exec, {
   unix: false,
   canReceivePipe: true,
   wrapOutput: false,
+  handlesFatalDynamically: true,
 });
 
 // We use this function to run `exec` synchronously while also providing realtime
 // output.
 function execSync(cmd, opts, pipe) {
   if (!common.config.execPath) {
-    common.error('Unable to find a path to the node binary. Please manually set config.execPath');
+    common.error('Unable to find a path to the node binary. Please manually set config.execPath', { continue: true, fatal: opts.fatal });
   }
 
   var tempDir = _tempDir();
@@ -187,7 +188,6 @@ function execAsync(cmd, opts, pipe, callback) {
 //@ Guidelines](https://github.com/shelljs/shelljs/wiki/Security-guidelines).
 function _exec(command, options, callback) {
   options = options || {};
-  if (!command) common.error('must specify command');
 
   var pipe = common.readFromPipe();
 
@@ -207,6 +207,8 @@ function _exec(command, options, callback) {
     fatal: common.config.fatal,
     async: false,
   }, options);
+  
+  if (!command) common.error('must specify command', { continue: true, fatal: options.fatal });
 
   if (options.async) {
     return execAsync(command, options, pipe, callback);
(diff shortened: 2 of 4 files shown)


WesCossick:
Once I receive feedback on the implementation itself, I can add documentation and tests!

src/exec.js
```diff
@@ -109,6 +110,7 @@ function execSync(cmd, opts, pipe) {
 function execAsync(cmd, opts, pipe, callback) {
   opts = common.extend({
     silent: common.config.silent,
+    fatal: common.config.fatal,
```
nfischer: This (and the line above it) seem redundant because we assign to this variable inside `_exec()`. I guess it's fine since `silent` left this, but could you leave a `TODO(nfischer)` to clean this up?

src/exec.js
```diff
@@ -28,6 +28,7 @@ function execSync(cmd, opts, pipe) {
 
   opts = common.extend({
     silent: common.config.silent,
+    fatal: common.config.fatal,
```
nfischer: This is also redundant (see comment below)

src/common.js
```diff
@@ -118,7 +118,7 @@ function error(msg, _code, options) {
   state.error += logEntry;
 
   // Throw an error, or log the entry
-  if (config.fatal) throw new Error(logEntry);
+  if (config.fatal && options.fatal !== false) throw new Error(logEntry);
```
nfischer: Instead of this, how about we assign `fatal: config.fatal` up above in `DEFAULT_OPTIONS`?

nfischer:
Here's a similar test you might be able to base this on:

https://github.com/shelljs/shelljs/blob/57df38c6ea264493a4219547331f2be049494ed5/test/config.js#L42-L48

src/common.js
WesCossick: Fixed it; not sure how I missed that pattern originally.

src/exec.js
WesCossick: Left a TODO comment for you on each of these lines. Let me know if you want it done another way.

nfischer: That TODO looks good, thanks

src/common.js
nfischer: You should update this to check `options.fatal` instead of `config.fatal` (otherwise, this should be OK)

WesCossick:
There is a very strange error where one of the new tests I added causes `common.js` to die mysteriously on [line 119](https://github.com/WesCossick/shelljs/blob/master/src/common.js#L119).

I drilled it down to that line by adding a `throw new Error('TEST');` line before and after that line and running `npm test` for each. When the error is thrown after line 119, it never gets thrown for that one test that's failing.

I have no idea why that is.

nfischer:
Not sure what the failure on 119 is (I don't see it on the bots), but https://travis-ci.org/shelljs/shelljs/jobs/572075940 failed here:

```
exec › options.fatal = true and unknown command
  /home/travis/build/shelljs/shelljs/test/exec.js:51
   50:   shell.config.fatal = false;                                         
   51:   t.throws(() => {                                                    
   52:     shell.exec('asdfasdf', { fatal: true }); // could not find command
```

This makes sense, because the implementation has a bug (per my previous comment, you need to check `options.fatal` rather than `config.fatal` in the code).

WesCossick:
The code does check `options.fatal`: https://github.com/WesCossick/shelljs/blob/master/src/common.js#L122

WesCossick:
When running `npm test`, the `options.fatal = true and unknown command` test dies at `common.js` precisely on line 119. No code after that point executes, and I have no idea why. But, just running the package normally, it's clear that what's really happening is that the `wrap()` function relies on checking `config.fatal` before deciding whether to throw an error ([this line](https://github.com/WesCossick/shelljs/blob/master/src/common.js#L428)).

How do you propose this should be addressed? There does not appear to be an easy way to wire the new `fatal` option through to the `wrap()` function without using the global config.

WesCossick:
I've implemented what I think is a decent solution, working within this project's existing design.

I've introduced a new `handlesFatalDynamically` option for the common `wrap()` function. Functions like `exec` can then use this boolean to indicate to the `wrap()` function that they suppress their own errors dynamically. This means that when they throw an error, it's because they want it to be seen, regardless of the global `fatal` value.

Let me know your thoughts, @nfischer.

src/common.js
```diff
@@ -88,7 +88,7 @@ CommandError.prototype = Object.create(Error.prototype);
 CommandError.prototype.constructor = CommandError;
 exports.CommandError = CommandError; // visible for testing
 
-// Shows error message. Throws if config.fatal is true
+// Shows error message. Throws if fatal is true (defaults to config.fatal, overridable with options.fatal)
```
nfischer: nit: "Throws if **options**.fatal is..."

src/exec.js
```diff
@@ -12,13 +12,15 @@ common.register('exec', _exec, {
...
   wrapOutput: false,
+  handlesFatalDynamically: true,
 });
 
 // We use this function to run `exec` synchronously while also providing realtime
 // output.
 function execSync(cmd, opts, pipe) {
   if (!common.config.execPath) {
-    common.error('Unable to find a path to the node binary. Please manually set config.execPath');
+    common.error('Unable to find a path to the node binary. Please manually set config.execPath', { continue: true, fatal: opts.fatal });
```
nfischer: Why `continue: true`?

src/exec.js
```diff
@@ -201,9 +205,15 @@ function _exec(command, options, callback) {
 
   options = common.extend({
     silent: common.config.silent,
+    fatal: common.config.fatal,
     async: false,
   }, options);
 
+  if (!command) {
+    common.error('must specify command', { continue: true, fatal: options.fatal });
+    return;
```
nfischer: It doesn't look right for `continue: true` to be followed by `return`...

test/exec.js
```diff
@@ -45,6 +45,15 @@ test('config.fatal and unknown command', t => {
   shell.config.fatal = oldFatal;
 });
 
+test('options.fatal = true and unknown command', t => {
+  const oldFatal = shell.config.fatal;
+  shell.config.fatal = false;
+  t.throws(() => {
+    shell.exec('asdfasdf', { fatal: true }); // could not find command
+  }, /asdfasdf/); // name of command should be in error message
```
nfischer: This isn't actually safe. If the assertion fails, we'll never reset `shell.config.fatal`. Can you file a TODO against me to fix this and the other spots? I might actually have a github issue filed for this somewhere...

test/exec.js
```diff
@@ -193,6 +202,15 @@ test('encoding option works', t => {
   t.is(result.stderr.toString(), '');
 });
 
+test('options.fatal = false and unknown command', t => {
+  const oldFatal = shell.config.fatal;
+  shell.config.fatal = true;
+  const result = shell.exec('asdfasdf', { fatal: false }); // could not find command
+  shell.config.fatal = oldFatal;
+  t.truthy(shell.error());
+  t.is(result.code > 0, true);
```
nfischer: `t.truthy(condition)`

src/common.js
```diff
@@ -450,6 +451,7 @@ var DEFAULT_WRAP_OPTIONS = {
   canReceivePipe: false,
   cmdOptions: null,
   globStart: 1,
+  handlesFatalDynamically: false,
```
nfischer: I need another look to wrap my head around this option. Not necessarily opposed to this, though

WesCossick:
@nfischer Any update on this?

src/common.js
nfischer: Ok, I see what's going on. This approach is fine, although I think the nicer long-term solution would be one of:

1. Permit an `options` parameter for all commands, and allow this to temporarily override `shell.config` (the next command will get the original copy of `shell.config`). Ex. `shell.cp('foo', 'bar', { fatal: true })`. I implemented this sort of parsing in #387, but for a different reason.
2. Reset `shell.config` after each command exits, unless it declares it mutates `shell.config` (we would need this option to support `shell.set('-e')`).

I was originally thinking this would look more like option 2 (set `shell.config.fatal = opts.fatal;` and let it get reset when the command exits), but that would require a bit more work to support `shell.set()`.

nfischer: Actually, for either of these we would need to make sure a snapshot of `shell.config` gets passed down to the closure returned by `wrap()`, to avoid race conditions if it's modified in the middle of an async method. But that seems feasible.

WesCossick: Agreed. Long term I think there's a better way to approach this problem that's more intuitive and reusable, but it'd probably require refactoring a fair bit of the existing functionality. Something to think about for the future!

PR #966, opened 2019-10-02
Cleanup LICENSE

reviewher:
Github does not currently recognize the license of this project.  It says "View License" at the top.  Other BSD-3-Clause projects like [freeCodeCamp](https://github.com/freeCodeCamp/freeCodeCamp) clearly say BSD-3-Clause in the top bar.

It seems that Github is very finicky and expects the exact formatting as well as body text.  The license was reshaped to match the formatting of other projects, so with this change GH will finally recognize that shelljs is BSD-3-clause

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,26 +1,29 @@
+BSD 3-Clause License
+
 Copyright (c) 2012, Artur Adib <arturadib@gmail.com>
 All rights reserved.
 
-You may use this project under the terms of the New BSD license as follows:
-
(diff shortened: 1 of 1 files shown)


nfischer:
The example you linked to is a markdown license; could you also link to a plain text license? Just want to double check since licenses are kind of a sensitive thing.

reviewher:
Better example: https://github.com/highlightjs/highlight.js/blob/master/LICENSE (same body text and formatting)

nfischer (approved):

PR #973, opened 2019-10-18
Silence potentially upcoming circular dependency warning

addaleax:
Node.js is currently considering printing a warning when a non-existent
property of `module.exports` is accessed while in a circular `require()`
dependency, in order to make it easier to catch issues with circular
dependencies.

In order to avoid printing these warnings for shelljs, checking for the
property’s existence rather than its truthiness suffices.

Refs: https://github.com/nodejs/node/pull/29935

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -478,7 +478,7 @@ function _register(name, implementation, wrapOptions) {
   // If an option isn't specified, use the default
   wrapOptions = Object.assign({}, DEFAULT_WRAP_OPTIONS, wrapOptions);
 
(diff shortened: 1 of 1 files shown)


addaleax:
Also: I’d understand if you don’t like this PR, because the previous code was perfectly fine.

I’m trying to implement this warning in Node.js because I believe that, in the big picture, it helps debug hard problems more than it annoy people – I’d guess that circular dependencies might have tripped you up in the past, too. If you have any feedback about the Node.js PR in general, please feel free to voice it there.

PR #982, opened 2019-12-23
fix(mv): Fix moving files across volumes

christopherthielen:
At some point options parsing happened inside `_mv()` and `_cp()` but that is no longer the case so a parsed options object should be used instead.


bangbang93:
Is there any plan to release?

PR #987, opened 2020-02-21
find() - use default platform path separators

cadon0:
I encountered an issue where [paths were assumed to be delimited with the current platforms path delimiter](https://github.com/banyudu/find-git-root/pull/5), and as such the output of shelljs's `find` - which currently returns paths with POSIX delimiters on all platforms - produced unexpected results.
The current `if (process.platform === 'win32') { file = file.replace(/\\/g, '/'); }` code seems to have been added as part of the initial implementation of `find` without a reason why.

I figure this may be a chance to review that behaviour.

I realise `/` is a suitable delimiter for all platforms, but I feel requiring an additional call to `path.normalize` for each path when using shelljs's `find` a bit odd. Especially considering the opening statement
> ShellJS is a portable (Windows/Linux/macOS) implementation of Unix shell commands on top of the Node.js API

It seems like expecting `/` delimiters when dealing with paths on Windows would be less intuitive than `\`, so those requiring such output would be prepared to intentionally check for & replace the delimiters as appropriate.

It would be in line with [Node.js](https://nodejs.org/api/path.html#path_path_sep) to return paths using the current platforms default delimiter.
>On Windows, both the forward slash (`/`) and backward slash (`\`) are accepted as path segment separators; however, the path methods only add backward slashes (`\`).

### DIFF
--- a/src/find.js
+++ b/src/find.js
@@ -30,13 +30,6 @@ function _find(options, paths) {
 
   var list = [];
 
-  function pushFile(file) {
-    if (process.platform === 'win32') {
-      file = file.replace(/\\/g, '/');
-    }
-    list.push(file);
-  }
-
   // why not simply do `ls('-R', paths)`? because the output wouldn't give the base dirs
   // to get the base dir in the output, we need instead `ls('-R', 'dir/*')` for every directory
 
@@ -48,11 +41,14 @@ function _find(options, paths) {
       common.error('no such file or directory: ' + file);
     }
 
-    pushFile(file);
+    // Node's path.join() below will return platform-appropriate paths,
+    // but on Windows paths with POSIX separators may be provided to this function.
+    // So normalize:
+    list.push(path.normalize(file));
 
     if (stat.isDirectory()) {
       _ls({ recursive: true, all: true }, file).forEach(function (subfile) {
-        pushFile(path.join(file, subfile));
+        list.push(path.join(file, subfile));
       });
     }
   });
--- a/test/find.js
+++ b/test/find.js
@@ -1,4 +1,5 @@
 import test from 'ava';
+import path from 'path';
 
 import shell from '..';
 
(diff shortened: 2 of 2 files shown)


levithomason:
# [Codecov](https://codecov.io/gh/shelljs/shelljs/pull/987?src=pr&el=h1) Report
> Merging [#987](https://codecov.io/gh/shelljs/shelljs/pull/987?src=pr&el=desc) into [master](https://codecov.io/gh/shelljs/shelljs/commit/a8452b89f2b9dd372da81e973a16074a5628b0b6&el=desc) will **decrease** coverage by `0.00%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/shelljs/shelljs/pull/987?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #987      +/-   ##
==========================================
- Coverage   97.22%   97.21%   -0.01%     
==========================================
  Files          35       35              
  Lines        1332     1329       -3     
==========================================
- Hits         1295     1292       -3     
  Misses         37       37              
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/shelljs/shelljs/pull/987?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/shelljs/shelljs/pull/987?src=pr&el=footer). Last update [a8452b8...903e755](https://codecov.io/gh/shelljs/shelljs/pull/987?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

nfischer:
> I realise `/` is a suitable delimiter for all platforms, but I feel requiring an additional call to `path.normalize` for each path when using shelljs's `find` a bit odd. Especially considering the opening statement

Why do you need `path.normalize()` if `/` is a valid delimiter?

> It seems like expecting `/` delimiters when dealing with paths on Windows would be less intuitive than `\`, so those requiring such output would be prepared to intentionally check for & replace the delimiters as appropriate.

I think this is designed for folks who _forget_ to check. In other words, `shell.find('**/exec.js')` should return an equivalent result on either Windows or Posix (forward slashes and all). As you said, most node APIs accept paths with `/` so this isn't usually a problem. This code is from before my time on the project, so I'm just guessing.

There's been talk in the past about enabling a setting to make this optional, to configure whether all command output gets converted to `/` or sticks with the platform path separator (`path.sep`). I think leaving output alone is probably the safer default, although I mostly just care about consistency between commands.

cadon0:
> Why do you need `path.normalize()` if `/` is a valid delimiter?

Right, you shouldn't need to. This point helps settle my thoughts - I feel now the root cause of my problem mainly lay in the other library, not this one as well.

It didn't seem like this would be worth a breaking change but appreciate the discussion. No further questions. Thanks! 🙂

PR #1006, opened 2020-08-28
Update ls with glob example.

smack0007:
Adding an example of using ls with globs. Will generate the README soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,6 +441,7 @@ Examples:
 
 ```javascript
 ls('projs/*.js');
+ls('projs/**/*.js'); // Find all js files recursively in projs
 ls('-R', '/users/me', '/tmp');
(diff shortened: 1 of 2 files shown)


smack0007:
I just added an example of using a glob pattern with ls. As a windows user this wasn't immediately obvious to me.

smack0007:
I took a look at the failed builds and I have no idea what to do. I only added a single comment to a function's description and regenerated the docs. Please advise.

nfischer:
It looks like you did everything right, but maybe the test has become flaky. Can you trigger a rebuild of the PR? I can't find the "rebuild" button in the Travis UI anymore.

The "brute force" way to trigger a rebuild is to `git commit --amend --no-edit && git push -f origin patch-2` (from your local branch).

nfischer (approved):

PR #1011, opened 2020-10-26
chore: change supported node versions

nfischer:
This adds testing for node v14. This removes testing for node v6 and v7
because codecov breaks on these versions. This omits node v15 because
appveyor doesn't seem to support this version. The nodejs org currently
supports [10, 12, 14, 15].

This makes a couple minor edits to the check-node-support script for
consistency with shelljs/shx#186.

Test: npm run check-node-support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,13 @@
 language: node_js
 sudo: false
 node_js:
-  - 6
-  - 7
   - 8
   - 9
   - 10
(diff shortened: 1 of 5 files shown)


nfischer:
Example run (before this PR) where codecov fails: https://ci.appveyor.com/project/shelljs/shelljs/builds/35951797

PR #1019, opened 2021-02-14
fix(mkdir): mitigate directory creation race condition

### LINKED ISSUE #1018: `mkdirSyncRecursive()` shouldn't fail when directory exists unexpectedly
When using `shx mkdir -p ...`, occasional, unnecessarily terminal, errors can occur if the target directory is created by another process during the execution of `mkdirSyncRecursive()`.

It's an obvious race condition that I've run into occasionally when using `shx` for builds with multiple parallel steps (eg, <https://github.com/rivy/js.xdg-portable/runs/1896221924?check_suite_focus=true>)

Ignoring 'EEXIST' errors in `mkdirSyncRecursive()` can mitigate this problem.

rivy:
Fixes #1018

### DIFF
--- a/src/mkdir.js
+++ b/src/mkdir.js
@@ -19,17 +19,18 @@ function mkdirSyncRecursive(dir) {
     common.error('dirname() failed: [' + dir + ']');
   }
 
-  // Base dir exists, no recursion necessary
-  if (fs.existsSync(baseDir)) {
-    fs.mkdirSync(dir, parseInt('0777', 8));
-    return;
+  // Base dir does not exist, go recursive
+  if (!fs.existsSync(baseDir)) {
+    mkdirSyncRecursive(baseDir);
   }
 
-  // Base dir does not exist, go recursive
-  mkdirSyncRecursive(baseDir);
-
-  // Base dir created, can create dir
-  fs.mkdirSync(dir, parseInt('0777', 8));
+  try {
+    // Base dir created, can create dir
+    fs.mkdirSync(dir, parseInt('0777', 8));
+  } catch (e) {
+    // swallow error if dir already exists
+    if (e.code !== 'EEXIST' || common.statNoFollowLinks(dir).isFile()) { throw e; }
+  }
 }
 
 //@


rivy:
Only previously created directories should be allowed/ignored, so I added an `isFile()` check to the catch logic.

rivy:
This doesn't touch 'test', so I think the single AppVeyor failure is unrelated (and likely a random "cosmic-ray" failure).

src/mkdir.js
```diff
@@ -28,8 +28,14 @@ function mkdirSyncRecursive(dir) {
   // Base dir does not exist, go recursive
   mkdirSyncRecursive(baseDir);
 
-  // Base dir created, can create dir
-  fs.mkdirSync(dir, parseInt('0777', 8));
+  try {
+    // Base dir created, can create dir
+    fs.mkdirSync(dir, parseInt('0777', 8));
```
nfischer, line 33: It looks like this same bug can happen above on line 24 (the other call to `fs.mkdirSync`). What do you think about refactoring this to recurse first (if necessary) and only call `fs.mkdirSync` in one spot?

```js
// Replace lines 22-39 with...

// Base dir does not exist, recursion necessary
if (!fs.existsSync(baseDir)) {
  mkdirSyncRecursive(baseDir);
}

try {
  // Base dir created, can create dir
  // etc. ...
}
```

nfischer (changes requested):

src/mkdir.js
rivy: Sure, I'll push a change as a "fixup! ..." commmit.
If it's corrent, tests ok, and you like it, I'll squash it into the initial commit prior to a merge.

rivy: Tests are passing, commits are squashed.

src/mkdir.js
```diff
@@ -19,17 +19,19 @@ function mkdirSyncRecursive(dir) {
...
+  }
 
-  // Base dir created, can create dir
-  fs.mkdirSync(dir, parseInt('0777', 8));
+  try {
+    // Base dir created, can create dir
+    fs.mkdirSync(dir, parseInt('0777', 8));
+  } catch (e) {
+    // swallow directory exists errors ('EEXIST'); mitigates race conditions
+    /* istanbul ignore next */
```
nfischer, line 32: I think we can remove `istanbul ignore next` because this now handles the non-racy condition where base dir exists.

You could also update the comment since this isn't only for race conditions (it's just the non-racy way to create a directory if it doesn't already exist).

nfischer (changes requested):
Looks good! I think we can remove the `istanbul ignore next` part, but this is great otherwise!

src/mkdir.js
rivy: Done.

nfischer: It looks like this caught some missing test coverage 😄  I think we weren't testing the case where `mkdir -p` is called on a directory which already exists. I filed issue #1022 to track this.

nfischer (approved):

Issue #1022, opened 2021-03-08
test coverage: mkdir -p over a directory which already exists


nfischer:
PR #1019 made me realize we're not testing the case where `mkdir -p` is used on a path which already exists. I think there are 2 interesting subcases here:

```sh
# Try making a folder which already exists
$ mkdir -p path/to/<folder which exists>

# Try making a folder which doesn't exist, but is a subdirectory of some folders which do exist
$ mkdir -p path/to/<folder which exists>/<some new folder which doesn't exist>
```

PR #1026, opened 2021-03-28
Added mkdir -p tests for subdirectories

JessieFrance:
This PR is for [#1022](https://github.com/shelljs/shelljs/issues/1022), and aims to expand testing for mkdir -p when (1) a sub-directory already exists, and (2) the sub-directory does not exist. Please let me know if you'd like me to change something, and thank you for the opportunity to work on this issue!

### DIFF
--- a/test/mkdir.js
+++ b/test/mkdir.js
@@ -165,6 +165,16 @@ test('-p flag: multiple dirs, array syntax', t => {
   t.truthy(fs.existsSync(`${t.context.tmp}/yyyc`));
 });
 
+test('-p flag: subdirectory already exists', t => {
+  t.falsy(fs.existsSync(`${t.context.tmp}/d1`));
+  shell.mkdir('-p', `${t.context.tmp}/d1/d2/d3`);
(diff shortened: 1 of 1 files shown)


nfischer (approved):

nfischer:
Hmm for some reason it looks like the coverage report isn't changing for `mkdir.js`: https://app.codecov.io/gh/shelljs/shelljs/compare/1026. Do you know why this condition [doesn't appear to be covered](https://codecov.io/gh/shelljs/shelljs/src/master/src/mkdir.js)?

https://github.com/shelljs/shelljs/blob/71085dba6527f86ff89ddbf62f07d8656190aa74/src/mkdir.js#L30-L33

---

It looks like appveyor is not reporting coverage info, as Windows-specific code doesn't appear to have coverage. This must be a codecov or appveyor bug though.

nfischer:
I see, thanks for looking into this. Let's land this as-is because it adds useful test coverage. If you can think of a way to de-duplicate the logic, that would be a welcome contribution in a separate PR!

Thank you!

nfischer (approved):

Issue #1013, opened 2021-01-12
exec not returning exit code properly


sethen:
### Node version (or tell us if you're using electron or some other framework):
v12.20.1
### ShellJS version (the most recent version/Github branch you see the bug on):
0.8.4
### Operating system:
Ubuntu 20.04 LTS
### Description of the bug:
`shell.exec` doesn't raise exit code if it is not zero.  
### Example ShellJS command to reproduce the error:

Here is the code I have:

```javascript
#! /usr/bin/env node

const shell = require("shelljs");

shell.exec("jest --config jest.config.json --coverage");
```

Here is what I had to do to make this work properly:

```javascript
#! /usr/bin/env node

const shell = require("shelljs");

shell.exec("jest --config jest.config.json --coverage", function (code) {
	if (code !== 0) {
		shell.exit(1);
	}
});
```

Shouldn't this be the default behavior?

nfischer:
If I understand correctly, you're trying to implement bash's behavior by exiting the script with the error code of the last shelljs command.

> Shouldn't this be the default behavior?

While I see some potential use cases for this, I think most use cases would prefer the way things are (no automatic manipulation of exit status). I think this could be a cool off-by-default feature, so this might be a good candidate for a [shelljs plugin](https://github.com/shelljs/shelljs/wiki/Using-ShellJS-Plugins).

We don't currently expose the error code of the last command, but you could get close to implementing this with something like:

```js
var shell = require('shelljs');

process.on('exit', code => {
  // Note: there's no shell.errorCode() API, so we just return 1 if
  // there's an error.
  var exitCode = shell.error() ? 1 : 0;
  shell.exit(exitCode);
});

// Your script here...
shell.echo('hello world');
```

If you were inclined, we would accept a PR to implement `shell.errorCode()`. This wouldn't be very difficult, and you could follow `shell.error()` as an example.

addlistener:
My use case here. I write bash as default, but later I need everything to work on Windows. So I tried to replace my bash lines with shelljs lines. That's why I need the exit feature.

nfischer:
Closing this one out. This was solved by https://github.com/shelljs/shelljs/pull/1036.

nfischer:
Suggested code (once shelljs v0.9.0 comes out) would look like:

```js
var shell = require('shelljs');

process.on('exit', code => {
  shell.exit(shell.errorCode());
});

// Your script here...
shell.echo('hello world');
```

[...]

PR #1036, opened 2021-06-24
Feature: Expose Error Code

JessieFrance:
This PR is for part of [#1013](https://github.com/shelljs/shelljs/issues/1013), and aims to implement ```shell.errorCode()```, similar to ```shell.error()```.  Please let me know if you'd like me to change something, and thank you for the chance to work on a new ShellJS feature!

### DIFF
--- a/test/grep.js
+++ b/test/grep.js
@@ -24,6 +24,7 @@ test('no args', t => {
   const result = shell.grep();
   t.truthy(shell.error());
   t.is(result.code, 2);
+  t.is(shell.errorCode(), 2);
 });
 
 test('too few args', t => {
--- a/README.md
+++ b/README.md
@@ -742,6 +742,11 @@ return value to be an error message. If you need the last error message, use
 the `.stderr` attribute from the last command's return value instead.
 
 
+### errorCode()
+
+Returns the error code from the last command.
+
+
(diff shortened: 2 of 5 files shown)


test/grep.js
```diff
@@ -24,6 +24,7 @@ test('no args', t => {
   const result = shell.grep();
   t.truthy(shell.error());
   t.is(result.code, 2);
+  t.is(shell.errorCode(), 2);
```
nfischer, line 27: Can you also add a test to check when errorCode is 0?

nfischer (changes requested):
Looks good! One small comment inline.

test/grep.js
JessieFrance: Thank you. Yes, I'll add that. Good thinking!

JessieFrance: Please let me know if the new test works. I'd also be happy to refactor the existing tests to use ```shell.errorCode()``` similar to how they use ```shell.error()``` instead of something like ```result.error```, but I know this would affect a lot of test files and wasn't sure if it would fit with this PR or with a new issue. In any case, thank you for your reviews, and I'm happy to make any changes, or add more tests.

nfischer (approved):

PR #1044, opened 2021-09-16
Grep directory support

### LINKED ISSUE #998: feat: grep -R/-r flag
This is a feature request to add the `-r` or `-R` flag to `grep()`. From the man page:

> ```
> -r, --recursive
>        Read  all  files under each directory, recursively, following symbolic links only if they are on the command line.  Note that if no file operand is given, grep searches the working directory.  This is equivalent
>        to the -d recurse option.
>
> -R, --dereference-recursive
>        Read all files under each directory, recursively.  Follow all symbolic links, unlike -r.
> ```

c-pranshu:
This Fixes #998 

- -r allows searching through directories in recursive way
- works with array syntax
- works with globs
- added tests

### DIFF
--- a/test/grep.js
+++ b/test/grep.js
@@ -56,6 +56,24 @@ test("multiple files, one doesn't exist, one doesn't match", t => {
   t.is(result.code, 2);
 });
 
+test('-r option not provided, but directory is present', t => {
+  t.truthy(fs.existsSync('test/resources'));
+  t.truthy('/test/resources/a.txt');
+  const result = shell.grep(/oogabooga/, 'test/resources', 'test/resources/a.txt');
+  t.truthy(shell.error());
+  t.is(result.stderr, 'grep: test/resources Is a directory');
+  t.is(result.code, 2);
+});
+
+test('atleast one directory or file does not exist', t => {
+  t.truthy(fs.existsSync('test/resources'));
+  t.falsy(fs.existsSync('test/resources/random.txt'));
+  const result = shell.grep('-r', /oogabooga/, 'test/resources', 'test/resources/random.txt');
+  t.truthy(shell.error());
+  t.is(result.code, 2);
+  t.is(result.stderr, 'grep: no such file or directory: test/resources/random.txt');
+});
+
 //
 // Valids
 //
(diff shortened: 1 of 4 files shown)


pranshu667:
Tests are passing on my local system but failing in ci.
Could someone please explain ?

c-pranshu:
Tests are passing on my local system,
but failing in ci build.
Could someone please help ?

c-pranshu:
continuous-integration/travis-ci status is not arriving.
i checked on the forum and there seems to be some problem in master branch.
could someone help ?

test/grep.js
```diff
@@ -162,6 +180,30 @@ test('-i option', t => {
...
+test('directory provided in glob form', t => {
+  const result = shell.grep('-r', 'test', 'test/r*');
+  t.falsy(shell.error());
+  t.is(result.split('\n').length, 66);
+});
+
+test('works with array syntax', t => {
+  const result = shell.grep('-r', 'test', ['test/r*', 'package.json'], 'scripts');
+  t.falsy(shell.error());
+  t.is(result.split('\n').length, 72);
```
nfischer, line 204: Is this the right value? I get a different answer in my shell:

```
$ grep -r 'test' test/r* package.json scripts | wc -l
69
```

test/grep.js
```diff
@@ -162,6 +180,30 @@ test('-i option', t => {
...
+});
+
+test('directory provided in glob form', t => {
+  const result = shell.grep('-r', 'test', 'test/r*');
+  t.falsy(shell.error());
+  t.is(result.split('\n').length, 66);
+});
+
+test('works with array syntax', t => {
+  const result = shell.grep('-r', 'test', ['test/r*', 'package.json'], 'scripts');
```
nfischer, line 202: Please do not include anything outside of the resources/ folder in the test (package.json, scripts, etc.). It makes the test much more difficult to maintain if we do unrelated changes in package.json etc.

I think the same goes for `test/r*` because that matches `test/rm.js`. Instead of this, how about adding something like `test/resources/grep2` and testing the glob `test/resources/g*` if you want to test globbing?

test/grep.js
```diff
@@ -56,6 +56,24 @@ test("multiple files, one doesn't exist, one doesn't match", t => {
...
+  t.is(result.code, 2);
+});
+
+test('atleast one directory or file does not exist', t => {
+  t.truthy(fs.existsSync('test/resources'));
+  t.falsy(fs.existsSync('test/resources/random.txt'));
+  const result = shell.grep('-r', /oogabooga/, 'test/resources', 'test/resources/random.txt');
+  t.truthy(shell.error());
+  t.is(result.code, 2);
+  t.is(result.stderr, 'grep: no such file or directory: test/resources/random.txt');
```
nfischer, line 74: I think we also need to check `result.stdout`. My `grep` will return results for the folders which actually do exist.

test/grep.js
```diff
@@ -162,6 +180,30 @@ test('-i option', t => {
   t.is(result.split('\n').length - 1, 3);
 });
 
+test('one directory provided with -r option', t => {
+  const result = shell.grep('-r', 'test', 'test/resources');
+  t.falsy(shell.error());
+  t.is(result.split('\n').length, 29);
+});
+
+test('multiple directories or files provided', t => {
```
nfischer, line 189: We should probably split this into two separate test cases for clarity.

1. We want a test case where multiple directories/files are provided, but they do not overlap. ex. `"test/", "different-folder/file.txt"`
2. We want a test case (like this one) where there is overlap. We should assert that the overlapping files are double-counted (because that's what POSIX grep does). ex. `"test/", "test/file.txt"` (file.txt should be `grep`ed twice).

nfischer (changes requested):
This is cool! Thanks for writing this up. I left some initial comments. I haven't double-checked if the behavior is right though (I spot checked a couple spots, but I got different results than you in one of them), so you may want to look a bit closer there.

Either way, this seems like a good starting point for a valuable feature.

Issue #739, opened 2017-06-17
Improve coverage for cp()


nfischer:
See [the coverage report](https://codecov.io/gh/shelljs/shelljs/src/c16fb7dac9dfec6cec29cfe633daece83a127d47/src/cp.js)

We're currently missing line coverage for:

 - `copyFileSync()` being called on a [non-existent file](https://github.com/shelljs/shelljs/blob/master/src/cp.js#L23)
 - [Cycles of symlinks](https://github.com/shelljs/shelljs/blob/master/src/cp.js#L119-L123)
 - `srcFileStat.isSymbolicLink() && srcFileStat.isDirectory()` (logic is [here](https://github.com/shelljs/shelljs/blob/master/src/cp.js#L138) and [here](https://github.com/shelljs/shelljs/blob/master/src/cp.js#L140)) -- probably a bug
 - calling `cpcheckcycle()` [on a directory](https://github.com/shelljs/shelljs/blob/master/src/cp.js#L174) (`fs.lstatSync(srcFile).isSymbolicLink() && fs.statSync(srcFile).isDirectory()`)

Blocks #671

PR #1045, opened 2021-09-23
[cp] Add test for cycle symlink

wafuwafu13:
Related: https://github.com/shelljs/shelljs/issues/739


test/cp.js
```diff
@@ -621,26 +621,20 @@ test('Test max depth.', t => {
   // Check last directory to exist is below maxdepth.
   t.truthy(shell.test('-d', `${t.context.tmp}/copytestdepth${directory32deep}`));
   t.falsy(shell.test('-d', `${t.context.tmp}/copytestdepth${directory32deep}/32`));
-  utils.skipOnWinForEPERM(shell.ln.bind(shell, '-s', `${t.context.tmp}/0`, `${t.context.tmp}/symlinktest`), () => {
```
wafuwafu13, line 624: Delete it because it doesn't create symlinks to check for cycle.

wafuwafu13:
test passes in the local environment... :sweat:

nfischer:
Could you rebase? I recently updated CI to use GitHub actions instead. This may be more reliable than the old environment.

wafuwafu13:
I don't know how to solve this problem in windows, so close

nfischer:
Thanks for trying the rebase. Looks like the error message from https://github.com/shelljs/shelljs/runs/5018531671?check_suite_focus=true is:

```
cp » Test with cycle symlink

  D:\a\shelljs\shelljs\tmp04812076144037267068[653](https://github.com/shelljs/shelljs/runs/5018531671?check_suite_focus=true#step:5:653)06818777819\test\cp.js:635

  Value is not falsy:

  'cp: cannot create directory \'new\': No such file or directory'

  _utils.default.skipOnWinForEPERM (test/cp.js:635:9)
  Object.skipOnWinForEPERM (test/utils/utils.js:31:5)
  t (test/cp.js:629:11)
```

I wonder if `new` is a reserved word on Windows, or maybe we just can't create the dir for some reason?

wafuwafu13:
`'cp: cannot create directory \'newsub\': No such file or directory'` 🤔

PR #1080, opened 2022-02-23
Added -L to find to visit symlinked folders too.

mperrando:
Actually find does not go inside a symlink that is a folder.

In order to preserve old behaviour, the -L flag has been introduced to modify the behaviour and enter symlink'ed folders.


test/find.js
```diff
@@ -40,7 +40,7 @@ test('simple path', t => {
   t.is(result.code, 0);
   t.truthy(result.indexOf('test/resources/find/.hidden') > -1);
   t.truthy(result.indexOf('test/resources/find/dir1/dir11/a_dir11') > -1);
-  t.is(result.length, 11);
+  t.is(result.length, 12);
```
nfischer, line 43: Why is this increased?

nfischer (commented):
Thanks! This seems reasonable overall

test/find.js
mperrando: Because there is actually a new entry: the link to the dir2 folder.
Unfortunately git did not add it in the first commit, now I pushed a new commit that adds it.

Ashry00: 264470c91100948905108f19995bb0b193fee982

nfischer: Ah, I see it now. Looks good!

nfischer (approved):

PR #1119, opened 2023-03-10
Explicitly require commands

Everspace:
Currently anything that includes shelljs in it's chain cannot be bundled into a singular file due to the dynamic require. By explicitly requiring everything in src, this allows singular bundles through things like esbuild.

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -21,9 +21,40 @@ var common = require('./src/common');
 //@commands
 
 // Load all default commands
-require('./commands').forEach(function (command) {
-  require('./src/' + command);
-});
+require('./src/cat');
+require('./src/cd');
+require('./src/chmod');
+require('./src/cmd');
+require('./src/common');
+require('./src/cp');
+require('./src/dirs');
+require('./src/echo');
+require('./src/error');
+require('./src/errorCode');
+require('./src/exec-child');
+require('./src/exec');
+require('./src/find');
+require('./src/grep');
+require('./src/head');
+require('./src/ln');
+require('./src/ls');
+require('./src/mkdir');
+require('./src/mv');
+require('./src/popd');
+require('./src/pushd');
+require('./src/pwd');
+require('./src/rm');
+require('./src/sed');
+require('./src/set');
+require('./src/sort');
+require('./src/tail');
+require('./src/tempdir');
+require('./src/test');
+require('./src/to');
+require('./src/toEnd');
+require('./src/touch');
+require('./src/unique');
+require('./src/which');
 
 //@
 //@ ### exit(code)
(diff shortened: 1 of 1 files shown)


TimothyJones:
I'd love to see this merged - is there a reason not to?

7PH:
I suggested another solution to this issue [in here](https://github.com/shelljs/shelljs/issues/1160), but I've only tested it with `esbuild`, it would be interesting to know if it also fixes the other bundlers

Everspace:
It would apply to every bundler @7PH

7PH:
Later, I've tested [more bundlers](https://github.com/shelljs/shelljs/pull/1161#issuecomment-2170539764) and some already work, some don't with this project's `require`. But [my PR](https://github.com/shelljs/shelljs/pull/1161) isn't getting a lot of attention by the owner of `shelljs`.

AlexanderLill:
Dear maintainers @nfischer @freitagbr, please consider merging this minor change to make shelljs compatible with esbuild and other bundlers (see previous comments and many of the open issues, e.g. #1160 #1172 etc).

shell.js
```diff
@@ -21,9 +21,40 @@ var common = require('./src/common');
...
+require('./src/cd');
+require('./src/chmod');
+require('./src/cmd');
+require('./src/common');
+require('./src/cp');
+require('./src/dirs');
+require('./src/echo');
+require('./src/error');
+require('./src/errorCode');
+// require('./src/exec-child'); excluded since it is for commandline only
```
nfischer, line 34: Can we remove this line? This is just a code comment so it doesn't do anything in regular node. Is this commented-out code necessary in order to trick the bundler?

Everspace:
No, I thought it was nice for the sake of completeness of all the files in
./src, but I can remove it.

shell.js
nfischer: I think you should change this to just actually `require()` the exec-child.js file. If you rebase on the lastest master branch, then I have changed this to be a NOOP. I think requiring is necessary to trick the bundler into realizing this file is necessary.

nfischer (changes requested):
Can you please do the following:

* Rebase on top of latest master branch, including PR #1187
* Remove the `commands.js` file
* Edit `package.json` to remove the reference to `commands.js`
* Please confirm if this PR will fix either #1160 and #1172

nfischer (approved):

nfischer:
I'm going to approve now and land this. I'll send the necessary patches for followup fixes.

Thanks for your contribution!

Everspace:
Ack I am so sorry, I've been busy. Thank you very much for your help and getting this merged in.

PR #1122, opened 2023-06-26
fix: shell.errorCode() honors shell.exit(code)

nfischer:
This changes shell.exit() to use a wrapper function. This is so that shell.errorCode() will have the correct error code after invoking shell.exit(code). This isn't normally relevant, however a caller may be listening for the exit status:

```js
process.on('exit', code => {
  shell.exit(shell.errorCode());
});
```

Issue #1013

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -29,7 +29,20 @@ require('./commands').forEach(function (command) {
 //@ ### exit(code)
 //@
 //@ Exits the current process with the given exit `code`.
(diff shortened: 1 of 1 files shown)


nfischer:
Also fixed `shell.exit()` to update `shell.error()` with a truthy value, which means you could also write code like this:

```js
process.on('exit', code => {
  shell.exit(shell.error() ? 1 : 0);
});
```

PR #1129, opened 2023-08-23
Added verbosity flag (-v) for cp, mv and rm commands

### LINKED ISSUE #1127: Feature request (cp): -v flag for verbose output
Original `cp` command in Linux has `-v` flag, it prints out source and destination for every copied file line-by-line, e. g.:

```
'foo/some-file.txt' -> 'bar/foo/some-file.txt'
```

It would be very useful to have this feature in the package, so that users could have more clue of what's going on when something is being copied.

nicky1038:
Closes #1127

The suggested verbose output is **not** claimed to be fully identical to the corresponding output in the original commands from coreutils. Though it is very similar, I also tried to make as few edits as possible.

### DIFF
--- a/src/cp.js
+++ b/src/cp.js
@@ -12,6 +12,7 @@ common.register('cp', _cp, {
     'L': 'followsymlink',
     'P': 'noFollowsymlink',
     'p': 'preserve',
+    'v': 'verbose',
   },
   wrapOutput: false,
 });
@@ -84,6 +85,10 @@ function copyFileSync(srcFile, destFile, options) {
     fs.closeSync(fdr);
     fs.closeSync(fdw);
   }
+
+  if (options.verbose) {
+    console.log("copied '" + srcFile + "' -> '" + destFile + "'");
+  }
 }
 
 // Recursively copies 'sourceDir' into 'destDir'
@@ -108,6 +113,9 @@ function cpdirSyncRecursive(sourceDir, destDir, currentDepth, opts) {
   var checkDir = common.statFollowLinks(sourceDir);
   try {
     fs.mkdirSync(destDir);
+    if (opts.verbose) {
+      console.log("created directory '" + sourceDir + "' -> '" + destDir + "'");
+    }
   } catch (e) {
     // if the directory already exists, that's okay
     if (e.code !== 'EEXIST') throw e;
--- a/src/mv.js
+++ b/src/mv.js
@@ -8,6 +8,7 @@ common.register('mv', _mv, {
   cmdOptions: {
     'f': '!no_force',
     'n': 'no_force',
+    'v': 'verbose',
   },
 });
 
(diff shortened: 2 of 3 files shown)


nicky1038:
I had to force-push in order to sign the commit, so that GitLab could unblock the PR. No code changed

src/rm.js
```diff
@@ -106,43 +116,53 @@ function handleFile(file, options) {
...
     common.error('path is a directory', { continue: true });
   }
 }
 
 function handleSymbolicLink(file, options) {
+  function _unlink() {
+    common.unlinkSync(file);
+    if (options.verbose) {
+      console.log("removed '" + file + "'");
+    }
```
nfischer, line 130: Can you change this so that `file` is a method parameter? I'd also suggest passing `options` as a second parameter and then moving this function definition to the global scope, similar to what you've done with `handleFIFO()`.

src/mv.js
```diff
@@ -102,15 +103,18 @@ function _mv(options, sources, dest) {
...
       /* istanbul ignore next */
       if (e.code === 'EXDEV') {
         // If we're trying to `mv` to an external partition, we'll actually need
         // to perform a copy and then clean up the original file. If either the
         // copy or the rm fails with an exception, we should allow this
         // exception to pass up to the top level.
-        cp({ recursive: true }, src, thisDest);
-        rm({ recursive: true, force: true }, src);
+        cp({ recursive: true, verbose: options.verbose }, src, thisDest);
+        rm({ recursive: true, force: true, verbose: options.verbose }, src);
```
nfischer, line 117: Good catch! :smile:

src/cp.js
```diff
@@ -84,6 +85,10 @@ function copyFileSync(srcFile, destFile, options) {
     fs.closeSync(fdr);
     fs.closeSync(fdw);
   }
+
+  if (options.verbose) {
+    console.log("copied '" + srcFile + "' -> '" + destFile + "'");
```
nfischer, line 90: I wonder if we should drop the "copied" prefix. When I tested this on my linux machine, I don't see this in the output:

```shell
$ cp -vr src/ dest/
'src/' -> 'dest/'
'src/file.txt' -> 'dest/file.txt'
'src/sub1' -> 'dest/sub1'
'src/sub1/sub2' -> 'dest/sub1/sub2'
'src/sub1/sub2/file2.txt' -> 'dest/sub1/sub2/file2.txt'
```

On the other hand, keeping "copied" would be more consistent with the other commands, since those do say "removed" or "renamed" before the file name.

What do you think?

src/rm.js
```diff
@@ -59,6 +63,9 @@ function rmdirSyncRecursive(dir, force, fromSymlink) {
       try {
         result = fs.rmdirSync(dir);
         if (fs.existsSync(dir)) throw { code: 'EAGAIN' };
+        if (verbose) {
+          console.log("removed directory'" + dir + "'");
```
nfischer, line 67: Add a space between `directory` and `'`

nfischer (changes requested):

nfischer:
It looks like there's a CI bug on the `main` branch (not a problem with your PR). I opened issue #1130 to work on that.

PR #1135, opened 2023-09-16
Exports shell.js and make.js on package.json

alexojegu:
Exports shell.js and make.js on package.json so make.js can be imported as "shelljs/make" without require file extension when using ESM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
   "main": "./shell.js",
   "exports": {
     ".": "./shell.js",
-    "./make": "./make.js"
+    "./make": "./make.js",
+    "./make.js": "./make.js"
   },
   "files": [
     "commands.js",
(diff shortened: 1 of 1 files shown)


nfischer:
Can you link me to documentation on the `exports` attribute of `package.json`? I'm not familiar with this. Sounds like it might be specific to ESM modules, but I'd like to make sure I understand before merging.

alexojegu:
The related section of the official documentation: https://nodejs.org/docs/latest-v20.x/api/packages.html#package-entry-points

nfischer (approved):

nfischer:
That link suggests this applies to both commonjs and ESM, but your PR mentions you're running into problems with ESM specifically. Do I understand correctly that ESM doesn't respect the `main` attribute, but does respect `exports`? Or does ESM just struggle with the `shelljs/make` syntax (as opposed to `shelljs/make.js`)?

shelljs/make isn't officially supported anymore, but I'm OK with adding this as an export since that's fairly low effort.

nfischer:
Do you want to add a line for

> ```
> "./make.js": "./make.js"
> ```

or should I merge as-is?

PR #1151, opened 2024-02-18
chore: try codecov token again

nfischer:
I tried this recently in 1a1a820569f0883a317944973c15cf34431dd6c2 and then reverted in b8b1c4224d0ec7d631d26c78922c018dd5b895f5. Unfortunately I'm still seeing spurious failures from codecov, so let's try the token one more time.

This time I double checked the token is provided by https://app.codecov.io/gh/shelljs/shelljs/settings. I regenerated the token for good measure.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,4 +51,5 @@ jobs:
         if: matrix.node-version >= 10
         uses: codecov/codecov-action@v3
         with:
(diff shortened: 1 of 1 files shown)


nfischer:
I verified that https://github.com/shelljs/shelljs/pull/1150 is being diffed against the right comment for coverage purposes. The real test will be to land 1 more change and see if the diff base stays up-to-date.

Issue #828, opened 2018-02-22
Consider switching to fast-glob for performance


nfischer:
This is a potential optimization.

I remember we switched to node-glob a while back (I think in v0.6) and got a bit of a performance hit as a result (but with significant gains in correctness/consistency). We might consider switching to [fast-glob](https://github.com/mrmlnc/fast-glob) or one of the related packages to see if we can get performance gains.

This should be an easy code change, and I can help run performance benchmarks. Also, feel free to take a peek at [shelljs/benchmarks](https://github.com/shelljs/benchmarks).

nfischer:
Another related package: [sindresorhus/globby](https://github.com/sindresorhus/globby) (fast-glob under the hood, with a nicer API and good support for async-await down the road).

pavelloz:
Looks like there are even better numbers to gain: https://github.com/terkelg/tiny-glob

pavelloz:
After some more time and looking around, it looks like it is very popular topic to optimize, those caught my eye especially:

https://www.npmjs.com/package/micromatch
https://www.npmjs.com/package/nanomatch

My use case is very glob-intentive thats why im very interested in this.
I can try to do it and make PR if there is a chance of it getting merged.

pavelloz:
I tested `fast-glob` - it looks like there are differences, meaning, nested bash brackets are not supported. This is not working compared to glob:

`modules/**/{private,public}/**/*.liquid` - if i remove `{private,public}` results are the same.

Other than that, i must say, results are pretty good.

## glob
`15.13s user 5.03s system 100% cpu 20.007 total`

## fast-glob
`1.05s user 0.82s system 104% cpu 1.797 total`

nfischer:
> This is not working compared to glob

Any way to get that working? I think fast-glob documents support for that pattern.

pavelloz:
I will try to prepare test case to be sure, maybe i did something wrong (i tested in my app).

API looks the same, having the same feature set while providing this kind of performance improvement would be huge.

775092252:
Git @nfischer

RomainLanz:
Hey there!

Do you know when a release including these changes might be available? 👀 👀

RandyVar:
Hello,

Are there any plans to release a new version of ShellJS that addresses the security issues reported by SNYK? The issues were introduced through shelljs@0.8.5.

Introduced through: scripts@1.0.0 → shelljs@0.8.5 → glob@7.2.3 → inflight@1.0.6

[...]

PR #1152, opened 2024-02-18
deprecate config.globOptions

nfischer:
This removes support for configuring `config.globOptions`. Exposing this variable makes it difficult to change (or upgrade) our glob library. It's best to consider our choice of glob library to be an implementation detail.

As far as I know, this is not a commonly used option: https://github.com/shelljs/shelljs/issues?q=globOptions currently shows no GitHub issues of users using this option, and there was never really a motivation for why this needed to be exposed (#400 exposed the option just because we could, not because we should).

This is one step toward supporting Issue #828.

### DIFF
--- a/README.md
+++ b/README.md
@@ -854,7 +854,11 @@ rm -rf foo.txt bar.txt
 exec echo hello
 ```
 
-### config.globOptions
+### config.globOptions (deprecated)
+
+**Deprecated**: we recommend that you do not edit `config.globOptions`.
+Support for this configuration option may be changed or removed in a future
+ShellJS release.
 
 Example:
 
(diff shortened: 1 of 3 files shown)


nfischer:
Let me see if it is possible to deprecate this but keep maintaining it for the time being (e.g., add the test cases back). I need to see if that approach will be compatible with PR #1153.

nfischer:
It does look like there's some usage of this option when I search all of GitHub: https://github.com/search?q=shell.config.globOptions&type=code

So I think the plan should be: deprecate official support, but don't break existing usage. I'll remove the "breaking" label since this technically is not a breaking change anymore.

Issue #1160, opened 2024-06-04
Dynamic require() without .js extension breaks bundling (esbuild)


7PH:
Node `v20`
ShellJS `0.8.5`

## Related issues
- https://github.com/shelljs/shelljs/issues/962
- https://github.com/shelljs/shelljs/issues/1047
- https://github.com/shelljs/shelljs/pull/1119
- https://github.com/kubernetes-client/javascript/issues/944
- https://github.com/aws/aws-toolkit-azure-devops/pull/539
- https://github.com/aws/aws-toolkit-azure-devops/pull/537
- https://github.com/microsoft/azure-pipelines-task-lib/issues/942

## Description of the bug

This library can't be bundled with esbuild

## Cause

https://github.com/shelljs/shelljs/blob/3529d5095f5331d126e8b0c0dbc4812f9ae1a34f/shell.js#L25

The combination of dynamic calls to `require()` and omiting `.js` extension in `require()` breaks bundlers. Bundlers can't know whether dynamic imports will include or not the file extension, so some of them (eg esbuild) assume it is included.

If we open the generated bundle, we can see that all required modules are there (`src/cat.js`, etc) but referenced with their extensions (eg `src/cat.js @ L4684`)
![image](https://github.com/shelljs/shelljs/assets/31401273/73056050-1266-4f72-a90a-48465358af41)

The dynamic `require()` is included as-is in the bundle (`@ L4762`):
![image](https://github.com/shelljs/shelljs/assets/31401273/99b5f785-5dee-413c-aec4-816a71cb6a3b)

But this fails at runtime with
```text
Error: Module not found in bundle: ./src/cat
```

Because `./src/cat` does not exist, `./src/cat.js` does.

## How to fix

To fix this, `shell.js @ L25` can be modified to add the `.js` file extension. For consistency, all other calls to require could also use the `.js` extension in calls to require.

I do not know what solution you prefer @nfischer . Happy to contribute on this if you give me pointers to how you think it should be done.

## Workaround

I've made [an esbuild plugin](https://github.com/7PH/esbuild-plugin-shelljs) to patch `shelljs` to unblock me as soon as now:
- https://github.com/7PH/esbuild-plugin-shelljs

nfischer:
Can you try this out in the latest ShellJS release (v0.9.1) and let me know if that fixed the problem?

PR #1161, opened 2024-06-08
Fix shelljs compatibility with esbuild bundler

7PH:
## Summary

Fixes https://github.com/shelljs/shelljs/issues/1160
- Add explicit file extension to dynamic command import.
- This call to require() is dynamic and does not add the file extension.
- Esbuild resolves files with their explicit file extension (eg './src/cat.js').
- Esbuild automatically adds .js extension in imports when not present but can't modify dynamic imports. For dynamic imports, file extension needs to be explicitly specified prior to bundling.

## Test

Here is a script to reproduce (to be run in the repository root):
```bash
# Prepare
rm -rf test-esbuild && mkdir test-esbuild && cd test-esbuild
npm init --yes
echo "console.log(require('../shell.js').pwd().stdout && '\n OK');" > index.js
npm i -D esbuild

# Run
npx esbuild --platform=node --bundle index.js --outfile=out.js; node out.js
```

In the current master branch of this repository:
```text
  out.js  230.5kb

⚡ Done in 10ms
/.../shelljs/test-esbuild/out.js:5
  throw new Error("Module not found in bundle: " + path);
  ^

Error: Module not found in bundle: ./src/cat
    at /.../

Node.js v20.3.1
```

In this branch
```text
  out.js  230.5kb

⚡ Done in 9ms

 OK
```

The bundle size doesn't change because all deps are already in the bundle, it's a resolution issue, as discussed in https://github.com/shelljs/shelljs/issues/1160

### DIFF
--- a/shell.js
+++ b/shell.js
@@ -22,7 +22,7 @@ var common = require('./src/common');
 
 // Load all default commands
 require('./commands').forEach(function (command) {
-  require('./src/' + command);
+  require('./src/' + command + '.js');
 });
 
 //@


nfischer:
Have you tested this change on other bundlers? I'm not familiar with the current state of bundlers in the community. Is esbuild a popular one?

7PH:
> Have you tested this change on other bundlers?

I've tested:
- [esbuild](https://github.com/evanw/esbuild) (37.5K stars, 31M weekly downloads)
- [rollup](https://github.com/rollup/rollup) (24.9K stars, 22.3M weekly downloads)
- [@vercel/ncc](https://github.com/vercel/ncc) (9K stars, 0.4M weekly downloads)

## rollup

With `rollup` it doesn't work, but you can tweak options so that these imports are treated dynamically ([dynamicRequireTargets](https://github.com/rollup/plugins/blob/master/packages/commonjs/README.md#dynamicrequiretargets)). 

<details>
<summary> Reproducer</summary>

```js
// rollup.config.js
var commonjs = require('@rollup/plugin-commonjs');
var nodeResolve = require('@rollup/plugin-node-resolve').nodeResolve;

module.exports = {
  input: 'index.js',
  output: {
    file: 'out.js',
    format: 'cjs',
  },
  plugins: [commonjs({
    dynamicRequireTargets: [
      'node_modules/shelljs/src/*.js',
    ],
  }), nodeResolve()],
};

// index.js

console.log(shelljs.pwd().stdout && '\n OK');

// bash
//$ npx rollup -c rollup.config.js
```

image
</details>

## ncc

With `ncc` it works out of the box because `ncc` assumes the require could be called without the .js extension as well 

<details>
<summary> Reproducer</summary>

image
</details>

## esbuild

See https://github.com/shelljs/shelljs/issues/1160

7PH:
Hello @nfischer , is there anything I can do more to help make this change go through?

jonsherrard:
Would also love to get this in.shelljs is very useful for github actions

AlexanderLill:
I am also highly interested in this, shelljs would be a lot more useful if it could be bundled! @nfischer

AlexanderLill:
I tried the fix proposed here, but it did not work in my setup with esbuild. However, just manually writing the `require` statements works for me (even though it is not so elegant):

https://github.com/AlexanderLill/shelljs/compare/master...esbuild-fix

qarlosalberto:
I have the same issue. It's very common to use `esbuild` to bundle VSCode extensions. Is there any plan to merge it? Thanks!

nfischer:
I'm willing to revisit this, however I need someone to show me how to test this change to make sure it actually works. I also see #1119 was raised for a similar issue, so I'm not sure which is the better approach.

7PH:
Hello @nfischer , thanks for answering!

>I'm not sure which is the better approach.

As a general rule of thumb, I say dynamic imports is something to avoid because that will always mess with build-time tree shaking. Bundlers can cope with this but this relies on blindly putting all the file content in the bundle.

TBH [moving to static imports](https://github.com/shelljs/shelljs/pull/1119) would be the preferred approach if it was my project, but I also think my change is fine if you want to keep the dynamic import. 

>however I need someone to show me how to test this change to make sure it actually works

For testing, I've put there a reproducer for ESBuild, do you want a reproducer for other bundlers?

nfischer:
Closing this in favor of #1119 and #1194

PR #1153, opened 2024-02-18
refactor: switch to fast-glob

nfischer:
This removes `node-glob` in favor of `fast-glob`. The main motivation
for this is because `node-glob` has a security warning and I can't
update to `node-glob@9` unless we drop compatibility for node v8.

Switching to `fast-glob` seems to be fairly straightforward, although
some options need to be changed by default for bash compatibility.

Fixes #828
Fixes #1149
Fixes #1148

PR #1165, opened 2024-06-19
Fix Windows test errors in src/ls.js and test/cp.js.

kmashint:
These are fixes for Windows in the switch-fast-blob branch as mentioned in this pull-request.
https://github.com/shelljs/shelljs/pull/1153

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   },
   "dependencies": {
     "execa": "^1.0.0",
-    "glob": "^7.0.0",
+    "fast-glob": "^3.3.2",
     "interpret": "^1.0.0",
     "rechoir": "^0.6.2"
   },
--- a/src/common.js
+++ b/src/common.js
@@ -19,7 +19,6 @@ var isElectron = Boolean(process.versions.electron);
 // Module globals (assume no execPath by default)
 var DEFAULT_CONFIG = {
   fatal: false,
-  globOptions: {},
   maxdepth: 255,
   noglob: false,
   silent: false,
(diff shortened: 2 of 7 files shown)


nfischer (approved):

nfischer:
@kmashint can you please rebase this PR on top of the `master` branch instead? I realized this change is missing some critical commits that I landed recently on `master`. I'm hoping that your fixes are backward compatible with both `glob` and `fast-glob` so this should be relatively easy to land.

kmashint:
@nfischer Will rebase on master tonight.

kmashint:
I merged master into this pull-request branch, and hoping I cut it the right way, removing references to glob in favour of fast-glob.

src/common.js
```diff
@@ -246,12 +246,6 @@ function parseOptions(opt, map, errorOptions) {
 }
 exports.parseOptions = parseOptions;
 
-function globOptions() {
```
nfischer, line 250: I think this was reverted when it should not have been. I think what happened is that you probably have an out-of-date copy of the `master` branch. Can you please sync this against the main repo's master branch? I think you can do this with something like:

```
git checkout master
git remote add upstream https://github.com/shelljs/shelljs.git
git fetch upstream
git rebase upstream/master
git checkout switch-fast-glob
git rebase master
git push origin switch-fast-glob --force-with-lease
```

nfischer (changes requested):

package.json
```diff
@@ -55,7 +55,7 @@
   },
   "dependencies": {
     "execa": "^1.0.0",
-    "glob": "^7.0.0",
+    "fast-glob": "^3.3.2",
```
nfischer, line 58: Also, I would like to exclude this part from this PR. My change will take care of switching over to `fast-glob`. I'd like to merge your change before mine though to make sure it is backward-compatible with "plain" `glob. Then I'll rebase my change on top of yours to test it with `fast-glob`.

It's also fine if you want to test this with `fast-glob` locally in your checkout. We should just remove this part before merging this on the `master` branch.

nfischer:
Closing this one in favor of #1166.

PR #1173, opened 2024-08-13
Update dependencies

mapachurro:
It looks like shelljs is still getting _lots_ of usage, even though several of the core dependencies are _very_ out of date.
image

Output from `npm outdated`:
image

I'm not sure if this initial bump works -- I need to test it -- but I thought I'd get the ball rolling with a PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,24 +54,24 @@
     "shjs": "./bin/shjs"
   },
   "dependencies": {
-    "execa": "^1.0.0",
-    "glob": "^7.0.0",
-    "interpret": "^1.0.0",
-    "rechoir": "^0.6.2"
+    "execa": "^9.3.0",
+    "glob": "^11.0.0",
+    "interpret": "^3.1.1",
+    "rechoir": "^0.8.0"
   },
   "ava": {
     "serial": true,
     "powerAssert": false
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


nfischer:
We intentionally use old versions of some projects because the old versions have better nodejs compatibility. I'm generally happy to accept package updates one-by-one IF those updates don't break old node versions, however I run this process occasionally and I think the project is fairly up to date on that.

nfischer:
Can you test this on node v8? I don't see an option to run the GitHub workflows for this change, so you may need to refresh this branch for the option to appear.

sebdroid:
> We intentionally use old versions of some projects because the old versions have better nodejs compatibility. I'm generally happy to accept package updates one-by-one IF those updates don't break old node versions, however I run this process occasionally and I think the project is fairly up to date on that.

just curious as to why older unsupported/unmaintained versions of node are supported, is it not better to remove support for those versions and update package dependencies? any projects relying on older versions of node can still use older versions of this package.

MikeKav:
I'd be keen to see an update, I have a project using shx which depends on this package and I've a number of warnings on glob versioning when building my project

nfischer:
I will manage dependency updates myself. Thanks for expressing interest.

PR #1179, opened 2025-02-12
Add the exit code to the fatal error thrown from common.error().

kmashint:
While `shelljs.exec(cmd)` on fatal errors does have the exit code in the logEntry message, it's not available as a separate property on the thrown Error instance.  The only way to find the exit code on a fatal thrown error is to `require('shelljs/src/common.js').state.errorCode`, an awkward use of an internal.

This adds the code from options.code, also used to set the internal state.errorCode, to the options.fatal error before throwing it, simplifying the ability to `shelljs.exit(err.code)` when catching the Error.  This is similar to how the child_process.exec() has a code on the err object given to its callback, e.g. in src/exec.js execASync().

This updates the package.json to an assumed 0.8.6 since the latest release is 0.8.5.

Something seems to be wrong with node 22.x builds over 22.9, causing errors with npm.  I downgraded locally on Win10 to node 22.9 and it ran without this low-level crash.
https://github.com/tinymce/tinymce/issues/10018
From Ubuntu errors, the problem is on the signal handling in foreground-child@2.0.0 index.js where the signal resolves like a string not a number. 
```
TypeError [ERR_INVALID_ARG_TYPE]: The "code" argument must be of type number. Received type string ('128SIGABRT')
    at process.set [as exitCode] (node:internal/bootstrap/node:123:9)
    at ChildProcess.<anonymous> (/home/runner/work/shelljs/shelljs/node_modules/foreground-child/index.js:63:22)
```
```
  child.on('close', (code, signal) => {
    // Allow the callback to inspect the child’s exit code and/or modify it.
    process.exitCode = signal ? 128 + signal : code
```
`npm ls foreground-child` indicates this comes via the `nyc` dependency which has a new version 17.1.0 that requires newer foreground-child@3.3.0.  But underlying that is spawn-wrap@2.0.0 that has no newer version and uses the problematic foreground-child@2.0.0.  I tried forcing some overrides and directly replacing versions but it's still broken with the node 22.14.0, as it is with 22.13.1 in the cloud build.  There are also version conflicts where foreground-child@3.3.0 wants signal-exit@3 while spawn-wrap@2 and other dependencies want signal-exit@3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "shelljs",
-  "version": "0.8.6",
+  "version": "0.8.3",
   "description": "Portable Unix shell commands for Node.js",
   "keywords": [
     "shelljs",
@@ -67,7 +67,7 @@
     "ava": "^1.4.1",
     "chalk": "^4.1.2",
     "coffee-script": "^1.12.7",
-    "eslint": "^8.2.0",
+    "eslint": "^5.16.0",
     "eslint-config-airbnb-base": "^15.0.0",
     "eslint-plugin-import": "^2.31.0",
     "js-yaml": "^4.1.0",
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "ava": "^1.4.1",
     "chalk": "^4.1.2",
     "coffee-script": "^1.12.7",
-    "eslint": "^5.16.0",
+    "eslint": "^8.2.0",
     "eslint-config-airbnb-base": "^15.0.0",
     "eslint-plugin-import": "^2.31.0",
     "js-yaml": "^4.1.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "shelljs",
-  "version": "0.8.3",
+  "version": "0.8.6",
   "description": "Portable Unix shell commands for Node.js",
   "keywords": [
     "shelljs",
--- a/src/common.js
+++ b/src/common.js
@@ -119,7 +119,11 @@ function error(msg, _code, options) {
   state.error += logEntry;
 
   // Throw an error, or log the entry
-  if (options.fatal) throw new Error(logEntry);
+  if (options.fatal) {
+    var err = new Error(logEntry);
+    err.code = options.code || DEFAULT_OPTIONS.code;
+    throw err;
+  }
   if (msg.length > 0 && !options.silent) log(logEntry);
 
   if (!options.continue) {
--- a/src/common.js
+++ b/src/common.js
@@ -121,7 +121,7 @@ function error(msg, _code, options) {
   // Throw an error, or log the entry
   if (options.fatal) {
     var err = new Error(logEntry);
-    err.code = options.code || DEFAULT_OPTIONS.code;
+    err.code = options.code;
     throw err;
   }
   if (msg.length > 0 && !options.silent) log(logEntry);
--- a/test/config.js
+++ b/test/config.js
@@ -49,6 +49,14 @@ test.cb('config.fatal = true', t => {
   });
 });
 
+test('config.fatal = false with an exec() failure returns, does not throw', t => {
+  const expected = { code: 2 };
+  t.notThrows(() => {
+    const result = shell.exec('exit 2');
+    t.is(result.code, expected.code);
+  });
+});
+
 test('config.fatal = true with an exec() failure includes a .code on the Error', t => {
   shell.config.fatal = true;
   try {
--- a/test/config.js
+++ b/test/config.js
@@ -49,6 +49,17 @@ test.cb('config.fatal = true', t => {
   });
 });
 
+test('config.fatal = true with an exec() failure includes a .code on the Error', t => {
+  shell.config.fatal = true;
+  try {
+    t.throws(() => {
+      shell.exec('exit 2');
+    }, { code: 2 });
+  } finally {
+    shell.config.fatal = false;
+  }
+});
+
 //
 // config.globOptions
 //
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -30,7 +30,6 @@
     "no-use-before-define": "off",
     "operator-linebreak": "off",
     "quote-props": "off",
-    "space-infix-ops": "off",
     "spaced-comment": ["error", "always", { "markers": ["@", "@include"], "exceptions": ["@", "@commands"] }],
     "strict": "off",
     "vars-on-top": "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -30,6 +30,7 @@
     "no-use-before-define": "off",
     "operator-linebreak": "off",
     "quote-props": "off",
+    "space-infix-ops": "off",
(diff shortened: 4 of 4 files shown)


kmashint:
@nfischer I've tried many variations including removing my changes, and it seems the problem is between the shelljs dependencies and node 22.13.1, or perhaps anything beyond v22.9.0, where limiting to node v22.9.0 works locally.

Is it possible limit the node 22 could tests to 22.9.0 until the deeper dependency issues can be resolved?
Part of the problem is related to getting forground-child@3.3.0 instead of @2.0.0 that has the 128SIGINT bug.  nyc@15.1.0 has an update to @17.1.0, but that breaks other platforms so code and or tests needs to be adjusted.

If we can limit the build tests to 22.9.0 then at least some small improvements could be added while the deeper compatibility issues with node@22 are investigated and resolved.

Thanks for your insight and support!

nfischer:
@kmashint can you file a new bug to share details about things being broken on Node 22? This sort of thing happens from time to time, I'll see what I can do to fix it.

I'll look through your PR when I get a chance.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "shelljs",
-  "version": "0.8.3",
+  "version": "0.8.6",
```
nfischer, line 3: Please undo this part. I will take care of bumping the version when I'm ready to cut a release.

src/common.js
```diff
@@ -119,7 +119,11 @@ function error(msg, _code, options) {
   state.error += logEntry;
 
   // Throw an error, or log the entry
-  if (options.fatal) throw new Error(logEntry);
+  if (options.fatal) {
+    var err = new Error(logEntry);
+    err.code = options.code || DEFAULT_OPTIONS.code;
```
nfischer, line 124: You can remove `|| DEFAULT_OPTIONS.code`. This is redundant because `options = Object.assign({}, DEFAULT_OPTIONS, options);` would already assign the value to `options.code`.

nfischer (changes requested):
Can you please add a unit test for this? `test/config.js` would be a good place for this. I think you can do something like calling `shell.exec('exit 2')`, catch the exception, and check that it has the `.code` property. Remember to unset `shell.config.fatal` at the end of the test (preferably in a `finally` block).

kmashint:
Thanks @nfischer , I've made adjustments to this.  I had a successful test with node@22.9.0, 22.10.0 onward have this new compatibility issue, on Ubuntu and OSX related to the 128SIGINT combination that stems from foreground-child@2.0.0 index.js where the signal resolves like a string not a number.  As you noted, I'll create a separate issue for this.
```
TypeError [ERR_INVALID_ARG_TYPE]: The "code" argument must be of type number. Received type string ('128SIGABRT')
    at process.set [as exitCode] (node:internal/bootstrap/node:123:9)
    at ChildProcess.<anonymous> (/home/runner/work/shelljs/shelljs/node_modules/foreground-child/index.js:63:22)

// related to foreground-child/index.js:63
  child.on('close', (code, signal) => {
    // Allow the callback to inspect the child’s exit code and/or modify it.
    process.exitCode = signal ? 128 + signal : code
```
On Windows I receive a more cryptic error, which I assume is also related to the signal.
```
C:\WINDOWS\system32\cmd.exe [8364]: void __cdecl node::fs::InternalModuleStat(const class v8::FunctionCallbackInfo<class v8::Value> &) at c:\ws\src\node_file.cc:1040
  #  Assertion failed: (args.Length()) >= (2)
```

kmashint:
I haven't worked with Ava tests, but have with Jest and Jasmine.  Jasmine has had a resurgence for its simplicity, and the latest Jest is more than 4x the size of Ava.  Jasmine is the safest against external dependency conflicts with only one, glob, but I'm not familiar enough with Ava to make comparisons.  The latest Ava is down to 180 nested deps, much better than at 1.4.1.

$ npm view jasmine@5.6.0 dependencies
{ glob: '^10.2.2', 'jasmine-core': '~5.6.0' }

$ npx howfat jasmine@5.6.0
**jasmine@5.6.0 (30 deps, 3.65mb, 416 files, ©MIT)**

$ npx howfat ava@6.2.0
**ava@6.2.0 (180 deps, 12.19mb, 3318 files, ©MIT)**

$ npx howfat ava@1.4.1
**ava@1.4.1 (560 deps, 31.72mb, 7635 files, ©MIT)**

$ npx howfat jest@29
Warning: error request aborted for https://registry.npmjs.org/graceful-fs
**jest@29.7.0 (454 deps, 51.86mb, 9332 files, ©MIT)**
├─┬ @jest/core@29.7.0 (359 deps, 42.7mb, 7209 files, ©MIT)

$ npx howfat tape
**tape@5.9.0 (221 deps, 9.1mb, 7503 files, ©MIT)**

package.json
```diff
@@ -65,19 +65,19 @@
   },
   "devDependencies": {
     "ava": "^1.4.1",
-    "chalk": "^1.1.3",
+    "chalk": "^4.1.2",
     "coffee-script": "^1.12.7",
-    "eslint": "^5.16.0",
-    "eslint-config-airbnb-base": "^13.1.0",
-    "eslint-plugin-import": "^2.17.3",
+    "eslint": "^8.2.0",
```
nfischer, line 70: Please remove these dependency changes from your PR. I would like your PR to be small and focused: only changes 1 thing to accomplish 1 goal. This PR should be only your feature (adding `.code` to the exception) and the unit test (`test/config.js`).

I can't merge your PR if it's doing all the things (updating dependencies, dropping nodejs version support, **and also** adding  a new feature + test). If you can rebase on top of the latest master branch and then `git push --force-with-lease` to your feature branch, then that will at least fix some of the errors you encountered on tip-of-tree.

test/config.js
```diff
@@ -49,6 +49,15 @@ test.cb('config.fatal = true', t => {
   });
 });
 
+test.cb('config.fatal = true with an exec() failure includes a .code on the Error', t => {
```
nfischer, line 52: Change this to a regular `test(`, not a `test.cb(`. You only need `test.cb(` when you're calling an async function which has a callback. In your case, you don't want that: you are testing the synchronous version of `shell.exec()`.

test/config.js
```diff
@@ -49,6 +49,15 @@ test.cb('config.fatal = true', t => {
...
+  shell.config.fatal = true;
+  t.throws(() => {
+    shell.exec('exit 2');
+  }, { code: 2 });
+  shell.config.fatal = false;
```
nfischer, line 57: Please move `shell.config.fatal = false` into a `finally` block. I want cleanup code to live in a `finally` block so that it is guaranteed to execute even when the test case fails. Otherwise if this one test case fails, it will leave a side effect which could break every test case which follows this one.

nfischer (changes requested):

kmashint:
Sorry for the confusion from a lingering commit from trying different branches.  I've cleaned up the commits and added another test variation since the code coverage complained.  Tests pass now with the 22.9.0 limiter, and I'll it to you to close the conversations after verifying.  Thanks!

nfischer (approved):

nfischer:
Thanks for the contribution! This should be part of the 0.9.0 release when that eventually goes out.

Issue #1180, opened 2025-02-14
nodejs v22.10.0 onward give errors: void node::fs::InternalModuleStat(const v8::FunctionCallbackInfo<v8::Value>&)


kmashint:
When creating a PR for an improvement to carry forward the exit code from a failed sync exec() to the thrown Error, CI tests in the cloud were failing for nodejs v22.31.1 on Ubuntu, OSX, and Windows.
https://github.com/shelljs/shelljs/pull/1179

Sample errors are below:
```
// From Ubuntu:
TypeError [ERR_INVALID_ARG_TYPE]: The "code" argument must be of type number. Received type string ('128SIGABRT')
    at process.set [as exitCode] (node:internal/bootstrap/node:123:9)
    at ChildProcess.<anonymous> (/home/runner/work/shelljs/shelljs/node_modules/foreground-child/index.js:63:22)

// From Windows:
  #  C:\WINDOWS\system32\cmd.exe [8364]: void __cdecl node::fs::InternalModuleStat(const class v8::FunctionCallbackInfo<class v8::Value> &) at c:\ws\src\node_file.cc:1040
  #  Assertion failed: (args.Length()) >= (2)
```

Some research showed that:

1. The problems started with nodejs v22.10.0.  v22.9.0 ran the shelljs tests successfully.
2. There seems to be a new incompatibility between the newer nodejs signals and forground-child@2.0.0, where forground-child@3.3.0 adjusts its handling (see below).
3. spawn-wrap and signal-exit may also be related, but the most obvious problem is the foreground-child and 128SIGABRT, although there it be some deeper error.
4. The nyc dependency which has a new version 17.1.0 that requires newer foreground-child@3.3.0. But underlying that is spawn-wrap@2.0.0 that has no newer version and uses the problematic foreground-child@2.0.0. I tried forcing some version overrides in package.json but it was still broken.  And with nyc 17.1.0, the test runs were breaking on older nodejs versions as well, so there's some incompatibility there.

```
// forground-child@2.0.0 near index.js:63
  child.on('close', (code, signal) => {
    // Allow the callback to inspect the child’s exit code and/or modify it.
    process.exitCode = signal ? 128 + signal : code

// https://github.com/tapjs/foreground-child/blob/v3.3.0/src/index.ts#L152
  const childHangup = () => {
    try {
      child.kill('SIGHUP')

      /* c8 ignore start */
    } catch (_) {
      // SIGHUP is weird on windows
      child.kill('SIGTERM')
[...]

PR #1182, opened 2025-02-14
chore: pin node v22.9.0

nfischer:
There's a bug when running on node v22.10. This temporarily pins 22 to 22.9.0 to workaround the problem and keep tests passing on CI.

Mitigation for issue #1180.


kmashint (approved):
Worked for the cloud builds.

PR #1206, opened 2025-04-04
Add -B and -A options to grep

abluescarab:
Adds the `-B` (before context) and `-A` (after context) options to grep. I may have gone overboard on tests but I wanted to make sure that most feasible configurations work (combined, separate, with space, without space).

### Notes
* Modifies `globStart` to accept a function with a parameter `args`. This allows `-B` and `-A` to be specified separately and a function to determine where the post-regex arguments begin.
* Enforces these styles:
  * `.editorconfig`: Single quotes in JavaScript files
  * `.eslintrc.json`: `radix` option to `off` (don't require base for `parseInt()`)
  * `package.json`: Prettier `arrowParens` option to `avoid` (don't wrap single parameters in parentheses for arrow functions)

### Usage
```
grep -B <num> [args...]
grep -A <num> [args...]
```
### Examples
```bash
# show 3 lines before
grep -B3 string file1 file2 ...
grep -B 3 string file1 file2 ...

# show 2 lines after
grep -A2 string file1 file2 ...
grep -A 2 string file1 file2 ...

# show 3 lines before and 2 after
grep -B3A2 string file1 file2 ...
grep -B3 -A2 string file1 file2 ...

# show 3 lines before and after
grep -AB3 string file1 file2 ...
grep -AB 3 string file1 file2 ...
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,5 +10,8 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.js]
+quote_type = single
+
 [*.{md,markdown}]
 trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,8 +10,5 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.js]
-quote_type = single
-
 [*.{md,markdown}]
 trim_trailing_whitespace = false
--- a/package.json
+++ b/package.json
@@ -91,8 +91,5 @@
   },
   "engines": {
     "node": ">=18"
-  },
-  "prettier": {
-    "arrowParens": "avoid"
   }
 }
--- a/package.json
+++ b/package.json
@@ -91,5 +91,8 @@
   },
   "engines": {
     "node": ">=18"
+  },
+  "prettier": {
+    "arrowParens": "avoid"
   }
 }
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "shjs": "./bin/shjs"
   },
   "dependencies": {
+    "audit": "^0.0.6",
     "execa": "^1.0.0",
     "fast-glob": "^3.3.2",
     "interpret": "^1.0.0",
@@ -66,6 +67,7 @@
   },
   "ava": {
     "serial": true,
+    "workerThreads": false,
     "powerAssert": false,
     "files": [
       "test/*.js"
@@ -76,7 +78,7 @@
     ]
   },
   "devDependencies": {
-    "ava": "^2.4.0",
+    "ava": "^6.2.0",
     "chalk": "^4.1.2",
     "coffee-script": "^1.12.7",
     "eslint": "^8.2.0",
--- a/src/common.js
+++ b/src/common.js
@@ -31,13 +31,13 @@ var DEFAULT_CONFIG = {
 };
 
 var config = {
-  reset: function () {
+  reset() {
     Object.assign(this, DEFAULT_CONFIG);
     if (!isElectron) {
       this.execPath = process.execPath;
     }
   },
-  resetForTesting: function () {
+  resetForTesting() {
     this.reset();
     this.silent = true;
   },
@@ -83,11 +83,12 @@ exports.convertErrorOutput = convertErrorOutput;
 
 // An exception class to help propagate command errors (e.g., non-zero exit
 // status) up to the top-level. {@param value} should be a ShellString.
-function CommandError(value) {
-  this.returnValue = value;
+class CommandError extends Error {
+  constructor(value) {
+    super(value.toString());
+    this.returnValue = value;
+  }
 }
-CommandError.prototype = Object.create(Error.prototype);
-CommandError.prototype.constructor = CommandError;
 exports.CommandError = CommandError; // visible for testing
 
 // Shows error message. Throws if fatal is true (defaults to config.fatal, overridable with options.fatal)
@@ -190,7 +191,7 @@ function parseOptions(opt, map, errorOptions) {
     throw new TypeError('parseOptions() internal error: map must be an object');
   } else if (!isObject(errorOptions)) {
     throw new TypeError(
-        'parseOptions() internal error: errorOptions must be object'
+        'parseOptions() internal error: errorOptions must be object',
     );
   }
 
--- a/src/common.js
+++ b/src/common.js
@@ -447,11 +447,7 @@ function wrap(cmd, fn, options) {
         // Perform glob-expansion on all arguments after globStart, but preserve
         // the arguments before it (like regexes for sed and grep)
         if (!config.noglob && options.allowGlobbing === true) {
-          var globStart =
-            options.globStart instanceof Function
-              ? options.globStart(args)
-              : options.globStart;
-          args = args.slice(0, globStart).concat(expand(args.slice(globStart)));
+          args = args.slice(0, options.globStart).concat(expand(args.slice(options.globStart)));
         }
 
         try {
@@ -521,20 +517,9 @@ function _register(name, implementation, wrapOptions) {
     if (!DEFAULT_WRAP_OPTIONS.hasOwnProperty(option)) {
       throw new Error("Unknown option '" + option + "'");
     }
-    if (
-      (option === 'globStart' &&
-        typeof wrapOptions[option] !== 'number' &&
-        typeof wrapOptions[option] !== 'function') ||
-      (option !== 'globStart' &&
-        typeof wrapOptions[option] !== typeof DEFAULT_WRAP_OPTIONS[option])
-    ) {
-      throw new TypeError(
-        "Unsupported type '" +
-          typeof wrapOptions[option] +
-          "' for option '" +
-          option +
-          "'"
-      );
+    if (typeof wrapOptions[option] !== typeof DEFAULT_WRAP_OPTIONS[option]) {
+      throw new TypeError("Unsupported type '" + typeof wrapOptions[option] +
+        "' for option '" + option + "'");
     }
   });
 
--- a/src/common.js
+++ b/src/common.js
@@ -447,7 +447,11 @@ function wrap(cmd, fn, options) {
         // Perform glob-expansion on all arguments after globStart, but preserve
         // the arguments before it (like regexes for sed and grep)
         if (!config.noglob && options.allowGlobbing === true) {
-          args = args.slice(0, options.globStart).concat(expand(args.slice(options.globStart)));
+          var globStart =
+            options.globStart instanceof Function
+              ? options.globStart(args)
+              : options.globStart;
+          args = args.slice(0, globStart).concat(expand(args.slice(globStart)));
         }
 
         try {
@@ -517,9 +521,20 @@ function _register(name, implementation, wrapOptions) {
     if (!DEFAULT_WRAP_OPTIONS.hasOwnProperty(option)) {
       throw new Error("Unknown option '" + option + "'");
     }
-    if (typeof wrapOptions[option] !== typeof DEFAULT_WRAP_OPTIONS[option]) {
-      throw new TypeError("Unsupported type '" + typeof wrapOptions[option] +
-        "' for option '" + option + "'");
+    if (
+      (option === 'globStart' &&
+        typeof wrapOptions[option] !== 'number' &&
+        typeof wrapOptions[option] !== 'function') ||
+      (option !== 'globStart' &&
+        typeof wrapOptions[option] !== typeof DEFAULT_WRAP_OPTIONS[option])
+    ) {
+      throw new TypeError(
+        "Unsupported type '" +
+          typeof wrapOptions[option] +
+          "' for option '" +
+          option +
+          "'"
+      );
     }
   });
 
--- a/src/grep.js
+++ b/src/grep.js
@@ -8,14 +8,14 @@ common.register('grep', _grep, {
 
 // Gets the index of the first argument after `regex_filter`
 function _getGlobStart(args) {
-  return (
-    Array.prototype.findIndex.call(args, function (v) {
-      return (
-        v instanceof RegExp ||
-        (typeof v === 'string' && v !== '' && !v.startsWith('-'))
-      );
-    }) + 1
-  );
+  return args[0] === '--'
+    ? 2
+    : Array.prototype.findIndex.call(args, function (v) {
+        return (
+          v instanceof RegExp ||
+          (typeof v === 'string' && v !== '' && !v.startsWith('-'))
+        );
+      }) + 1;
 }
 
 // Gets the value of the `beforeContext` or `afterContext` options, removing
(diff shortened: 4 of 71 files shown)


.editorconfig
```diff
@@ -10,5 +10,8 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.js]
+quote_type = single
```
nfischer, line 14: Please split this into a separate PR

package.json
```diff
@@ -91,5 +91,8 @@
   },
   "engines": {
     "node": ">=18"
+  },
+  "prettier": {
```
nfischer, line 95: What is this?

src/common.js
```diff
@@ -447,7 +447,11 @@ function wrap(cmd, fn, options) {
         // Perform glob-expansion on all arguments after globStart, but preserve
         // the arguments before it (like regexes for sed and grep)
         if (!config.noglob && options.allowGlobbing === true) {
-          args = args.slice(0, options.globStart).concat(expand(args.slice(options.globStart)));
+          var globStart =
+            options.globStart instanceof Function
```
nfischer, line 451: Is this strictly necessary? We were able to implement `head -n <num>` without this. See https://github.com/shelljs/shelljs#head-n-num-file--file-

test/grep.js
```diff
@@ -174,3 +174,118 @@ test('the pattern looks like an option', t => {
...
+//
+const contextOptions = ['test*', 'test/resources/grep/file3'];
+
+// key is expected results and value is an array of arguments to test
+const beforeAndAfterTests = {
+  // before context
+  'line1\nline2 test line\n--\nline1\nline2 test line\nline3 test line\n--\nline7\nline8\nline9\nline10 test line\n--\nline12\nline13\nline14\nline15 test line\n':
+    [['-B3'], ['-B', 3], ['-B3i'], ['-Bi', 3], ['-iB3'], ['-iB', 3]],
+  // before context + line numbers
+  '1-line1\n2:line2 test line\n--\n1-line1\n2-line2 test line\n3:line3 test line\n--\n7-line7\n8-line8\n9-line9\n10:line10 test line\n--\n12-line12\n13-line13\n14-line14\n15:line15 test line\n':
```
nfischer, line 189: I think it's too complicated to keep this all as one mega test case which loops over the array. Can you split this into multiple different test cases?

Simpler is going to be better for me to review.

nfischer (changes requested):
Thanks for implementing this feature. I've provided some initial feedback that you can get started on now, but I'll need to spend more time reviewing this (expect more feedback later).

package.json
abluescarab: Prettier can enforce its rules using a `prettier` key in `package.json` or with a separate file. The default option for Prettier is to always wrap single parameters in arrow functions with parentheses. This meant that when formatting code with Prettier,
```js
test('test name', t => { ... });
```
would become
```js
test('test name', (t) => { ... });
```
This option disables that formatting.

.editorconfig
abluescarab: Moved to #1207.

test/grep.js
abluescarab: These are actually split into separate tests so when run via the terminal, it will run `test('-B3')`, `test('-B 3')`, `test('-Bi 3')`, etc. They're not separated in the code but they run separately for each element in the array. That means there are 78 tests:
```
            B:    -B3, -B 3, -B3i, -Bi 3, -iB3, -iB 3, -B3n, -Bn 3, -B3ni, -Bni 3,
                   -nB3, -nB 3, -niB3, -niB 3, -A2
            A:    -A 2, -A2i, -Ai 2, -iA2, -iA 2, -A2n, -An 2, -A2ni, -Ani 2, -nA2,
                   -nA 2, -niA2, -niA 2, -AB3
    AB (same):    -AB 3, -BA3, -BA 3, -AB3i, -ABi 3, -BA3i, -BAi 3, -iAB3, -iAB 3,
                   -iBA3, -iBA 3, -A2B3
    AB (diff):    -A2 -B3, -A 2 -B 3, -A2B3i, -A2 -B3 -i, -A 2 -B 3 -i, -iA2B3,
                   -i -A2 -B3, -i -A 2 -B 3, -AB3n
AB (-n, same):    -nAB3, -n -AB3, -n -AB 3, -nAB 3, -AB3ni, -niAB3, -ni -AB3,
                   -ni -AB 3, -A2B3n
AB (-n, diff):    -A2 -B3 -n, -A 2 -B 3 -n, -nA2B3, -n -A2 -B3, -n -A 2 -B 3,
                   -nA2 -B3, -nA 2 -B 3, -A2 -nB3, -A 2 -nB 3, -A2n -B3, -An 2 -B 3,
                   -A2 -B3n, -A 2 -Bn 3, -A2B3in, -A2 -B3 -in, -A 2 -B 3 -in, 
                   -inA2B3, -in -A2 -B3, -in -A 2 -B 3
```
The loop in question:
```js
// loop through the test results
Object.keys(beforeAndAfterTests).forEach(k => {
  // loop through each set of arguments
  beforeAndAfterTests[k].forEach(v => {
    // run each set of arguments as a separate test
    test(v.join(' '), t => {
      const result = shell.grep(...v, ...contextOptions);
      t.falsy(shell.error());
      t.is(result.toString(), k);
    });
  });
});
```
Would you still like me to rewrite them separately? So they look like:
```js
test('-B3', t => { ... });
test('-B 3', t => { ... });
test('-B3i', t => { ... });
...
```

src/common.js
abluescarab: I'll check it out!

package.json
nfischer: ShellJS isn't configured to use `prettier` and we don't have that package as a dependency. Realistically, I can't accept configs for stuff that I don't use for this project.

I won't object if you want to use `prettier` for your own changes, as long as the final PR also passes `npm run lint`. But you would have to find a way to do this without adding extra configs in our project.

src/grep.js
```diff
@@ -2,16 +2,55 @@ var common = require('./common');
...
+  var value = null;
+  var replaced = options.replace(regex, function (_, before, after, val, rest) {
+    var result = '';
+    if (before || after || rest) {
+      result = (before + after + rest).trim();
+      if (result && !result.startsWith('-')) {
+        result = '-' + result;
+      }
+    }
+    value = parseInt(val);
```
nfischer, line 34: Please specify `parseInt(val, 10)`. Then you can remove the `radix` eslint suppression.

nfischer: We have some code in shelljs which actually does parseInt using a different radix (see chmod.js and exec.js), so I would like all calls to be explicit about the radix.

package.json
abluescarab: That's understandable, I've removed it.

src/common.js
abluescarab: The latest changes have been pushed which revert `globStart` back to normal. Thanks for linking to `head` so I could figure it out from that!

test/grep.js
abluescarab: The tests have been split out separately and simplified.

src/grep.js
abluescarab: This line is no longer needed.

abluescarab:
Sorry about the commit mess. I'm not sure why checks are failing on macos and ubuntu now.

nfischer:
> Sorry about the commit mess. I'm not sure why checks are failing on macos and ubuntu now.

Oh, this should be a simple fix. Can you run `npm run gendocs` in a terminal? This will regenerate README.md, then you can commit the README changes. That's probably the issue.

If that doesn't fix it, then try running the tests locally and run `git status`. It looks like there's some uncommitted file after the tests finish, and `git status` should tell you exactly what file that is.

src/grep.js
```diff
@@ -22,12 +24,15 @@ common.register('grep', _grep, {
 //@ + `-l`: Print only filenames of matching files.
 //@ + `-i`: Ignore case.
 //@ + `-n`: Print line numbers.
+//@ + `-B <num>`: Show `<num>` lines before each result.
```
nfischer, line 27: Do you want to implement `-C` too? Seems like this is the same as `-AB <num>` which you've already implemented.

abluescarab:
> Oh, this should be a simple fix. Can you run `npm run gendocs` in a terminal? This will regenerate README.md, then you can commit the README changes. That's probably the issue.

Hey, that worked, thanks! Hopefully everything looks good now.

test/grep.js
```diff
@@ -174,3 +174,225 @@ test('the pattern looks like an option', t => {
...
+      '5-line5\n' +
+      '--\n' +
+      '7-line7\n' +
+      '8-line8\n' +
+      '9-line9\n' +
+      '10:line10 test line\n' +
+      '11-line11\n' +
+      '12-line12\n' +
+      '--\n' +
+      '12-line12\n' +
```
nfischer, line 393: This extra line separator doesn't match the behavior I see in unix grep. If the context regions overlap, then unix grep's behavior is to omit the line separator and treat it like one big region.

```
$ grep -A 2 -B 3 'test' test.txt
line1
line2 test line
line3 test line
line4
line5
--
line7
line8
line9
line10 test line
line11
line12
line13
line14
line15 test line
```

test/grep.js
```diff
@@ -174,3 +174,225 @@ test('the pattern looks like an option', t => {
...
+      '--\n' +
+      'line10 test line\n' +
+      'line11\n' +
+      'line12\n' +
+      '--\n' +
+      'line15 test line\n\n'
+  );
+});
+
+test('-A option, -n option', t => {
```
nfischer, line 250: I think you can omit this case. I think `test('-B option, -n option')` and `test('-A option, separated')` cover this case nicely.

src/grep.js
```diff
@@ -22,12 +24,15 @@ common.register('grep', _grep, {
...
+//@ + `-A <num>`: Show `<num>` lines after each result.
 //@
 //@ Examples:
 //@
 //@ ```javascript
 //@ grep('-v', 'GLOBAL_VARIABLE', '*.js');
 //@ grep('GLOBAL_VARIABLE', '*.js');
+//@ grep('-B', 3, 'GLOBAL_VARIABLE', '*.js');
```
nfischer, line 35: Can you please also document a case showing the object-style of passing the `-B` option?

nfischer (changes requested):

src/grep.js
```diff
@@ -39,7 +44,17 @@ function _grep(options, regex, files) {
 
   if (!files && !pipe) common.error('no paths given', 2);
 
-  files = [].slice.call(arguments, 2);
+  var idx = 2;
+  if (options.beforeContext === true) {
```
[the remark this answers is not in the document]
abluescarab: I've added a comment explaining that line instead of rewriting it (maintains consistency with the rest of the code).

src/grep.js
abluescarab: Done!

test/grep.js
abluescarab: I've fixed this, but I had to write an extra loop to solve the issue. That means it will be minimally slower but I'm not sure if there's a better option.

test/grep.js
abluescarab: Done!

src/grep.js
abluescarab: Done!

src/grep.js
```diff
@@ -62,23 +86,101 @@ function _grep(options, regex, files) {
...
+          }
+          result[lineNumber] = {
+            line: line,
+            match: true,
+          };
+          if (options.beforeContext > 0) {
+            // Store the lines with their line numbers to check for overlap
+            lines
+              .slice(Math.max(index - options.beforeContext, 0), index)
+              .forEach(function (v, i, a) {
```
nfischer, line 128: It looks like this will involve a doubly-nested loop (O(n^2) complexity). Would it be possible to instead write this with a single loop (O(n) complexity)?

I think we can iterate through and maintain a queue of a max size B. Once we see a regex match, output the queue of size B (this is all of the "before" context) and output the current line. You can do a similar approach for the "after" context. If you see another regex match before the "after" context is complete, then just print the regex match and start over the with "after" queue.

Let me know if you feel up to trying this out, otherwise I can try to work on this after the PR lands.

src/grep.js
```diff
@@ -62,23 +86,101 @@ function _grep(options, regex, files) {
...
+              )
+              .forEach(function (v, i) {
+                var lineNum = '' + (index + 1 + i + 1);
+                if (!result.hasOwnProperty(lineNum)) {
+                  result[lineNum] = { line: v, match: false };
+                }
+              });
+          }
+          // Only add the result if it's new
+          if (!matches.includes(result)) {
```
nfischer, line 150: This condition seems redundant to me. How could the result already be in the array if the object was created fresh at the start of the loop iteration (line 105)?

One way to double check this would be to temporarily change this to:

```
if (!matches.includes(result)) {
  matches.push(result);
} else {
  console.log('Do not submit this part. This line exists only to check if it shows up in the test coverage result');
}
```

nfischer (approved):

src/grep.js
nfischer: I forgot to mention, but you can run code coverage with `npm run test-with-coverage`. Be warned, this will run a bit slower than a regular `npm test`. You'll need to read through the coverage report carefully to see if this line number is in the "uncovered lines" column.

abluescarab: On line 117, `result` is assigned to the last match, so it's no longer a new object. I've been using `test-with-coverage` while writing tests, and it confirms that this line is indeed being covered!

src/grep.js
abluescarab: I’ve been trying to get it down to O(n) for a while today and still haven’t figured it out. Would you mind giving it a go?

test/grep.js
```diff
@@ -174,3 +174,166 @@ test('the pattern looks like an option', t => {
...
+      '11-line11\n' +
+      '12-line12\n' +
+      '13-line13\n' +
+      '14-line14\n' +
+      '15:line15 test line\n'
+  );
+});
+
+test('-C option', t => {
+  const result = shell.grep('-C', 3, 'test*', 'test/resources/grep/file3');
```
nfischer, line 296: Can you add a test case for `{'-C': 2}` or `{'-C': 1}`? I noticed these test cases use C values which are so large that the output doesn't include any line separators. That's actually a good thing to test, but we should additionally have a test for smaller C values just to verify the code does add the line separators when needed.

test/grep.js
```diff
@@ -174,3 +174,166 @@ test('the pattern looks like an option', t => {
   t.falsy(shell.error());
   t.is(result.toString(), '-v\n-vv\n');
 });
+
+//
+// Before & after contexts
```
nfischer, line 179: Can you add a few more test cases?

* `-C 0` should behave like regular grep, except it should insert line separators
* `-C -1` should trigger an error with `common.error('grep: -1: invalid context length argument')`
* `-C 100` (some number bigger than the total input file length) should just print the whole file without errors, assuming there is at least one match for the regex pattern

nfischer (approved):
LGTM, but with a few more suggestions.

Thanks for working on this feature!

test/grep.js
abluescarab: I added a test case for `{'-C', 1}`.

nfischer (approved):

nfischer:
Released in shelljs v0.10. Thanks for the contribution!

PR #1220, opened 2025-04-10
refactor: explicit handling for execa errors

nfischer:
This adds a dedicated if-condition to handle internal errors from execa (timeout, maxBuffer, etc.) and to distinguish these from "regular" errors (the command executed, but it failed with non-zero status).


y-nk (approved):
as you asked for my PoV, i think it's valid this way.

nfischer:
Thanks!

PR #1225, opened 2025-05-09
Deprecate the shjs binary

nfischer:
This removes the shjs binary. This feature was not widely used and it was buggy. The recommended workaround is to invoke scripts directly with 'node' (you may also need to set your NODE_PATH environmental variable).

Dropping this script lets us trim down the dependency tree.

For those who still need the original script, the source code has moved to https://github.com/shelljs/shjs.


nfischer:
Released in shelljs v0.10

Issue #1227, opened 2025-07-01
`shelljs.find` doesn't return contents of a dir on windows


cristiand391:
Repro: https://github.com/cristiand391/shelljs-find-windows-bug

I put a dummy project dir for testing:
```
project/
└── source
    └── index.js
```

`shelljs.find` used to return the contents of the dir (recursive read):
https://github.com/cristiand391/shelljs-find-windows-bug/blob/db712a1b83e66ed005e78893e7b480d58175f66c/test.js#L5

this works on linux and windows when using shelljs v0.8.5 but fails on windows after updating to v0.9.0 or higher:

### CI runs:
✅  v0.8.5 linux
✅ v0.8.5 windows:
https://github.com/cristiand391/shelljs-find-windows-bug/actions/runs/16007671602

✅ v.0.9.0 linux
❌ v0.9.0 windows:
https://github.com/cristiand391/shelljs-find-windows-bug/actions/runs/16007821671

only 1 filepath was returned:
![Image](https://github.com/user-attachments/assets/a9025846-e891-4307-ba6c-dbdc1bb00f17)

✅  v0.10.0 linux
❌  v0.10.0 windows
https://github.com/cristiand391/shelljs-find-windows-bug/actions/runs/16007701904

only 1 filepath was returned:
![Image](https://github.com/user-attachments/assets/e4ddf4fb-0ceb-4017-bfed-cb2a83a80841)

nfischer:
It's possible that the glob change affected this. Do you think you would be able to write a unit test which would assert the correct behavior? If you can do that then I can see if I can change the glob expression to do the right thing on Windows.

cristiand391:
> Do you think you would be able to write a unit test (as a PR for the shelljs/shelljs repo) which would assert the correct behavior?

sure, will do it tomorrow 👍🏼

PR #1228, opened 2025-07-03
test(find): add test case for absolute path read

cristiand391:
Add test case for `shelljs.find` taking an absolute path.

The current tests are passing on windows but only cover relative cases, I suspect the bug on windows may be with the drive -letter at the start of an absolute path.

https://github.com/shelljs/shelljs/issues/1227

### DIFF
--- a/test/find.js
+++ b/test/find.js
@@ -43,7 +43,7 @@ test('simple path', t => {
   t.is(result.length, 12);
 });
 
-test('absolute path', t => {
+test('absolute path (directory)', t => {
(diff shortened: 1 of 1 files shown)


test/find.js
```diff
@@ -43,6 +43,15 @@ test('simple path', t => {
   t.is(result.length, 12);
 });
 
+test('absolute path', t => {
```
nfischer, line 46: Can you change the test title to "absolute path (directory)"?

PR #1252, opened 2026-04-17
Fix handling of glob expansions for paths with a backslash

### LINKED ISSUE #1246: Backslashes do not work on Windows after switch to fast-glob
The [switch to fast-glob](https://github.com/shelljs/shelljs/commit/13cfe8a258b891d865871b8b443358f04f0043af) broke the handling of paths with backslashes on Windows.

This seems to be a limitation of the fast-glob library. From the [README](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#how-to-write-patterns-on-windows):

>Always use forward-slashes in glob expressions (patterns and [ignore](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#ignore) option). Use backslashes for escaping characters.
[...]
Use the [.convertPathToPattern](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#convertpathtopatternpath) package to convert Windows-style path to a Unix-style path.

This test passes with previous versions but now fails:

```ts
test('cp with Windows-style backslash and glob pattern', t => {
  const result = shell.cp('test\\resources\\file*.txt', t.context.tmp);
  t.falsy(shell.error());
  t.falsy(result.stderr);
  t.is(result.code, 0);
  t.truthy(fs.existsSync(`${t.context.tmp}/file1.txt`));
  t.truthy(fs.existsSync(`${t.context.tmp}/file2.txt`));
});
```

One workaround is to change [this line](https://github.com/shelljs/shelljs/blob/a776c50b92c3f676f4f2afe6c74f0e21d8a9596b/src/common.js#L346) to the following:

```ts
ret = glob.sync(glob.convertPathToPattern(convertQuestionMarkForGlob(listEl)), globOpts);
```

This fixes this specific test case, I'm not sure if this will cause problems in other tests or on other platforms. I tried running all of the existing tests locally but ran into issues with different numbers of tests running each time. Even before making the change, there were also many failing tests (on both Windows and WSL).

Also see https://github.com/shelljs/shx/issues/245, which describes the same issue with the [shx package](https://github.com/shelljs/shx).

BrendanC23:
This PR adds `convertPathToPattern` to `expand` in `common.js` to fix the handling of glob patterns for Windows file paths.

The [switch to fast-glob](https://github.com/shelljs/shelljs/commit/13cfe8a258b891d865871b8b443358f04f0043af) broke the handling of paths with backslashes on Windows.

This is a limitation of the fast-glob library. From the [README](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#how-to-write-patterns-on-windows):

>Always use forward-slashes in glob expressions (patterns and [ignore](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#ignore) option).
Use backslashes for escaping characters.
[...]
Use the [.convertPathToPattern](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#convertpathtopatternpath) package to convert Windows-style path to a Unix-style path.

This PR includes tests for expand and cp that include paths with backslashes. Both of these tests fail without the fix and pass once the fix is added.

Fixes https://github.com/shelljs/shelljs/issues/1246

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -344,7 +344,7 @@ function expand(list) {
       var globOpts = globOptions();
       try {
         // Call convertPathToPattern to handle paths with backslashes
-        // See https://github.com/mrmlnc/fast-glob/blob/096a5b620f4224eb692bd8ff2c8de0e634e50d8e/README.md?plain=1#L658-L682
+        // See https://github.com/mrmlnc/fast-glob#convertpathtopatternpath
         ret = glob.sync(glob.convertPathToPattern(convertQuestionMarkForGlob(listEl)), globOpts);
       } catch (e) {
         // if glob fails, interpret the string literally
--- a/src/common.js
+++ b/src/common.js
@@ -343,7 +343,9 @@ function expand(list) {
       var ret;
       var globOpts = globOptions();
       try {
-        ret = glob.sync(convertQuestionMarkForGlob(listEl), globOpts);
+        // Call convertPathToPattern to handle paths with backslashes
+        // See https://github.com/mrmlnc/fast-glob/blob/096a5b620f4224eb692bd8ff2c8de0e634e50d8e/README.md?plain=1#L658-L682
+        ret = glob.sync(glob.convertPathToPattern(convertQuestionMarkForGlob(listEl)), globOpts);
       } catch (e) {
         // if glob fails, interpret the string literally
         ret = [listEl];
--- a/test/cp.js
+++ b/test/cp.js
@@ -995,10 +995,12 @@ test('copy single file to itself should not erase content', t => {
 });
 
 test('cp with Windows-style backslash and glob pattern', (t) => {
+  utils.skipOnUnix(t, () => {
   const result = shell.cp('test\\resources\\file*.txt', t.context.tmp);
-  t.falsy(shell.error());
-  t.falsy(result.stderr);
-  t.is(result.code, 0);
-  t.truthy(fs.existsSync(`${t.context.tmp}/file1.txt`));
-  t.truthy(fs.existsSync(`${t.context.tmp}/file2.txt`));
+    t.falsy(shell.error());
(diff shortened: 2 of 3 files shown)


src/common.js
```diff
@@ -343,7 +343,9 @@ function expand(list) {
       var ret;
       var globOpts = globOptions();
       try {
-        ret = glob.sync(convertQuestionMarkForGlob(listEl), globOpts);
+        // Call convertPathToPattern to handle paths with backslashes
+        // See https://github.com/mrmlnc/fast-glob/blob/096a5b620f4224eb692bd8ff2c8de0e634e50d8e/README.md?plain=1#L658-L682
```
nfischer: How about https://github.com/mrmlnc/fast-glob#convertpathtopatternpath instead?

test/cp.js
```diff
@@ -993,3 +993,12 @@ test('copy single file to itself should not erase content', t => {
   // File content must be preserved (no data loss)
   t.is(shell.cat(`${t.context.tmp}/myfile.txt`).toString(), 'important data');
 });
+
+test('cp with Windows-style backslash and glob pattern', (t) => {
+  const result = shell.cp('test\\resources\\file*.txt', t.context.tmp);
```
nfischer: I think you will need to wrap this test case with `utils.skipOnUnix(t, () => { ... }`. Check out `test/which.js` for an example of how this works.

nfischer (changes requested):

BrendanC23:
It looks like this is still breaking some other tests on Windows locally (ex: `common.js` tests for `glob syntax, ? for single char in ...` and `config.globOptions respects nobrace`). The former are fixed by swapping the calls to `convertPathToPattern` and `convertQuestionMarkForGlob`, but that doesn't fix the latter. I'll need to take a closer look at what's going wrong.

PR #1253, opened 2026-04-27
fix: report chdir errors from cd

### LINKED ISSUE #1112: The cd command can report an incorrect error when executing in a worker
v16.14.2

v0.8.5

### Operating system:
Mac

### Description of the bug:

When trying to use `shell.cd` within a worker specifying a folder which does exist, the `cd` command fails with the following error:

~~~
not a directory: foo
~~~

However, the underlying `process.chdir` command that shell uses, reports the following:

~~~
process.chdir() is not supported in workers
~~~

lawrence3699:
Fixes #1112.

When `process.chdir()` fails for an existing directory, `cd()` currently reclassifies the target as `not a directory`. In worker threads this hides Node's actual `process.chdir() is not supported in workers` error.

This keeps the existing `no such file or directory` and `not a directory` paths, but reports the original `process.chdir()` error when the target exists and is a directory.

Validation:
- `npx ava test/cd.js --match='reports process.chdir() errors for existing directories'`
- `npx ava test/cd.js`
- `npm run test-with-coverage -- test/cd.js`
- `npm run lint`
- `git diff --check`

I also ran `npm test`; the `cd` tests passed, and the remaining local failures reproduced on a clean checkout before this change: `chmod › Test setgid` and four `ls › long option` assertions where this environment reports gid `0`.

### DIFF
--- a/test/cd.js
+++ b/test/cd.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const { Worker } = require('worker_threads');
 
 const test = require('ava');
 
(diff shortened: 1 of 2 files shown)


test/cd.js
```diff
@@ -41,6 +42,40 @@ test('file not dir', t => {
   t.is(result.stderr, 'cd: not a directory: test/resources/file1');
 });
 
+test('reports process.chdir() errors for existing directories', async t => {
```
nfischer, line 45: Can we rename this test case to "shell.cd() not supported in workers" or something like that?

nfischer (approved):

PR #1254, opened 2026-05-15
ci: run tests on Node 24 and 26

gameroman:


nfischer:
Thanks for the PR! I actually resolved this earlier today when I was fixing #1257. Sorry about that, but please do feel free to send future PRs like this.

PR #1256, opened 2026-06-02
test: add coverage for mkdir -p on existing directories

### LINKED ISSUE #1022: test coverage: mkdir -p over a directory which already exists

MD-Mushfiqur123:
Adds test cases for  on directories that already exist:

1.  on a directory that already exists (should succeed silently)
2.  to create a new subdirectory within an already-existing tree

Fixes #1022

### DIFF
--- a/test/mkdir.js
+++ b/test/mkdir.js
@@ -201,3 +201,24 @@ test('non-normalized paths are still ok with -p', t => {
   t.is(result.code, 0);
   t.truthy(fs.existsSync(`${t.context.tmp}/asdf`));
 });
+
+test('-p flag: existing directory', t => {
+  shell.mkdir(`${t.context.tmp}/existing`);
+  t.falsy(shell.error());
+  t.truthy(fs.existsSync(`${t.context.tmp}/existing`));
+  const result = shell.mkdir('-p', `${t.context.tmp}/existing`);
+  t.falsy(shell.error());
+  t.is(result.code, 0);
+  t.truthy(fs.existsSync(`${t.context.tmp}/existing`));
+});
+
+test('-p flag: create new subdirectory within existing tree', t => {
+  shell.mkdir('-p', `${t.context.tmp}/grandparent`);
+  t.falsy(shell.error());
+  t.truthy(fs.existsSync(`${t.context.tmp}/grandparent`));
+  const result = shell.mkdir('-p', `${t.context.tmp}/grandparent/parent/child`);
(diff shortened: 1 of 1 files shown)


test/mkdir.js
```diff
@@ -196,8 +196,28 @@ test('globbed dir', t => {
...
-  t.is(result.code, 0);
-  t.truthy(fs.existsSync(`${t.context.tmp}/asdf`));
+	const result = shell.mkdir('-p', `${t.context.tmp}/asdf/../asdf/./`);
+	t.falsy(shell.error());
+	t.is(result.code, 0);
+	t.truthy(fs.existsSync(`${t.context.tmp}/asdf`));
+});
+
+test('-p flag: existing directory', t => {
+	shell.mkdir('-p', `${t.context.tmp}/existing`);
```
nfischer: Remove the '-p' from this call, to make it clearer that this is part of the setup and not the real test (which is the next call to `shell.mkdir('-p', ...)`).

test/mkdir.js
```diff
@@ -196,8 +196,28 @@ test('globbed dir', t => {
...
+	t.falsy(shell.error());
+	t.truthy(fs.existsSync(`${t.context.tmp}/existing`));
+	const result = shell.mkdir('-p', `${t.context.tmp}/existing`);
+	t.falsy(shell.error());
+	t.is(result.code, 0);
+	t.truthy(fs.existsSync(`${t.context.tmp}/existing`));
+});
+
+test('-p flag: create new subdirectory within existing tree', t => {
+	shell.mkdir('-p', `${t.context.tmp}/grandparent/parent`);
```
nfischer: suggestion: let's remove `parent` from this call (but leave it in the next call).

Then in the portion of the test below, let's have calls to `fs.existsSync()` for both `'.../parent'` and `'.../parent/child'` to verify both directories are created. This will give us test coverage that `mkdir('-p')` can create multiple directories in the hierarchy while also testing that we can work within an existing directory tree.

nfischer (approved):
Thank you! Two optional suggestions, but this looks good

PR #1268, opened 2026-09-12
fix(grep): include final context line without trailing newline

emme1t:
`grep -A` and `grep -C` omit the final context line when input ends without a newline. For example, `ShellString('match\nafter').grep('-A', 1, 'match')` returns only `"match\n"` with exit code 0.

Calculate the actual line count when bounding the after-context slice. This includes the final text line for inputs without a terminating newline and continues to exclude the empty split element created by a terminating newline. Regression tests cover `-A`, `-C`, line numbering, both input endings, and a real final blank line.

Validation on Node.js v24.16.0 / Windows:

- Before the fix: 2 of the 5 new regressions fail, both for input without a trailing newline.
- `npx ava test/grep.js test/pipe.js --tap`: 50 passing.
- `npm run lint` and `git diff --check`: passing.
- Five input/option combinations match GNU grep's stdout and exit code.
- `npm test -- --tap`: 619 passing, 10 failing, 6 skipped. Original HEAD has the exact same 10 failing test names (614 passing, 10 failing, 6 skipped): two `cmd` tests expect Unix-style command errors, and eight concern symlinks in `cp`, `find`, and `ln` on this Windows setup.

Investigated and implemented with OpenAI Codex assistance; reproduction and tests were executed locally.


lbesecker195 (approved):
Compared `shell.grep` against the system grep (ugrep 7.8.4, GNU-compatible output) on 8 inputs. On main, `-A`/`-C` drop the last context line whenever the file doesn't end in a newline (e.g. `-A 1` on `before\\nmatch\\nafter` gives `match\\n` instead of `match\\nafter\\n`, and `-C 1` on `match\\nb\\nmatch\\nc` loses `c`). With this branch all 8 match grep, including the trailing-newline and trailing blank line cases, and `npx ava test/grep.js` passes 40 tests. LGTM.

